PR #1, opened 2023-09-28
Added beginnings of a Dockerfile

jdob:


e-minguez:
I don't think we should install mkisofs but use a Golang library to create the iso.

Also we should build the binary in a builder container step then just copy the binary to a final container with the bare minimum.

jdob:
I shouldn't have even included the mkisofs stuff in the first place, we're not even at the point of being able to run the CLI yet :) I'll take that part out.

The builder container is a good idea. I don't have much experience with productized container images and was able to cut a lot of those corners when I was doing dev advocacy, so I'm glad to have eyes on this. Is that sort of thing done with manual scripting (i.e. build the image that builds eib, get it out of the image, and shove it into the new one) or can that be done completely in the Dockerfile?

Dockerfile
```diff
@@ -0,0 +1,14 @@
+# ----- EIB Builder Image -----
+FROM golang:1.21
+
+WORKDIR /src
+COPY . ./
+RUN go build ./cmd/eib
+
+
+# ----- Deliverable Image -----
+FROM registry.suse.com/bci/bci-base:15.4
```
atanasdinov, line 10: I'm not sure which of the BCI images is the most suitable for our use case. AFAICS, we have the following options:
- bci-base
- bci-minimal
- bci-micro

The size difference of the final image between `bci-base` and `bci-micro` exceeds 100 MB so this is definitely something we'd want to investigate later. I'm good with starting with `bci-base` though.

I believe we should be using SP5 though (i.e. 15.5).

diconico07: eib seems to be a nice static binary, so bci-micro is the right one

Dockerfile
```diff
@@ -0,0 +1,14 @@
+# ----- EIB Builder Image -----
+FROM golang:1.21
```
diconico07, line 2: you should use the bci-golang here to stay consistent:
```suggestion
FROM registry.suse.com/bci/golang:1.21
```

jdob: I didn't realize this was a thing, but I dig it. Thanks for the heads up :)

Dockerfile
jdob: Tested a change to micro (and the correct version, not sure why I used 15.4) and it looks good, I'll make the change.

atanasdinov (approved):

diconico07 (approved):

PR #2, opened 2023-10-03
Added basic structure for configuration parsing

jdob:
This does not yet include any sort of validation on the data, that'll come in a future change.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,10 +1,13 @@
-module eib
+module github.com/suse-edge/edge-image-builder
 
 go 1.21.1
 
 require (
+	github.com/stretchr/testify v1.8.4
+	gopkg.in/yaml.v3 v3.0.1
+)
+
+require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
-	github.com/stretchr/testify v1.8.4 // indirect
-	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
--- a/go.mod
+++ b/go.mod
@@ -1,3 +1,10 @@
 module eib
 
 go 1.21.1
+
+require (
+	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/pmezard/go-difflib v1.0.0 // indirect
+	github.com/stretchr/testify v1.8.4 // indirect
+	gopkg.in/yaml.v3 v3.0.1 // indirect
+)
--- a/pkg/config/image.go
+++ b/pkg/config/image.go
@@ -7,10 +7,7 @@ import (
 )
 
 type ImageConfig struct {
-	// EIB Metadata
-	ApiVersion string `yaml:"apiVersion"`
-
-	// Image Base Configuration
+	APIVersion string `yaml:"apiVersion"`
 	ImageType  string `yaml:"imageType"`
 }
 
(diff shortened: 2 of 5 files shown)


pkg/config/image_test.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestParse(t *testing.T) {
+	// Setup
+	filename := "./testdata/valid_example.yaml"
+	configData, err := os.ReadFile(filename)
+	if err != nil {
+		t.Errorf("Cannot read example YAML file: %v", filename)
+	}
```
atanasdinov, line 16:
```suggestion
	require.NoError(t, err)
```

While failing a test with `t.Errorf()` is fine, its execution will still continue unless you also manually include a `return` statement right after. E.g. if we invalidate the example YAML in any way, the test will fail here, however we will also run into nil pointer dereferencing panics on the verifications steps below which will be confusing.

The proposed change will fail and stop the execution, as well as keep up the consistency of using the `testify` module. Also applies to the assertion of the parsing error below.

pkg/config/image_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+func TestParseBadConfig(t *testing.T) {
+	// Setup
+	badData := []byte("Not actually YAML")
+
+	// Test
+	_, err := Parse(badData)
+
+	// Verify
+	assert.Error(t, err)
```
atanasdinov, line 37:
```suggestion
	require.Error(t, err)
	assert.ErrorContains(t, err, "could not parse the image configuration")
```

Nit: Adding some more context for the error.

pkg/config/image.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"fmt"
+
+	"gopkg.in/yaml.v3"
+)
+
+type ImageConfig struct {
+	// EIB Metadata
+	ApiVersion string `yaml:"apiVersion"`
```
atanasdinov, line 11:
```suggestion
	APIVersion string `yaml:"apiVersion"`
```
Acronyms must be capitalised. Linters will complain otherwise.

pkg/config/image.go
```diff
@@ -0,0 +1,26 @@
+package config
+
+import (
+	"fmt"
+
+	"gopkg.in/yaml.v3"
+)
+
+type ImageConfig struct {
+	// EIB Metadata
```
atanasdinov, line 10: Such comments are unnecessary and might be misleading. Every comment above a field or struct definition serves as a documentation entry point i.e. if you open up the "documentation" for `ApiVersion` it will display this text.

I imagine we'll end up with nested structs for the metadata, base config, etc. anyway so getting rid of the comments is probably the best idea.

pkg/config/image_test.go
jdob: Oh perfect, I was looking for something like this but I didn't find it.

pkg/config/image.go
jdob: Ahh, interesting. Good call though, you get my intention but I see how from the language standpoint it's the wrong way to achieve it.

go.mod
```diff
@@ -1,3 +1,10 @@
 module eib
```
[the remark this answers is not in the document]
jdob: I hate that git repos are mentioned in the code (pet peeve), but I also recognize that's the convention.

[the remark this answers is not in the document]
atanasdinov: It's for making your module unique. You can use any arbitrary string but this is the easiest way to understand where something is coming from so everyone is following this convention.

PR #3, opened 2023-10-04
Added initial CLI arguments

jdob:


cmd/eib/main.go
```diff
@@ -1,7 +1,85 @@
...
+func init() {
+	log.SetFormatter(&log.TextFormatter{
+		FullTimestamp:    true,
+		QuoteEmptyFields: true,
+	})
+	log.SetOutput(os.Stdout)
+}
+
+func parseFlags(programName string, args []string) (imageConfig *config.ImageConfig, output string, err error) {
+	flags := flag.NewFlagSet(programName, flag.ContinueOnError)
```
atanasdinov, line 27: Flag sets are used when a command is expected to have subcommands. Are there such plans?

cmd/eib/main.go
```diff
@@ -1,7 +1,85 @@
...
+
+func init() {
+	log.SetFormatter(&log.TextFormatter{
+		FullTimestamp:    true,
+		QuoteEmptyFields: true,
+	})
+	log.SetOutput(os.Stdout)
+}
+
+func parseFlags(programName string, args []string) (imageConfig *config.ImageConfig, output string, err error) {
```
atanasdinov, line 26: This function does a whole lot more than just parsing of flags. Adding more flags to it and using them for business logic will make this grow exponentially.

One approach is to leave the parsing in the main function and to create separate functions covering their usages e.g. loading the config.

jdob: The reason I pulled them out of main was so I could more easily test them. I do see that point that its also loading the image config. Lemme think about that, but part of me just wants to leave it as is and refactor once we have more flags.

cmd/eib/main.go
jdob: Ah, no, at least not currently. It wasn't an intentional decision, just me misreading the API.

cmd/eib/main.go
```diff
@@ -1,7 +1,74 @@
...
+
+func init() {
+	log.SetFormatter(&log.TextFormatter{
+		FullTimestamp:    true,
+		QuoteEmptyFields: true,
+	})
+	log.SetOutput(os.Stdout)
+}
+
+func processArgs() (imageConfig *config.ImageConfig) {
```
atanasdinov, line 24: `processArgs`, as the name suggests, should not return anything. It's better to inline the parsing in the main function.

Bonus tip: The return values should not be named unless you specifically use the variables for some type of modification.

cmd/eib/main.go
```diff
@@ -1,7 +1,74 @@
...
+	return imageConfig
+}
+
+func handleVerbose(verbose bool) {
+	if verbose {
+		log.SetLevel(log.DebugLevel)
+	}
+}
+
+func handleImageConfig(configFile string) *config.ImageConfig {
```
atanasdinov, line 46: It'd be better for this function to return (`*config.ImageConfig, error`), evaluate the error when calling it from `main()` and if present - exit the app with a non-0 code (i.e. `log.Fatalf()`).

The key takeaway is to avoid having unexpected parts of the code causing an early exit. Functions like this one should either return the expected value on success or error on failure - it's not different from the perfectly designed [Parse()](https://github.com/suse-edge/edge-image-builder/blob/main/pkg/config/image.go#L14) one you added previously.

cmd/eib/main.go
jdob: You read my mind, I was gonna specifically ask about the named return values.

cmd/eib/main.go
```diff
@@ -1,7 +1,74 @@
...
+	if configFile == "" {
+		log.Fatalf("the \"%s\" argument must be specified", argConfigFile)
+	}
+
+	info, err := os.Stat(configFile)
+	if os.IsNotExist(err) {
```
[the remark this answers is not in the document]
jdob: Good catch

cmd/eib/main.go
jdob: Makes sense, not sure why I changed the strategy between that one and this.

cmd/eib/main.go
jdob: The first part I disagree with though. IMO, `processArgs` should do everything related to the arguments to convert them from the CLI call into the pieces that will be handed to the image build process. I'm open to another name if the term `process` feels misleading (I intentionally chose that over simply `parse`), but I like the idea of a single entry point that does all of the messing around with arguments.

I'm going to leave it as is for now and see how it evolves when/if we add new flags.

cmd/eib/main.go
```diff
@@ -1,7 +1,72 @@
...
+
+func handleVerbose(verbose bool) {
+	if verbose {
+		log.SetLevel(log.DebugLevel)
+	}
+}
+
+func handleImageConfig(configFile string) (*config.ImageConfig, error) {
+	if configFile == "" {
+		return nil, fmt.Errorf("the \"%s\" argument must be specified", argConfigFile)
```
atanasdinov, line 49: Just FYI, we should avoid embedding such data in the errors and use structured logging instead but this can be done once we replace logrus.

atanasdinov (approved):

PR #4, opened 2023-10-09
initial structure for assembling the combustion directory

jdob:


pkg/build/build.go
```diff
@@ -0,0 +1,86 @@
...
+	return err
+}
+
+//go:embed scripts/script_base.sh
+var scriptBase string
+
+
+func prepareBuildDir(buildConfig *config.BuildConfig) error {
+
+	/* Combustion works by creating a volume with a subdirectory named "combustion"
```
dbw7, line 33: Nothing major, there's an extra space in front of the */

pkg/build/build.go
```diff
@@ -0,0 +1,86 @@
...
+	   and a file named "script". This function builds out that structure and updates
+	   the BuildConfig so that the other functions can populate it as necessary.
+	 */
```
dbw7, line 36: Nothing major, there's an extra space in front of the `*/`

jdob: Fixed

dbw7 (approved):
LGTM

pkg/build/message.go
```diff
@@ -0,0 +1,23 @@
...
+
+	"github.com/suse-edge/edge-image-builder/pkg/config"
+)
+
+const (
+	messageScriptName = "message.sh"
+)
+
+func ConfigureMessage(buildConfig *config.BuildConfig) error {
+	// There's currently intentionally no way to disable this in the image config,
```
atanasdinov, line 14: Not sure what we refer to here. Could you please explain?

pkg/build/message.go
```diff
@@ -0,0 +1,23 @@
...
+	"path/filepath"
+
+	"github.com/suse-edge/edge-image-builder/pkg/config"
+)
+
+const (
+	messageScriptName = "message.sh"
+)
+
+func ConfigureMessage(buildConfig *config.BuildConfig) error {
```
atanasdinov, line 13: We should opt for using unexported functions if we don't expect them to be called outside of a particular package.

pkg/build/build.go
```diff
@@ -0,0 +1,85 @@
...
+	err = cleanUpBuildDir(buildConfig)
+	return err
+}
+
+//go:embed scripts/script_base.sh
+var scriptBase string
+
+func prepareBuildDir(buildConfig *config.BuildConfig) error {
+
+	/* Combustion works by creating a volume with a subdirectory named "combustion"
```
atanasdinov, line 32: Comments describing the purpose of a function should instead be moved just above it. Usually multiple single-line comments ("//") are preferred to multi-line ones ("/*"). For reference: https://go.dev/doc/effective_go#commentary

pkg/build/build.go
```diff
@@ -0,0 +1,85 @@
...
+func cleanUpBuildDir(buildConfig *config.BuildConfig) error {
+	if buildConfig.DeleteArtifacts {
+		err := os.Remove(buildConfig.BuildTempDir)
+		return err
+	}
+	return nil
+}
+
+func generateCombustionScript(buildConfig *config.BuildConfig) error {
+
```
atanasdinov, line 61: Linters complain if we start a function with an empty line.

pkg/build/build.go
```diff
@@ -0,0 +1,85 @@
...
+	   the BuildConfig so that the other functions can populate it as necessary.
+	*/
+
+	if buildConfig.BuildTempDir == "" {
+		tmpDir, err := os.MkdirTemp("", "eib-")
+		if err != nil {
+			return err
+		}
+		buildConfig.BuildTempDir = tmpDir
+	}
```
atanasdinov, line 43: I'm a little confused here. Generally, this is part of the config initialization logic so we should not be concerned with it in the `build` package. Do we expect that if a `BuildTempDir` is provided, then it exists?

atanasdinov (changes requested):

pkg/build/util.go
```diff
@@ -0,0 +1,22 @@
...
+import (
+	"github.com/suse-edge/edge-image-builder/pkg/config"
+	"os"
+	"path/filepath"
+)
+
+func copyCombustionFile(sourcePath string, buildConfig *config.BuildConfig) error {
+	src, err := os.ReadFile(sourcePath)
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
jdob: Really glad you commented on this because I was wondering about this. I have played with Errorf before but wasn't sure if it was overkill.

pkg/config/build.go
```diff
@@ -0,0 +1,14 @@
...
+type BuildConfig struct {
+	ImageOutputDir  string
+	BuildTempDir    string
+	DeleteArtifacts bool
+
+	CombustionDir     string
+	CombustionScripts []string
+}
+
+func (bc *BuildConfig) AddCombustionScript(filename string) {
```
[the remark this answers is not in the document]
jdob: I wasn't viewing it as an input config per se, I was actually considering it more of a context type of object. Sure, it's pre-populated from things passed in from the user (and some generated), but I didn't view it as something that needs to be immutable.

What about if I change it to BuildContext. Does that feel better to you?

pkg/build/message.go
jdob: This was another place I was going to specifically ask you about. I'll change.

pkg/build/build.go
jdob: I'll change to using //, but I don't consider this documentation about the function itself since the caller doesn't need to know the details. The details are for the devs reading the function to give some idea of what the structure we're assembling looks like and why.

pkg/build/build.go
jdob: Good to know, I'll fix and remember in the future.

pkg/build/build.go
jdob: Yes, exactly. My next patch (the one I was talking about with regard to rebases in Slack) is actually going to allow the user to indicate the build directory (I may change the name of this depending on how it evolves). The idea was that this function takes care of any sort of operations related to initializing that build directory, so creating subdirs, etc. I considered the optional-ness of it neededing to be specified as something the builder itself is responsible for resolving, because I don't think that's logic that should exist in the CLI itself (it shouldn't care about how to resolve optional values, simply that it can or cannot accept them) and it didn't make sense to live in the config itself.

pkg/config/build.go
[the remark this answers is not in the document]
jdob: For completeness, I'm migrating the build state stuff to be in a Builder struct and converting the existing functions to use that.

pkg/build/message.go
jdob: Probably overkill. There's no flag in the image config file that says "Don't drop a 'configured by'" message. It wasn't an omission to check for that flag, we just drop it in every case. But I'm overthinking things with the comment and removed it.

pkg/build/build.go
jdob: Going to leave as is for now, but call me out on this in the future once we're reading in the user build dir.

pkg/build/build_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/suse-edge/edge-image-builder/pkg/config"
+)
+
+func TestPrepareBuildDir(t *testing.T) {
+	// Setup
+	bc := config.BuildConfig{}
+	builder := New(nil, &bc)
```
atanasdinov, line 16: Nit: This is an example where instantiating looks much cleaner. Doesn't need to happen in this PR but it can look like:
```
	builder := Builder{
		buildConfig: &config.BuildConfig{},
	}
```

pkg/build/build_test.go
```diff
@@ -0,0 +1,92 @@
...
+	bc.BuildTempDir = testDir
+
+	// Test
+	err = builder.cleanUpBuildDir()
+
+	// Verify
+	require.NoError(t, err)
+	_, err = os.Stat(bc.BuildTempDir)
+	require.Error(t, err)
+	require.True(t, os.IsNotExist(err))
```
atanasdinov, line 48: Nit: Can be `assert`. We usually only `require` the critical statements. Also applies to other tests in the same file.

pkg/build/build_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+	builder.combustionScripts = append(builder.combustionScripts, "foo.sh", "bar.sh")
+
+	// Test
+	err = builder.generateCombustionScript()
+
+	// Verify
+	require.NoError(t, err)
+
+	scriptBytes, err := os.ReadFile(filepath.Join(builder.combustionDir, "script"))
```
atanasdinov, line 87: `err` is not checked.

pkg/build/message.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"fmt"
+)
+
+const (
+	messageScriptName = "message.sh"
+	messageScriptsDir = "message"
+)
+
+func configureMessage(b *Builder) error {
```
atanasdinov, line 12: I'd suggest to convert this to a method instead. It's not in the same file as the `Builder` definition but it is allowed.

atanasdinov (approved):

PR #5, opened 2023-10-11
Modifications to build from an image config dir

jdob:
There still needs to be work done in here, but this is the basis of the new design around an image config dir instead of just a file. There are likely going to be some changes to make this run more smoothly in a container, but I'm not looking to answer that yet.

### DIFF
--- a/cmd/eib/main.go
+++ b/cmd/eib/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"errors"
 	"flag"
 	"fmt"
 	"os"
@@ -10,7 +11,8 @@ import (
 )
 
 const (
-	argConfigFile = "config"
+	argConfigFile = "config-file"
+	argConfigDir  = "config-dir"
 	argVerbose    = "verbose"
 )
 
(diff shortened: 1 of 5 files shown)


cmd/eib/main.go
```diff
@@ -62,8 +74,19 @@ func handleImageConfig(configFile string) (*config.ImageConfig, error) {
 	return imageConfig, nil
 }
 
+func validateImageConfigDir(configDir string) error {
+	if configDir == "" {
+		return errors.New(fmt.Sprintf("-%s must be specified", argConfigDir))
```
atanasdinov, line 79: In general, you'd do:

```suggestion
		return fmt.Errorf("-%s must be specified", argConfigDir)
```

However, avoiding "dynamic" errors is almost always better. I'd just hardcode the name of the arg.

cmd/eib/main.go
```diff
@@ -22,20 +24,34 @@ func init() {
 	log.SetOutput(os.Stdout)
 }
 
-func processArgs() (*config.ImageConfig, error) {
+func processArgs() (*config.ImageConfig, *config.BuildConfig, error) {
```
atanasdinov, line 27: Nit: The more things we add to this function, the more I'd like to just remove it and inline it in main().

atanasdinov (approved):

PR #9, opened 2023-10-12
reworked the message script to be loaded by go:embed

jdob:


pkg/build/build.go
```diff
@@ -122,10 +132,13 @@ func (b *Builder) copyCombustionFile(scriptSubDir string, scriptName string) err
...
 
-	// Keep a running list of all added combustion files. When we add the combustion
+	return nil
+}
+
+func (b *Builder) registerCombustionScript(scriptName string) {
+	// Keep a running list of all added combustion scripts. When we add the combustion
 	// "script" file (the one Combustion itself looks at), we'll concatenate calls to
 	// each of these to that script.
-	b.combustionScripts = append(b.combustionScripts, scriptName)
```
jdob, line 128: Removing this from the default behavior was intentional. We're not going to necessarily want every file in the combustion directory added and attempted to be executed by script, so I took this opportunity to pull it out into its own call.

pkg/build/build.go
```diff
@@ -109,6 +109,16 @@ func (b *Builder) generateCombustionScript() error {
 	return nil
 }
 
+func (b *Builder) writeCombustionFile(contents string, filename string) error {
```
atanasdinov, line 112: Can we reuse this in `copyCombustionFile()`?

atanasdinov (approved):

pkg/build/build.go
jdob: Probably. I thought about deprecating the copy one but I'm not 100% convinced yet that everything will be embeddable. I'll keep it in mind for a future refactor.

PR #10, opened 2023-10-12
wired up CLI to build call

jdob:


cmd/eib/main.go
```diff
@@ -89,10 +90,14 @@ func validateImageConfigDir(configDir string) error {
 }
 
 func main() {
-	_, _, err := processArgs()
+	ic, bc, err := processArgs()
```
dbw7, line 93: Personal opinion, doesn't really matter, I think `imageConfig` and `buildConfig` would make this easier to read on a first pass

jdob: I went back and forth about this. It's (currently) a short enough function that I'm not too worried just yet.

atanasdinov: I also believe we should use the full names - very rarely (if ever) do we us project specific "acronyms" to name variables. This is usually only done in cases where the full version collides with the package name (e.g. `config config.Config`) or a shorter version has already been documented / described in some way or capacity.

jdob: This actually super works for me, especially if both of you feel this way. I tend towards longer, more descriptive names but I feel like I've seen less of that in Go, which is why I went short on this one.

atanasdinov (approved):

PR #11, opened 2023-10-12
added output image name and restructured image config

jdob:


dbw7 (approved):
LGTM

pkg/config/image.go
```diff
@@ -8,7 +8,11 @@ import (
 
 type ImageConfig struct {
 	APIVersion string `yaml:"apiVersion"`
-	ImageType  string `yaml:"imageType"`
+	Image      struct {
+		ImageType       string `yaml:"imageType"`
+		BaseImage       string `yaml:"baseImage"`
+		OutputImageName string `yaml:"outputImageName"`
```
atanasdinov, line 14: Nit: The name of the variable does not need to be the same as the configuration field e.g. we can use a `Type` field instead of `ImageType` here. This way we can avoid a lot of repetition and it's still clear that we're describing the image related fields since they're all part of the `Image` struct.

atanasdinov (approved):

pkg/config/image.go
jdob: Yeah, I know the mapping allows that and it's a cool feature. For now, the ones we have aren't bad and I like the ease of knowing how they map up without needing to check.

PR #13, opened 2023-10-15
ported logging to go zap

jdob:
This isn't meant to be a full audit of the existing code for where logging should be done. Nor is it a necessarily complete configuration of zap, but we can add to that with time based on requirements. What this PR does is configures a basic logger and makes it accessible throughout the app using `zap.L()`


cmd/eib/main.go
```diff
@@ -52,16 +44,26 @@ func processArgs() (*config.ImageConfig, *config.BuildConfig, error) {
...
 	if verbose {
-		log.SetLevel(log.DebugLevel)
+		logLevel = zap.DebugLevel
+	}
+
+	logConfig := zap.Config{
+		Level: zap.NewAtomicLevelAt(logLevel),
 	}
+
+	logger := zap.Must(logConfig.Build())
```
atanasdinov, line 63: I tested it and it panics for me with:
```panic: no encoder name specified```

If I set that up, it no longer panics but still doesn't log anything. Have you tried it?

File: cmd/eib/main.go [hunk 2]
```diff
@@ -92,12 +94,12 @@ func validateImageConfigDir(configDir string) error {
 func main() {
 	imageConfig, buildConfig, err := processArgs()
 	if err != nil {
-		log.Error(err)
+		zap.L().Error("parsing CLI arguments", zap.Error(err))
```
atanasdinov, line 97:
```suggestion
		zap.L().Fatal("parsing CLI arguments", zap.Error(err))
```

We need to either log on `ERROR` level and manually `os.Exit(1)` here or use `Fatal()` which will log on `FATAL` level and exit automatically. We can't proceed with invalid config.

atanasdinov, line 97: The message is suitable for error wrapping, however when it comes to logging we usually want to provide a complete message / event e.g. `processing CLI arguments failed`.

atanasdinov (changes requested):
Nit: We could also add a DEBUG level message logging the configuration options we start with.

cmd/eib/main.go
jdob: Oh shit, I completely misread the error message. I tried it simulating an error so that I'd see the logs and it looked like the error was what I expected.

jdob:
Example error output:
```
./eib
2023-10-16T11:57:37.148-0400 fatal eib/main.go:110 parsing CLI arguments {"error": "parsing image config file : image configuration file \"\" cannot be read: open : no such file or directory"}
main.main
	/Users/jdobies/Code/jdob/edge-image-builder/cmd/eib/main.go:110
runtime.main
	/opt/homebrew/Cellar/go/1.21.1/libexec/src/runtime/proc.go:267
```

atanasdinov (approved):

PR #19, opened 2023-10-25
Reenabled build dir deletion and put behind a config value and CLI arg

jdob:


pkg/build/build_test.go
```diff
@@ -42,6 +42,51 @@ func TestPrepareBuildDirExistingDir(t *testing.T) {
 	require.Equal(t, tmpDir, builder.eibBuildDir)
 }
 
+func TestCleanUpBuildDirTrue(t *testing.T) {
```
atanasdinov, line 45: These two test cases are prime examples of why table testing is great. The only differences between them are the value of the `DeleteBuildDir` flag and the expected output. We can look into revamping test suites later though.

atanasdinov (approved):

pkg/build/build_test.go
jdob: Yeah, I wanted to ask you about that. I did recognize that table testing was probably super useful here, but I couldn't puzzle out on my own how to differentiate the error checking for the existence check.

PR #26, opened 2023-10-26
Refined write* methods in Builder to apply templates if specified

### LINKED ISSUE #23: Migrate the modify-raw-image.sh script to use a templating engine
From the beginning, we recognized the need for a templating engine. There will likely be a number of cases where we have files that will be put into combustion that will need some customization specific to the image.

The first time we hit this is with the modify-raw-image.sh script. As there were only two values at the time it was first implemented (https://github.com/suse-edge/edge-image-builder/pull/18), it used simple string substitution. 

When we decide on an actual templating solution (https://pkg.go.dev/text/template has been suggested), we need to migrate the modify-raw-image.sh script to use that instead.

jdob:
closes #23

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -131,30 +131,39 @@ func (b *Builder) generateCombustionScript() error {
 	return nil
 }
 
-func (b *Builder) writeBuildDirFile(contents string, templateData any, filename string) error {
+func (b *Builder) writeBuildDirFile(filename string, contents string, templateData any) error {
 	destFilename := filepath.Join(b.eibBuildDir, filename)
-	return b.writeFile(contents, templateData, destFilename)
+	return b.writeFile(destFilename, contents, templateData)
 }
 
-func (b *Builder) writeCombustionFile(contents string, templateData any, filename string) error {
+func (b *Builder) writeCombustionFile(filename string, contents string, templateData any) error {
 	destFilename := filepath.Join(b.combustionDir, filename)
-	return b.writeFile(contents, templateData, destFilename)
+	return b.writeFile(destFilename, contents, templateData)
 }
 
-func (b *Builder) writeFile(contents string, templateData any, filename string) error {
-	tmpl, err := template.New(filename).Parse(contents)
-	if err != nil {
-		return fmt.Errorf("creating template for file %s: %w", filename, err)
+func (b *Builder) writeFile(filename string, contents string, templateData any) error {
+	if templateData != nil {
+		tmpl, err := template.New(filename).Parse(contents)
+		if err != nil {
+			return fmt.Errorf("creating template for file %s: %w", filename, err)
+		}
+
+		file, err := os.Create(filename)
+		if err != nil {
+			return fmt.Errorf("creating file %s: %w", filename, err)
+		}
+		defer file.Close()
+
+		err = tmpl.Execute(file, templateData)
+		if err != nil {
+			return fmt.Errorf("applying the template at %s: %w", filename, err)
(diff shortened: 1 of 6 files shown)


pkg/build/scripts/modify-raw-image.sh.tpl
```diff
@@ -1,18 +1,18 @@
...
+#  OutputImage   - Full path to the image to modify
+#  CombustionDir - Full path to the combustion directory
 
 #  Guestfish Commands Explanation
 #
 #  sh "btrfs property set / ro false"
 #  - Enables write access to the read only filesystem
 #
-#  copy-in _s /
+#  copy-in __.OutputImage__ /
```
atanasdinov, line 13: Must be `CombustionDir` instead.

pkg/build/scripts/modify-raw-image.sh.tpl
```diff
@@ -1,18 +1,18 @@
...
 #  - Copies the combustion directory into the root of the image
-#  - _s should be populated with the full path to the built combustion directory
+#  - __.OutputImage__ should be populated with the full path to the built combustion directory
```
atanasdinov, line 15: Nit: We could remove this since we've explained it above already.

pkg/build/build.go
```diff
@@ -130,13 +131,30 @@ func (b *Builder) generateCombustionScript() error {
...
+		return fmt.Errorf("creating template for file %s: %w", filename, err)
 	}
 
+	file, err := os.Create(filename)
+	if err != nil {
+		return fmt.Errorf("writing file %s: %w", filename, err)
+	}
+	defer file.Close()
+
+	tmpl.Execute(file, templateData)
```
atanasdinov, line 156: We need to handle the error here.

pkg/build/build.go
```diff
@@ -130,13 +131,30 @@ func (b *Builder) generateCombustionScript() error {
...
+func (b *Builder) writeFile(contents string, templateData any, filename string) error {
+	tmpl, err := template.New(filename).Parse(contents)
 	if err != nil {
-		return fmt.Errorf("writing file %s: %w", destFilename, err)
+		return fmt.Errorf("creating template for file %s: %w", filename, err)
 	}
 
+	file, err := os.Create(filename)
+	if err != nil {
+		return fmt.Errorf("writing file %s: %w", filename, err)
```
atanasdinov, line 152:
```suggestion
		return fmt.Errorf("creating file %s: %w", filename, err)
```

We're not technically writing yet.

pkg/build/build.go
```diff
@@ -130,13 +131,30 @@ func (b *Builder) generateCombustionScript() error {
...
+	return b.writeFile(contents, templateData, destFilename)
+}
+
+func (b *Builder) writeCombustionFile(contents string, templateData any, filename string) error {
 	destFilename := filepath.Join(b.combustionDir, filename)
-	err := os.WriteFile(destFilename, []byte(contents), os.ModePerm)
+	return b.writeFile(contents, templateData, destFilename)
+}
+
+func (b *Builder) writeFile(contents string, templateData any, filename string) error {
```
atanasdinov, line 144: The problem with this design is that standard files (where `templateData` is nil) will still go through the process of parsing content as a template and executing it.

Can we either restructure the `writeX()` functions or add a check for the `templateData` in `writeFile()` and only follow templating logic if we're dealing with a template?

atanasdinov (changes requested):

atanasdinov (approved):

PR #27, opened 2023-10-27
Set up initial CI config

atanasdinov:


.golangci.yml
```diff
@@ -0,0 +1,45 @@
+linters-settings:
+  gocritic:
+    enabled-tags:
+      - diagnostic
+      - style
+      - performance
+    disabled-checks:
+      - whyNoLint
```
jdob, line 8: Can we add `paramTypeCombine` to the disabled checks? I'm willing to adapt to many of the Go conventions, but I feel like this one sacrifices readability in favor of a slightly shorter function signature.

PR #28, opened 2023-10-30
added handling for adding rpms to an image as well as the relevant tests

dbw7:
Currently working on getting elemental registration through combustion working in eib. So far, I've been able to get it work with some few hacks (which are omitted from here, along with the Elemental specific stuff as it is not complete). To get there though, I needed to be able to add the RPMs to the image, so I built the handling for that as well as the tests. I’m not sure if this is how you guys would have gone about it, so I’ve made this PR a draft for now

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -180,7 +180,7 @@ func (b *Builder) generateBaseImageFilename() string {
 	return filename
 }
 
-func (b *Builder) copyFile(sourcePath string, destPath string) error {
+func copyFile(sourcePath string, destPath string) error {
 	sourceFile, err := os.Open(sourcePath)
 	if err != nil {
 		return fmt.Errorf("opening file from source path: %w", err)
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -23,7 +23,6 @@ type Builder struct {
 	eibBuildDir       string
 	combustionDir     string
 	combustionScripts []string
-	rpmBuildDir       string
 	rpmFileNames      []string
 	rpmSourceDir      string
 }
@@ -93,16 +92,10 @@ func (b *Builder) prepareBuildDir() error {
 		b.eibBuildDir = b.buildConfig.BuildDir
 	}
 	b.combustionDir = filepath.Join(b.eibBuildDir, "combustion")
-	b.rpmBuildDir = filepath.Join(b.combustionDir, "rpms")
 
 	err := os.MkdirAll(b.combustionDir, os.ModePerm)
 	if err != nil {
 		return fmt.Errorf("creating the build directory structure for combustion: %w", err)
-	}
-
-	err = os.MkdirAll(b.rpmBuildDir, os.ModePerm)
-	if err != nil {
-		return fmt.Errorf("creating the build directory structure for rpms: %w", err)
 	}
 
 	return nil
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -3,6 +3,7 @@ package build
 import (
 	_ "embed"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 
@@ -23,8 +24,6 @@ type Builder struct {
 	eibBuildDir       string
 	combustionDir     string
 	combustionScripts []string
-	rpmFileNames      []string
-	rpmSourceDir      string
 }
 
 func New(imageConfig *config.ImageConfig, buildConfig *config.BuildConfig) *Builder {
@@ -95,7 +94,7 @@ func (b *Builder) prepareBuildDir() error {
 
 	err := os.MkdirAll(b.combustionDir, os.ModePerm)
 	if err != nil {
-		return fmt.Errorf("creating the build directory structure for combustion: %w", err)
+		return fmt.Errorf("creating the build directory structure: %w", err)
 	}
 
 	return nil
@@ -179,4 +178,25 @@ func (b *Builder) generateOutputImageFilename() string {
 func (b *Builder) generateBaseImageFilename() string {
 	filename := filepath.Join(b.buildConfig.ImageConfigDir, "images", b.imageConfig.Image.BaseImage)
 	return filename
+}
+
+func (b *Builder) copyFile(sourcePath string, destPath string) error {
+	sourceFile, err := os.Open(sourcePath)
+	if err != nil {
+		return fmt.Errorf("opening file from source path: %w", err)
+	}
+	defer sourceFile.Close()
+
+	destFile, err := os.Create(destPath)
+	if err != nil {
+		return fmt.Errorf("creating file at dest path: %w", err)
+	}
+	defer destFile.Close()
+
+	_, err = io.Copy(destFile, sourceFile)
+	if err != nil {
+		return fmt.Errorf("copying file from source path to dest path: %w", err)
+	}
+
+	return nil
 }
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -23,6 +23,9 @@ type Builder struct {
 	eibBuildDir       string
 	combustionDir     string
 	combustionScripts []string
+	rpmBuildDir       string
+	rpmFileNames      []string
+	rpmSourceDir      string
 }
 
 func New(imageConfig *config.ImageConfig, buildConfig *config.BuildConfig) *Builder {
@@ -46,6 +49,11 @@ func (b *Builder) Build() error {
 	err = b.generateCombustionScript()
 	if err != nil {
 		return fmt.Errorf("generating combustion script: %w", err)
+	}
+
+	err = b.copyRPMs()
+	if err != nil {
+		return fmt.Errorf("copying RPMs over: %w", err)
 	}
 
 	switch b.imageConfig.Image.ImageType {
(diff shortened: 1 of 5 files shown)


alknopfler:
Question:
- the folder `rpms` is required? I think we should follow the same as currently doing here: https://documentation.suse.com/sle-micro/5.1/html/SLE-Micro-all/cha-images-combustion.html
WDYT??? (To be fully compatible with the current combustion + ignition implementation)

pkg/build/build.go
```diff
@@ -85,10 +93,16 @@ func (b *Builder) prepareBuildDir() error {
 		b.eibBuildDir = b.buildConfig.BuildDir
 	}
 	b.combustionDir = filepath.Join(b.eibBuildDir, "combustion")
+	b.rpmBuildDir = filepath.Join(b.combustionDir, "rpms")
```
alknopfler, line 96: Why not the same directory like: https://documentation.suse.com/sle-micro/5.5/html/SLE-Micro-all/cha-images-combustion.html

pkg/build/build.go
```diff
@@ -23,6 +23,9 @@ type Builder struct {
 	eibBuildDir       string
 	combustionDir     string
 	combustionScripts []string
+	rpmBuildDir       string
```
alknopfler, line 26: I would use the same directory than combustion dir

pkg/build/rpm.go
```diff
@@ -0,0 +1,52 @@
...
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+)
+
+func (b *Builder) getRPMFileNames() error {
+	b.rpmSourceDir = filepath.Join(b.buildConfig.ImageConfigDir, "rpms")
+
+	rpms, err := os.ReadDir(b.rpmSourceDir)
```
alknopfler, line 13: same

pkg/build/build.go
jdob: My gut reaction was that everything would be in a flat combustion directory. That might get a bit messy for debugging large builds, but I think we should start with that model and only make it more complex if we find a need to. So +1 to Alberto's comment.

dbw7: sounds good to me, my initial thinking was just maintaining the organization from the config dir which I commented on in another reply, but I'll make this change

pkg/build/rpm.go
dbw7: For this one, "rpmSourceDir" refers to the user-created configuration directory. In here, I think it makes it significantly easier for us when the user knows to make an RPM dir and put all of their RPMs in there to be copied over to the image, though I'm interested to see other views

pkg/build/rpm.go
```diff
@@ -0,0 +1,55 @@
+package build
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+)
+
+func (b *Builder) getRPMFileNames() error {
```
atanasdinov, line 10: This will currently modify the internal state of the builder by storing results which should generally not be a part of it.

A function named `get...`, `retrieve...`, `find...`, etc. should return the value itself (and an error if it's possible to fail). This way we decouple temporary data from the struct state and are able to more easily test it.

pkg/build/rpm.go
```diff
@@ -0,0 +1,55 @@
...
+	b.rpmSourceDir = filepath.Join(b.buildConfig.ImageConfigDir, "rpms")
```
atanasdinov, line 11: This is another state modifier which shouldn't be included. In general, we don't store the results of one-off tasks e.g. we do need to store `combustionScripts` because a lot of components will use that variable to store data but the builder itself does not care which directory is being used for RPMs.

Ideally, the RPM source directory should be a local variable which we can pass wherever necessary.

pkg/build/rpm.go
```diff
@@ -0,0 +1,55 @@
...
+		if err != nil {
+			return fmt.Errorf("opening rpm dest path: %w", err)
+		}
+
+		_, err = io.Copy(destFile, sourceFile)
+		if err != nil {
+			return fmt.Errorf("copying rpm: %w", err)
+		}
+
+		sourceFile.Close()
```
atanasdinov, line 50: While this will not trigger the linter, it is not the approach which should be taken.

What will happen if `io.Copy()` fails in this function? Both `sourceFIle` and `destFile` will remain open and leak memory.

The proper thing to do is to encapsulate each iteration in a separate function which in turn would be able to leverage `defer` statements and properly free up the resources.

You could either to that or take a look at `filepath.Walk` which will require you to execute a separate function for each call.

pkg/build/rpm_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/suse-edge/edge-image-builder/pkg/config"
+)
+
+func TestGetRPMFileNames(t *testing.T) {
+	// Setup
+	bc := config.BuildConfig{
+		ImageConfigDir: "../config/testdata",
```
atanasdinov, line 16: Since we're dealing with empty files, we can opt for creating temporary files in order to keep the repository clean. Please take a look at the `os.CreateTemp()` function.

pkg/build/build.go
```diff
@@ -88,7 +95,7 @@ func (b *Builder) prepareBuildDir() error {
 
 	err := os.MkdirAll(b.combustionDir, os.ModePerm)
 	if err != nil {
-		return fmt.Errorf("creating the build directory structure: %w", err)
+		return fmt.Errorf("creating the build directory structure for combustion: %w", err)
```
atanasdinov, line 98: Nit: Do we need this comment change?

atanasdinov (changes requested):

pkg/build/rpm.go
dbw7: oh I see, I forgot to consider that, thank you for pointing that out

pkg/build/build.go
dbw7: not anymore, it was originally changed because I was creating a second dir so I wanted to differentiate them, but I just removed it the second dir creation so I'll change it back

pkg/build/rpm.go
dbw7: 👍  will update

pkg/build/rpm.go
dbw7: 👍  will update

pkg/build/rpm_test.go
dbw7: My thinking at the time was keeping the dir as a dir users could reference as an example to what the structure of config dir may look like. Let me know your thoughts

dbw7:
So a few things, first, I added a test for the situation where the rpm dir is empty, as well as handling for that, let me know how you guys feel about that. Also, because I decided to keep the rpm dir in "testData" using `.gitkeep` I had to add handling for that in `getRPMFileNames()` which I really dislike so that might not survive this pull request. 

As for the other tests, I'm not sure if there was a more graceful way to go about what I did, so let me know if what I did was fine or if it's not convential. The main reason I didn't use `defer` here was to ensure the order of events

pkg/build/rpm.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+func (b *Builder) getRPMFileNames() ([]string, error) {
+	var rpmFileNames []string
+	rpmSourceDir := filepath.Join(b.buildConfig.ImageConfigDir, "rpms")
```
atanasdinov, line 11: We already do this in `copyRPMs()` so we can pass the directory as an argument to this function.

pkg/build/rpm.go
```diff
@@ -0,0 +1,50 @@
...
+
+	rpmSourceDir := filepath.Join(b.buildConfig.ImageConfigDir, "rpms")
+
+	for _, rpm := range rpmFileNames {
+		sourcePath := filepath.Join(rpmSourceDir, rpm)
+		destPath := filepath.Join(b.combustionDir, rpm)
+
+		err = b.copyFile(sourcePath, destPath)
+		if err != nil {
+			return fmt.Errorf("looping through rpms to copy: %w", err)
```
atanasdinov, line 45:
```suggestion
			return fmt.Errorf("copying file %s: %w", sourcePath, err)
```
We aim to provide as specific context to the error as possible. While it is true that this function call is executed during a loop, it itself doesn't loop over anything but copies a single file.

pkg/build/rpm_test.go
```diff
@@ -0,0 +1,117 @@
...
+	require.NoError(t, err)
+
+	err = file2.Close()
+	require.NoError(t, err)
+
+	err = os.Remove(file1Path)
+	require.NoError(t, err)
+
+	err = os.Remove(file2Path)
+	require.NoError(t, err)
```
atanasdinov, line 52: Nit:
```suggestion
	assert.NoError(t, file1.Close())
	assert.NoError(t, file2.Close())
	assert.NoError(t, os.Remove(file1Path))
	assert.NoError(t, os.Remove(file2Path))
```

atanasdinov: Also, it's better to defer the `os.Remove()` calls because the test could exit early (if a `require.F()` fails) and leave the files on the FS. Both comments apply to the other tests as well.

pkg/build/rpm.go
```diff
@@ -0,0 +1,50 @@
...
+		return nil, fmt.Errorf("reading rpm source dir: %w", err)
+	}
+
+	for _, rpmFile := range rpms {
+		rpmFileNames = append(rpmFileNames, rpmFile.Name())
+	}
+
+	if len(rpmFileNames) == 0 {
+		return nil, fmt.Errorf("no rpms found")
+	} else if len(rpmFileNames) == 1 && rpmFileNames[0] == ".gitkeep" {
```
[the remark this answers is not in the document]
dbw7: good idea, thank you

atanasdinov (approved):

PR #32, opened 2023-11-02
Sort combustion scripts before appending them to the combustion scrip…

jdob:
…t file

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"slices"
 	"text/template"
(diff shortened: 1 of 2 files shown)


atanasdinov (approved):

alknopfler:
How we need to support the custom scripts to be executed inside the combustion script? inside the folder I guess, right?

atanasdinov:
Yes, you can just put all custom scripts you wish to execute during combustion at `/scripts` as specified in the README.

This PR only addresses an initial deterministic order (sorting alphabetically the scripts). In a follow up, we'll come with a better solution for this.

PR #35, opened 2023-11-08
modify platform openstack

alknopfler:
add the platform to operating system in order to enable on grub.cfg the kernel args platform.id=openstack required for metal3 workflow

### DIFF
--- a/pkg/build/raw.go
+++ b/pkg/build/raw.go
@@ -51,9 +51,11 @@ func (b *Builder) writeModifyScript() error {
 	values := struct {
 		OutputImage   string
 		CombustionDir string
+		Platform      string
 	}{
 		OutputImage:   b.generateOutputImageFilename(),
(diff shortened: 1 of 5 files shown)


jdob:
I don't think this is the right approach to expose this single kernel argument as a top level config item. I'm not sure if we'd have other platforms besides OpenStack, but if we don't, then I don't think it makes sense to make this an open variable that may only have on value.

We've had on the roadmap to be able to pass in arbitrary kernel arguments for other reasons. I think that functionality is what will answer this, using the general kernel arg setting for this one-off use case rather than treating this case as a specific config item. I'll prioritize working on that and will hopefully have something in the next few days.

alknopfler:
Yes, I think you've got more context than me on the code structure and roadmap of features to be implemented. This PR was only an idea of implementation just to have a quick discussion. I'm gonna close it ;)

PR #38, opened 2023-11-13
added additional rpm handling

dbw7:
Very open to feedback on this


pkg/build/rpm.go
```diff
@@ -27,9 +36,23 @@ func (b *Builder) getRPMFileNames(rpmSourceDir string) ([]string, error) {
...
+		err := copyFile(sourcePath, destPath)
+		if err != nil {
+			return fmt.Errorf("copying file %s: %w", sourcePath, err)
+		}
+	}
+
+	return nil
+}
+
+func (b *Builder) processRPMs() error {
```
jdob, line 53: Personal style note: I like to keep the main entry point (i'd normally say "public" if we weren't working with methods) at the top of the file, with the supporting methods (like copyRPMs) later on. But don't feel like you need to change it, just offering another perspective.

pkg/build/rpm_test.go
```diff
@@ -117,9 +121,110 @@ func TestCopyRPMsNoRPMDir(t *testing.T) {
 	require.NoError(t, err)
 	defer os.Remove(builder.eibBuildDir)
 
+	rpmSourceDir := filepath.Join(builder.buildConfig.ImageConfigDir, "rpms")
```
jdob, line 124: Consider adding a constant to rpm.go with the name of the directory and then referencing that both here and within the RPM handling code itself. I could see us still needing to reorganize the image config directory at some point and it'll be simpler if we have all of the knowledge of how to deal with specific components contained in their respective source code.

pkg/build/rpm_test.go
```diff
@@ -117,9 +121,110 @@ func TestCopyRPMsNoRPMDir(t *testing.T) {
...
+	require.NoError(t, err)
+	defer os.RemoveAll(tmpDir)
+
+	buildConfig := config.BuildConfig{
+		ImageConfigDir: "../config/testdata",
+	}
+	builder := New(nil, &buildConfig)
+	require.NoError(t, builder.prepareBuildDir())
+
+	rpmSourceDir := filepath.Join(builder.buildConfig.ImageConfigDir, "rpms")
```
jdob, line 146: Another reference to "rpms" that should be a constant.

pkg/build/rpm_test.go
jdob: Another alternative, which I actually like better and would like to change in the custom scripts handling, would be to write a `generateRPMPath()` method in rpm.go that returns the full path, so that the knowledge of where it lives doesn't leak out of rpm.go. Then both rpm.go and rpm_test.go can use that method instead of having multiple places where we join the file paths.

pkg/build/rpm_test.go
```diff
@@ -117,9 +121,110 @@ func TestCopyRPMsNoRPMDir(t *testing.T) {
...
+
+	file1Path := filepath.Join(rpmSourceDir, "rpm1.rpm")
+	defer os.Remove(file1Path)
```
jdob, line 149: I believe this should be after the `Create()` call, since if that fails there won't be a file to remove. I'm not sure what will happen if you leave them in this order (`Remove()` may simply exit cleanly since the file isn't there), but from an anal-retentive point of view, defering the remove before the file is created feels weird.

pkg/build/rpm_test.go
```diff
@@ -117,9 +121,110 @@ func TestCopyRPMsNoRPMDir(t *testing.T) {
...
+	file1, err := os.Create(file1Path)
+	require.NoError(t, err)
+
+	file2Path := filepath.Join(rpmSourceDir, "rpm2.rpm")
+	defer os.Remove(file2Path)
```
jdob, line 154: Whatever you choose for the comment above, apply here too.

pkg/build/rpm_test.go
```diff
@@ -117,9 +121,110 @@ func TestCopyRPMsNoRPMDir(t *testing.T) {
...
+	builder := New(nil, &bc)
+	err := builder.prepareBuildDir()
+	require.NoError(t, err)
+	defer os.Remove(builder.eibBuildDir)
+
+	rpmSourceDir := filepath.Join(builder.buildConfig.ImageConfigDir, "rpms")
+	rpmDestDir := builder.combustionDir
+
+	file1Path := filepath.Join(rpmSourceDir, "rpm1.rpm")
+	defer os.Remove(file1Path)
```
jdob, line 198: Here too.

pkg/build/rpm_test.go
```diff
@@ -117,9 +121,110 @@ func TestCopyRPMsNoRPMDir(t *testing.T) {
...
+	file1, err := os.Create(file1Path)
+	require.NoError(t, err)
+
+	file2Path := filepath.Join(rpmSourceDir, "rpm2.rpm")
+	defer os.Remove(file2Path)
```
jdob, line 203: And here.

pkg/build/scripts/rpms/10_rpm_install.sh.tpl
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -euo pipefail
+
+#  Template Fields
+#  RPMs - A string that contains all of the rpms present in the user created config directory, separated by spaces.
+
+rpm -ivh --nosignature {{.RPMs}}
```
jdob, line 7: I think using `rpm` here is okay. The alternative would be to use zypper and hand it the list of local RPM files (I haven't tried it, but I know you can do that with yum). The benefit there would be dependency resolution, but that's a non-issue here because we have to account for the air gapped case. 

For now I think we run with this approach because of a combination of the fact that I can't think of an actual reason why using zypper would be better and because this currently works. We can readdress during testing if issues come up.

pkg/build/rpm.go
```diff
@@ -27,9 +36,23 @@ func (b *Builder) getRPMFileNames(rpmSourceDir string) ([]string, error) {
 	return rpmFileNames, nil
 }
 
-func (b *Builder) copyRPMs() error {
+func (b *Builder) copyRPMs(rpmSourceDir string, rpmDestDir string, rpmFileNames []string) error {
```
jdob, line 39: Since you're passing in all of the information you need, as compared to having this method derive it from the state of Builder, this doesn't need to be bound to Builder.

pkg/build/rpm.go
```diff
@@ -44,15 +67,38 @@ func (b *Builder) copyRPMs() error {
...
+	writtenFilename, err := b.writeCombustionFile(modifyRPMScriptName, modifyRPMScript, &values)
+	if err != nil {
+		return fmt.Errorf("writing rpm script %s: %w", modifyRPMScriptName, err)
+	}
+	err = os.Chmod(writtenFilename, modifyScriptMode)
+	if err != nil {
+		return fmt.Errorf("changing permissions on the rpm script %s: %w", modifyRPMScriptName, err)
+	}
+
+	fmt.Println(writtenFilename, "xaksdokasd")
```
jdob, line 100: I'm going to assume this wasn't meant to be in the PR :)

jdob (changes requested):

pkg/build/rpm.go
dbw7: LMAO, my mistake on that, knew I had to have missed something

pkg/build/rpm.go
dbw7: Yeah I was actually planning to ask about that, I agree with what you said

pkg/build/rpm.go
dbw7: I like that as well, I'll reorganize it

pkg/build/rpm_test.go
```diff
@@ -117,8 +124,142 @@ func TestCopyRPMsNoRPMDir(t *testing.T) {
...
+	require.NoError(t, err)
+	assert.Equal(t, fs.FileMode(modifyScriptMode), stats.Mode())
+
+	foundContents := string(foundBytes)
+	assert.Contains(t, foundContents, "rpm1.rpm")
+	assert.Contains(t, foundContents, "rpm2.rpm")
+
+	// Cleanup
+	assert.NoError(t, file1.Close())
+	assert.NoError(t, file2.Close())
```
atanasdinov, line 186: Closing should always be deferred after opening or creating.

pkg/build/rpm_test.go
```diff
@@ -117,8 +124,142 @@ func TestCopyRPMsNoRPMDir(t *testing.T) {
...
+	// Setup
+	bc := config.BuildConfig{
+		ImageConfigDir: "../config/testdata",
+	}
+	builder := New(nil, &bc)
+	err := builder.prepareBuildDir()
+	require.NoError(t, err)
+	defer os.Remove(builder.eibBuildDir)
+
+	rpmDestDir := builder.combustionDir
```
atanasdinov, line 199: Dropping this variable is actually a cleaner indicator for the assertions. If we'd prefer to keep it instead, then it should be moved to where it is actually used.

pkg/build/rpm.go
```diff
@@ -1,12 +1,50 @@
...
+func (b *Builder) processRPMs() error {
+	rpmSourceDir, err := b.generateRPMPath()
+	if err != nil {
+		return fmt.Errorf("generating rpm path: %w", err)
+	}
+	// Only proceed with processing the RPMs if the directory exists
+	if rpmSourceDir == "" {
+		return nil
+	}
+	rpmDestDir := b.combustionDir
```
ipetrov117, line 27: I am not sure what use we get from moving the `b.combustionDir` property to `rpmDestDir` and using it in `copyRPMs`. Can't we just use `b.combustionDir` directly as a parameter to the `copyRPMs` function? That way we will save up one more line of code that we have to write without making the existing code too complex.

ipetrov117:
> Looks good to me. There are several suggestions that we'd need to take care of but nothing major in those - good job!
> 
> One thing to mention is that there's no documentation around the `/rpms` directory in the README and you might also want to sync with @ipetrov117 since he's also looking into how RPMs will be handled.

The doc around the `/rpms` directory would be of great help both to me and to anyone who is getting familiar with the EIB project.

Regarding the RPM handling, the suggested changes here offer a good base line from where we could continue with the air-gapped RPM handling. Granted, some files would have to be changed (the `rpm_install.sh` script for instance), but most of the code here can be reused. IMO this PR is a step in the right direction.

ipetrov117:
One more thing that got me wondering looking at the [script resolution](https://github.com/suse-edge/edge-image-builder/pull/38/files#r1392183985) comment that @atanasdinov made.  Does it make sense for us to start naming our scripts with `XX_` prefix as @dbw7 has done here? IMO this would provide additional visibility over the flow of the script resolution and will make it easier to track our script resolution when we have more scripts. Furthermore this will decrease any uncertainty of where in the flow to add a new script.

That said, I still do not have the full picture so I might be wrong here. @atanasdinov @jdob @dbw7 wdyt?

atanasdinov:
You're completely correct.

My comment was indeed to point out that the introduced `XX_` scheme in this PR will conflict with the initial alphabetical sorting (which was put there just for determinism purposes). I don't see any issues with starting here and adjusting the other scripts in a follow up. We don't have that many scripts in the first place so the ordering itself shouldn't be hard to figure out.

What we need to consider is how the structure of the `scripts` dir should look like though. Currently there are scripts on root level, in a `message` subdirectory and in a `grub` subdirectory introduced in https://github.com/suse-edge/edge-image-builder/pull/39. I suppose we should only "enumerate" the scripts which make it into the final Combustion script.

jdob:
In this case, the XX_ was intentionally used because of the alphabetical sorting. We're gonna need the RPMs in place before the elemental register stuff kicks in.

But you're right that we need to go back through and evaluate all of them for this sort of thing. Danny and I were talking yesterday, I'm thinking something like 0X-5X are what we use, opening up 6X-8X for users (this will need to be documented, of course). That way the user has some room to play with their own custom scripts while also having hooks if they need to fit into a specific point in what we do.

pkg/build/rpm.go
```diff
@@ -1,12 +1,50 @@
...
+
+	rpmFileNames, err := getRPMFileNames(rpmSourceDir)
+	if err != nil {
+		return fmt.Errorf("getting rpm file names: %w", err)
+	}
+
+	err = copyRPMs(rpmSourceDir, rpmDestDir, rpmFileNames)
+	if err != nil {
+		return fmt.Errorf("copying RPMs over: %w", err)
```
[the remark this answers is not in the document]
dbw7: Very good point, was thinking about that as well, will update

pkg/build/rpm_test.go
```diff
@@ -117,8 +124,142 @@ func TestCopyRPMsNoRPMDir(t *testing.T) {
...
+
+	// Verify
+	require.NoError(t, err)
+}
+
+func TestWriteRPMScript(t *testing.T) {
+	// Setup
+	tmpDir, err := os.MkdirTemp("", "eib-")
+	require.NoError(t, err)
+	defer os.RemoveAll(tmpDir)
```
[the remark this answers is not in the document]
dbw7: Good catch, thank you

pkg/build/rpm_test.go
dbw7: Sounds good, will update.

pkg/build/rpm.go
dbw7: I mainly added this so that it's clear that there was a variable that held where the rpms should go in the end, but like you said we can save that line by omitting it, will update 👍

pkg/build/rpm.go
```diff
@@ -1,60 +1,116 @@
...
+func (b *Builder) writeRPMScript(RPMFileNames []string) error {
+	values := struct {
+		RPMs string
+	}{
+		RPMs: strings.Join(RPMFileNames, " "),
+	}
+
+	writtenFilename, err := b.writeCombustionFile(modifyRPMScriptName, modifyRPMScript, &values)
+	if err != nil {
+		return fmt.Errorf("writing RPM script %s: %w", modifyRPMScriptName, err)
```
atanasdinov, line 93: The caller already formats in the name of the script. We can simplify this and reduce the duplication:
```suggestion
		return fmt.Errorf("writing RPM script: %w", err)
```

pkg/build/rpm_test.go
```diff
@@ -12,113 +13,252 @@ import (
...
+
+	RPMSourceDir := filepath.Join(tmpDir, "rpms")
+	err = os.Mkdir(RPMSourceDir, 0o755)
+	require.NoError(t, err)
+
+	bc := config.BuildConfig{}
 	builder := New(nil, &bc)
-	err := builder.prepareBuildDir()
+	err = builder.prepareBuildDir()
+	require.NoError(t, err)
```
atanasdinov, line 27: Do we even need builder here?

pkg/build/rpm_test.go
```diff
@@ -12,113 +13,252 @@ import (
...
+	RPMSourceDir := filepath.Join(tmpDir, "rpms")
+	err = os.Mkdir(RPMSourceDir, 0o755)
 	require.NoError(t, err)
-	defer os.Remove(builder.eibBuildDir)
 
-	rpmSourceDir := filepath.Join(builder.buildConfig.ImageConfigDir, "rpms")
+	bc := config.BuildConfig{}
+	builder := New(nil, &bc)
+	err = builder.prepareBuildDir()
+	require.NoError(t, err)
```
atanasdinov, line 62: Same as above.

pkg/build/rpm_test.go
```diff
@@ -12,113 +13,252 @@ import (
...
+
+	RPMSourceDir := filepath.Join(tmpDir, "rpms")
+	err = os.Mkdir(RPMSourceDir, 0o755)
+	require.NoError(t, err)
+
+	bc := config.BuildConfig{}
 	builder := New(nil, &bc)
-	err := builder.prepareBuildDir()
+	err = builder.prepareBuildDir()
 	require.NoError(t, err)
```
atanasdinov, line 103: Same as above.

pkg/build/rpm_test.go
```diff
@@ -12,113 +13,252 @@ import (
...
 
-	file1Path := filepath.Join(rpmSourceDir, "rpm1.rpm")
-	defer os.Remove(file1Path)
+	bc := config.BuildConfig{}
+	builder := New(nil, &bc)
+	err = builder.prepareBuildDir()
+	require.NoError(t, err)
+
+	RPMSourceDir, err := builder.generateRPMPath()
+	require.NoError(t, err)
```
atanasdinov, line 126: Same as above.

pkg/build/rpm_test.go
```diff
@@ -12,113 +13,252 @@ import (
...
+
 	bc := config.BuildConfig{
-		ImageConfigDir: "../config/ThisDirDoesNotExist",
+		ImageConfigDir: tmpDir,
 	}
 	builder := New(nil, &bc)
-	err := builder.prepareBuildDir()
+
+	expectedPath := filepath.Join(builder.buildConfig.ImageConfigDir, "rpms")
```
atanasdinov, line 244: The path that we expect is `RPMSourceDir` - not the one that the builder internally builds. We should not confuse test expectations with how the builder is initialized.

pkg/build/rpm_test.go
```diff
@@ -12,113 +13,252 @@ import (
...
+	require.NoError(t, err)
+
 	bc := config.BuildConfig{
-		ImageConfigDir: "../config/testdata",
+		ImageConfigDir: tmpDir,
 	}
 	builder := New(nil, &bc)
-	err := builder.prepareBuildDir()
+	err = builder.prepareBuildDir()
 	require.NoError(t, err)
```
atanasdinov, line 196: Nit: All of this can be moved below to where it is actually used. Here we do it in the middle between creating the source dir and populating it with files.

pkg/build/rpm_test.go
```diff
@@ -12,113 +13,252 @@ import (
...
 
-	file2Path := filepath.Join(rpmSourceDir, "rpm2.rpm")
-	defer os.Remove(file2Path)
+	file2Path := filepath.Join(RPMSourceDir, "rpm2.rpm")
 	file2, err := os.Create(file2Path)
 	require.NoError(t, err)
+	defer file2.Close()
+
+	RPMFileNames, err := getRPMFileNames(RPMSourceDir)
+	require.NoError(t, err)
```
atanasdinov, line 160: Nit: The test aims to verify the logic of `writeRPMScript` and not of `getRPMFileNames` so we can just pass a pre-populated slice to `writeRPMScript`.

atanasdinov (commented):
A couple of issues with the tests. Happy to approve after those are resolved, linter issues are fixed (using camelCase for variable names) and the `/testdata/rpms` dir is dropped.

rdoxenham:
Just to ensure that we're all on the same page, I just pushed this (https://github.com/suse-edge/edge-image-builder/pull/38) as an example for how a local repo could be generated; the repo here is important as it can be simply added from Combustion:

```
zypper ar file:///dev/shm/combustion/config/repo airgap-repo
zypper --no-gpg-checks install -y --force-resolution --auto-agree-with-licenses <list of packages>
```

I know that @ipetrov117 is working on making that script more consumable via EIB but it's important that we're considering user-provided RPM's, along with a list of package names that can be pulled from SLE repo's as well as user-provided repo's. Therefore we need to think about this PR in the context of generating the entire repo, and installing an identified list of packages, rather than just rpm's placed into the `/rpms` directory.

pkg/build/rpm_test.go
```diff
@@ -83,48 +227,54 @@ func TestCopyRPMs(t *testing.T) {
...
 	// Verify
-	require.ErrorContains(t, err, "no rpms found")
+	require.NoError(t, err)
 
-	assert.Empty(t, rpmFileNames)
+	assert.Equal(t, expectedPath, generatedPath)
 }
 
-func TestCopyRPMsNoRPMDir(t *testing.T) {
+func TestGenerateRPMPathNoRPMDir(t *testing.T) {
```
atanasdinov, line 266: Does the test address the expectation that the name applies?

atanasdinov (commented):
A final touch would be to reduce the duplicate setup code in tests with something similar to the following:

```
func setupRPMSourceDir(t *testing.T) (string, func()) {
	tmpDir, err := os.MkdirTemp("", "eib-RPM-")
	require.NoError(t, err)

	rpmSourceDir := filepath.Join(tmpDir, "rpms")
	err = os.Mkdir(rpmSourceDir, 0o755)
	require.NoError(t, err)

	file1Path := filepath.Join(rpmSourceDir, "rpm1.rpm")
	file1, err := os.Create(file1Path)
	require.NoError(t, err)

	file2Path := filepath.Join(rpmSourceDir, "rpm2.rpm")
	file2, err := os.Create(file2Path)
	require.NoError(t, err)

	return tmpDir, func() {
		assert.NoError(t, file1.Close())
		assert.NoError(t, file2.Close())
		assert.NoError(t, os.RemoveAll(tmpDir))
	}
}
```

You can then use this in the various different test functions with:
```
rpmSourceDir, teardown := setupRPMSourceDir(t)
defer teardown()
```

pkg/build/rpm_test.go
dbw7: I thought so, but I updated it just to be sure by adding a temporary directory that doesn't contain `/rpms`

dbw7:
> A final touch would be to reduce the duplicate setup code in tests with something similar to the following:
> 
> ```
> func setupRPMSourceDir(t *testing.T) (string, func()) {
> 	tmpDir, err := os.MkdirTemp("", "eib-RPM-")
> 	require.NoError(t, err)
> 
> 	rpmSourceDir := filepath.Join(tmpDir, "rpms")
> 	err = os.Mkdir(rpmSourceDir, 0o755)
> 	require.NoError(t, err)
> 
> 	file1Path := filepath.Join(rpmSourceDir, "rpm1.rpm")
> 	file1, err := os.Create(file1Path)
> 	require.NoError(t, err)
> 
> 	file2Path := filepath.Join(rpmSourceDir, "rpm2.rpm")
> 	file2, err := os.Create(file2Path)
> 	require.NoError(t, err)
> 
> 	return tmpDir, func() {
> 		assert.NoError(t, file1.Close())
> 		assert.NoError(t, file2.Close())
> 		assert.NoError(t, os.RemoveAll(tmpDir))
> 	}
> }
> ```
> 
> You can then use this in the various different test functions with:
> 
> ```
> rpmSourceDir, teardown := setupRPMSourceDir(t)
> defer teardown()
> ```

Now this is awesome, thank you for writing this. I wasn't sure if it was a normal thing to make helper functions for tests so I avoided it, but I'm glad to see it's possible. Extra thanks as this allowed me to clean up the code significantly.

pkg/build/rpm_test.go
```diff
@@ -9,123 +10,194 @@ import (
...
-	file2Path := filepath.Join(rpmSourceDir, "rpm2.rpm")
-	defer os.Remove(file2Path)
-	file2, err := os.Create(file2Path)
+	// Verify
+	require.ErrorContains(t, err, "opening source file")
+}
+
+func TestWriteRPMScript(t *testing.T) {
+	// Setup
+	_, _, teardown := setupRPMSourceDir(t, true)
```
atanasdinov, line 119: We can merge after getting rid of this call.

Thanks for addressing all comments!

atanasdinov (approved):

jdob (approved):

PR #39, opened 2023-11-13
Working implementation of GRUB configuration for raw images

jdob:
We may have to revisit the actual sed commands once we more thoroughly test this, but this does currently work. Hopefully, those sorts of changes should have a minimal footprint and the rest of the structure used in this PR will remain.

We also may have to revisit how we get things into the modify-raw-image.sh template. That'll be affected by both the need to do this for manipulating the squashed raw image inside of an ISO (that's coming up soon on my todo list) as well as a refactoring if we end up jamming a lot into the guestfish command. But for now, this should be a reasonable start to build on and readdress when we have a clearer picture of what everything looks like.

### DIFF
--- a/pkg/build/raw_test.go
+++ b/pkg/build/raw_test.go
@@ -49,6 +49,9 @@ func TestWriteModifyScript(t *testing.T) {
 		Image: config.Image{
 			OutputImageName: "output-image",
 		},
+		OperatingSystem: config.OperatingSystem{
+			KernelArgs: []string{"alpha", "beta"},
+		},
 	}
 	buildConfig := config.BuildConfig{
 		ImageConfigDir: "config-dir",
@@ -74,6 +77,7 @@ func TestWriteModifyScript(t *testing.T) {
 	foundContents := string(foundBytes)
 	assert.Contains(t, foundContents, "guestfish --rw -a config-dir/output-image")
 	assert.Contains(t, foundContents, "copy-in "+builder.combustionDir)
+	assert.Contains(t, foundContents, "download /boot/grub2/grub.cfg /tmp/grub.cfg")
 }
 
 func TestCreateModifyCommand(t *testing.T) {
--- a/pkg/build/scripts/grub/guestfish-snippet.tpl
+++ b/pkg/build/scripts/grub/guestfish-snippet.tpl
@@ -1,0 +1,13 @@
+# Configure GRUB for first boot
+# - Without this, the values wouldn't be used until after the first time the
+#   grub configuration is regenerated
+download /boot/grub2/grub.cfg /tmp/grub.cfg
+! sed -i '/ignition.platform/ s/$/ {{.KernelArgs}} /' /tmp/grub.cfg
+upload /tmp/grub.cfg /boot/grub2/grub.cfg
+
+# Configure GRUB defaults
+# - Without this, when `transactional-update grub.cfg` is run it will overwrite
+#   settings used in the above change
+download /etc/default/grub /tmp/grub
+! sed -i '/^GRUB_CMDLINE_LINUX_DEFAULT="/ s/"$/ {{.KernelArgs}} "/' /tmp/grub
+upload /tmp/grub /etc/default/grub
--- a/pkg/build/scripts/modify-raw-image.sh.tpl
+++ b/pkg/build/scripts/modify-raw-image.sh.tpl
@@ -13,8 +13,10 @@ guestfish --rw -a {{.OutputImage}} -i <<'EOF'
   # Enables write access to the read only filesystem
   sh "btrfs property set / ro false"
 
-  # GRUB configuration (if specified)
-  {{.ConfigureGRUB}}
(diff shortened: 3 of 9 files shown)


pkg/build/scripts/grub/guestfish-snippet.tpl
```diff
@@ -0,0 +1,13 @@
+# Configure GRUB for first boot
+# - Without this, the values wouldn't be used until after the first time the
+#   grub configuration is regenerated
+download /boot/grub2/grub.cfg /tmp/grub.cfg
+! sed -i '/ignition.platform/ s/$/ {{.KernelArgs}} /' /tmp/grub.cfg
```
atanasdinov, line 5: I'd suggest to either add a TODO or open an issue to track a better way of customising the usage of the different args. We shouldn't be building on top of this.

pkg/build/scripts/modify-raw-image.sh.tpl
```diff
@@ -4,28 +4,28 @@ set -euo pipefail
...
-#  sh "btrfs property set / ro true"
-#  - Resets the filesystem to read only
+# Guestfish Command Documentation: https://libguestfs.org/guestfish.1.html
 
 guestfish --rw -a {{.OutputImage}} -i <<'EOF'
+  # Enables write access to the read only filesystem
   sh "btrfs property set / ro false"
+
+  # GRUB configuration (if specified)
+  {{.ConfigureGRUB}}
```
atanasdinov, line 17: What do you think about introducing a conditional to make it more clear? It should let us remove / reduce the comments in the Go code.

```suggestion
  {{ if ne .ConfigureGRUB "" }}
      {{ .ConfigureGRUB }}
  {{ end }}
```
Another option is just `{{ if .ConfigureGRUB }}`.

pkg/build/scripts/grub/guestfish-snippet.tpl
```diff
@@ -0,0 +1,13 @@
...
+upload /tmp/grub.cfg /boot/grub2/grub.cfg
+
+# Configure GRUB defaults
+# - Without this, when `transactional-update grub.cfg` is run it will overwrite
+#   settings used in the above change
+download /etc/default/grub /tmp/grub
+! sed -i '/^GRUB_CMDLINE_LINUX_DEFAULT="/ s/"$/ {{.KernelArgs}} "/' /tmp/grub
```
atanasdinov, line 12: Would this not replace the `GRUB_CMDLINE_LINUX_DEFAULT` part as well? I've seen either repetitions or the use of `\1` to keep the matched string but none of those apply here.

pkg/build/raw_test.go
```diff
@@ -74,6 +77,7 @@ func TestWriteModifyScript(t *testing.T) {
 	foundContents := string(foundBytes)
 	assert.Contains(t, foundContents, "guestfish --rw -a config-dir/output-image")
 	assert.Contains(t, foundContents, "copy-in "+builder.combustionDir)
+	assert.Contains(t, foundContents, "download /boot/grub2/grub.cfg /tmp/grub.cfg")
```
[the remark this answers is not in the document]
jdob: That's covered by the tests in grub_test.go itself. Those are verifying that the template is properly applied. For this test, my only goal is to prove that the grub stuff was called, figuring that if it didn't work, the other test specifically looking for that stuff would fail.

pkg/build/scripts/grub/guestfish-snippet.tpl
jdob: Sanity check me here, but it doesn't appear to. Here's the test I ran on the sed command:

`$ cat line-default
GRUB_CMDLINE_LINUX_DEFAULT="console=ttyS0,115200 console=tty0 security=selinux selinux=1 quiet net.ifnames=0 \$ignition_firstboot ignition.platform.id=metal"

$ sed '/^GRUB_CMDLINE_LINUX_DEFAULT="/ s/"$/ additional_option=xyz"/' ./line-default
GRUB_CMDLINE_LINUX_DEFAULT="console=ttyS0,115200 console=tty0 security=selinux selinux=1 quiet net.ifnames=0 \$ignition_firstboot ignition.platform.id=metal additional_option=xyz"
`

atanasdinov: Interesting, I didn't expect that but I'm not that good with sed in the first place. Thanks for the example!

diconico07: This sed command replaces the trailing `"` with ` {{.KernelArgs}} "`, my question is, are you sure you'll never have anything after the last `"` (i.e. no trailing spaces, no comments or anything) ?

jdob: As far as we know right now. The sed approach may go away in the future, but looking at the context of that line, there's nothing that could be after the ending "

atanasdinov (approved):

PR #40, opened 2023-11-14
Bootstrap fileio package

atanasdinov:
- Extracts common files writing and copying logic to a new fileio package
- Transfers and extends unit tests
- Removes unused business logic

### DIFF
--- a/pkg/fileio/file_io.go
+++ b/pkg/fileio/file_io.go
@@ -7,27 +7,30 @@ import (
 )
 
 func WriteFile(filename string, contents string, templateData any) error {
-	if templateData != nil {
-		tmpl, err := template.New(filename).Parse(contents)
-		if err != nil {
-			return fmt.Errorf("creating template for file %s: %w", filename, err)
+	if templateData == nil {
+		if err := os.WriteFile(filename, []byte(contents), os.ModePerm); err != nil {
+			return fmt.Errorf("writing file: %w", err)
 		}
 
-		file, err := os.Create(filename)
-		if err != nil {
-			return fmt.Errorf("creating file %s: %w", filename, err)
-		}
-		defer file.Close()
-
-		err = tmpl.Execute(file, templateData)
-		if err != nil {
-			return fmt.Errorf("applying the template at %s: %w", filename, err)
-		}
-	} else {
-		err := os.WriteFile(filename, []byte(contents), os.ModePerm)
-		if err != nil {
-			return fmt.Errorf("writing file %s: %w", filename, err)
-		}
+		return nil
 	}
+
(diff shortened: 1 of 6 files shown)


pkg/fileio/file_io.go
```diff
@@ -0,0 +1,61 @@
...
+	"fmt"
+	"io"
+	"os"
+	"text/template"
+)
+
+func WriteFile(filename string, contents string, templateData any) error {
+	if templateData == nil {
+		if err := os.WriteFile(filename, []byte(contents), os.ModePerm); err != nil {
+			return fmt.Errorf("writing file: %w", err)
```
jdob, line 13: I'm not familiar with what the output will be if this error happens, but I like to include the name of the file so there's some added context in the log message:

```
return fmt.Errorf("writing file %s: %w", filename, err)
```

pkg/fileio/file_io.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	tmpl, err := template.New(filename).Parse(contents)
+	if err != nil {
+		return fmt.Errorf("parsing template: %w", err)
+	}
+
+	file, err := os.Create(filename)
+	if err != nil {
+		return fmt.Errorf("creating file: %w", err)
```
jdob, line 26: If you do decide to add the filename to the previous error message, for consistency this should be updated too.

pkg/fileio/file_io.go
```diff
@@ -0,0 +1,61 @@
...
+		return fmt.Errorf("applying template: %w", err)
+	}
+
+	return nil
+}
+
+func CopyFile(src string, dest string) error {
+	sourceFile, err := os.Open(src)
+	if err != nil {
+		return fmt.Errorf("opening source file: %w", err)
```
jdob, line 42: Same as above for source and destination files here.

pkg/fileio/file_io_test.go
```diff
@@ -0,0 +1,144 @@
...
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestWriteFile(t *testing.T) {
+	const tmpDir = "write-test"
```
jdob, line 13: Nit: I prefer to prefix temp directories with some indication of who created them (so something like `eib-write-test-`), but that's just a personal style.

pkg/fileio/file_io.go
jdob: Leaving for completeness, but Atanas and I discussed this in chat and his approach is to leave it to the caller to mention the filename.

jdob (approved):
Nothing major, just some personal preference comments.

PR #42, opened 2023-11-15
Introduce Context type

atanasdinov:
- Introduces a new Context type
- Context stores all necessary information required by planned components structure
- Removing dir handling from Builder making it only responsible for actually building the image and allowing more robust testing (dropping the stateful `.prepareDir()` calls) and further combustion decoupling
- Drops BuildConfig type

### DIFF
--- a/pkg/build/dir_structure.go
+++ b/pkg/build/dir_structure.go
@@ -1,0 +1,50 @@
+package build
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+type DirStructure struct {
+	// ImageConfigDir is the root directory storing all configuration files.
+	ImageConfigDir string
+	// BuildDir is the directory used for assembling the different components used in a build.
+	BuildDir string
+	// CombustionDir is a subdirectory under BuildDir containing the Combustion script and all related files.
+	CombustionDir string
+	// DeleteBuildDir indicates whether the BuildDir should be cleaned up after the image is built.
+	DeleteBuildDir bool
+}
+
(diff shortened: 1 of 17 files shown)


pkg/build/dir_structure.go
```diff
@@ -0,0 +1,63 @@
+package build
```
atanasdinov, line 1: Not sure whether this is the right package for this. It's also why I kept all fields of the type exported for now. Thoughts?

pkg/build/dir_structure.go
```diff
@@ -0,0 +1,63 @@
...
+	if err := os.MkdirAll(combustionDir, os.ModePerm); err != nil {
+		return nil, fmt.Errorf("creating the combustion directory: %w", err)
+	}
+
+	return &DirStructure{
+		ImageConfigDir: imageConfigDir,
+		BuildDir:       buildDir,
+		CombustionDir:  combustionDir,
+		ScriptsDir:     filepath.Join(imageConfigDir, scriptsSubDir), // may not exist
+		RPMDir:         filepath.Join(imageConfigDir, rpmsSubDir),    // may not exist
```
atanasdinov, line 50: Added these two in order to avoid separate functions (and potentially future components) having to assemble them separately. Perhaps we could also add a check whether they exist instead of doing it in the `configureXX()` functions. Thoughts?

atanasdinov: These will be dropped and the type will be renamed to `Context` which the different components will use.

jdob (approved):

PR #43, opened 2023-11-16
Add example script to help build airgapped package repo

rdoxenham:
This script provides the ability to specify a list of local packages, along with any packages that you want to pull from SLE Micro (or additionally specified package repos) to generate a local package repository that contains all of the RPM's (and any identified dependencies) based on a base image provided by the user as the source, so it can be used within Combustion. It generates the file `/tmp/eib/repo/package-list.txt` with all of the packages that can be passed into `zypper in` at the Combustion stage.

### DIFF
--- a/docs/building-images.md
+++ b/docs/building-images.md
@@ -46,6 +46,12 @@ operatingSystem:
     password: 456
   - username: user3
     sshKey: user3Key
+  systemd:
+    enable:
+      - service0
+      - service1
+    disable:
+      - serviceX
 ```
 
 * `kernelArgs` - Optional; Provides a list of flags that should be passed to the kernel on boot.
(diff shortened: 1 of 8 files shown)


jdob (changes requested):
I appreciate the resource effort that went into this, but I don't see the value in checking this into the repo as an executable script. It's not something an EIB user or dev would ever need to run, especially once the important bits have been extracted out of this into EIB functionality. Arguably, putting it here as a script puts us on the hook to maintain it and keep feature parity with this script v. what EIB does internally.

If you want to commit this as research, I'd prefer to see it under `/docs` as a design doc with some context around it (the comments do a really good job of that, so mostly a note that says this was a POC to show how to do it).

PR #44, opened 2023-11-16
Added ability to create OS users

jdob:
This PR does not include checks to ensure at least one of password/SSH key are specified per user. We likely want that (I can't think of a use case for manually creating a user that can't be logged in that we'd want to support here), but that'll wait until the larger validation infra is in place for the entire image definition.


pkg/config/image_test.go
```diff
@@ -29,6 +29,18 @@ func TestParse(t *testing.T) {
 		"beta=bar",
 	}
 	assert.Equal(t, expectedKernelArgs, imageConfig.OperatingSystem.KernelArgs)
+
+	userConfigs := imageConfig.OperatingSystem.Users
+	assert.Len(t, userConfigs, 3)
```
atanasdinov, line 34: Should be `require` so that assertions are guaranteed to only be performed on a slice with the expected length.
```suggestion
	require.Len(t, userConfigs, 3)
```

pkg/build/scripts/users/add-users.sh.tpl
```diff
@@ -0,0 +1,34 @@
...
+  {{ end }}
+  {{ if .SSHKey }}
+    mkdir -pm700 /home/{{.Username}}/.ssh/
+    echo '{{.SSHKey}}' >> /home/{{.Username}}/.ssh/authorized_keys
+    chown -R {{.Username}} /home/{{.Username}}/.ssh
+  {{ end }}
+{{ end }}
+
+{{/* Root */}}
+{{ if (eq .Username "root") }}
```
atanasdinov, line 24: Is there a particular reason not to `{{ else }}` instead?

pkg/build/users_test.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/suse-edge/edge-image-builder/pkg/config"
+)
+
+func TestConfigureUsers(t *testing.T) {
+	// Setup
+	configData, err := os.ReadFile("./testdata/users-test-definition.yaml")
+	require.NoError(t, err)
+	imageConfig, err := config.Parse(configData)
```
atanasdinov, line 18: I'd suggest to drop parsing from these tests. What we want to verify here is that `configureUsers` is working as expected. Manually populating the config should be more than enough to cover these needs and decouple the config implementation from the resulting script assertions:

```
imageConfig = &config.ImageConfig{
   OperatingSystem: config.OperatingSystem{
	Users: []config.OperatingSystemUser{
		{
			Username: "alpha",
			Password: "alpha123",
			SSHKey:   "alphakey",
		},
		...
	},
   },
}
```

Note that I fully agree that we should have "integration" tests down the line involving fully fleshed YAML configs, however, here we're just testing the users unit.

pkg/build/scripts/users/add-users.sh.tpl
```diff
@@ -0,0 +1,34 @@
...
+mount /home
+
+{{ range . }}
+
+{{/* Non-root users */}}
+{{ if (ne .Username "root") }}
+  useradd -m {{.Username}}
+  {{ if .Password }}
+    echo '{{.Username}}:{{.Password}}' | chpasswd -e
+  {{ end }}
```
atanasdinov, line 15: If this is the same for both cases can we extract it down below in order to avoid the repetition?

atanasdinov (commented):
Looking great! Just a comment on the tests setup that I'd like to bring up.

pkg/config/image_test.go
jdob: Good pick up

pkg/build/scripts/users/add-users.sh.tpl
jdob: Nothing specific. My head was in a different place editing a template v. code so I just wasn't thinking of it as a if/else construct so much as a flag for inclusion (not sure if that made sense).

pkg/build/scripts/users/add-users.sh.tpl
jdob: IMO that makes it less readable, but that may be a personal distaste for the templating syntax. I like the model here where we have distinct section on root v. non-root and, IMO, that's worth the minimal duplication.

pkg/build/users_test.go
jdob: This makes sense and it'll be nice to remove the extra files. In hindsight, I took this approach because I wasn't sure how the parser would handle no OS section in the other test (would it be nil or still created) and then never rethought the approach.

atanasdinov (approved):

PR #48, opened 2023-11-28
Configure network

atanasdinov:
- Adds the option to configure network for various different hosts
- Introduces a helper function to generate paths to the optional components
- Follow ups should address:
  - Allowing cross target builds (currently not working since the same NMC artefact is used in both build and combustion phases)
  - Logging messages for each component
  - Unified log file name structure (reusing the timestamp of the build itself)
  - More test coverage (once validation is introduced in NMC and the EIB integration is stable)


pkg/combustion/network.go
```diff
@@ -0,0 +1,151 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("generating network path: %w", err)
+	}
+
+	if networkDir == "" {
+		// Networking is not configured, exit early
+		return nil, nil
+	}
+
+	zap.L().Info("Configuring network component...")
```
jdob, line 60: What do you think about moving this above the punch out check and then logging in there something like "No network configuration detected"?

It means the output is chattier since it'll show things that aren't relevant to the image in question, but it also makes it more clear what EIB is doing.

Alternatively, do we want the policy to be that we log to debug when we punch out early, so that way we can still see that's what happened but only for debugging cases?

jdob: BTW, I mean this as part of the general logging policy we were talking about for these sorts of Info messages.

pkg/combustion/network.go
```diff
@@ -0,0 +1,151 @@
...
+		return fmt.Errorf("running nmc generate: %w", err)
+	}
+
+	return nil
+}
+
+func generateNetworkLogFilename(ctx *image.Context) string {
+	const networkConfigLogFile = "network-config-%s.log"
+
+	timestamp := time.Now().Format("Jan02_15-04-05")
```
jdob, line 111: This uses the same format as the xorriso log file, which I like for consistency (I don't have particularly strong feelings right now about what the format looks like and didn't love it at the time). I know you mentioned an issue with constants, but part of me likes the idea of pulling this sort of thing out into a util--- UM, I mean, single timestamp format across EIB.

pkg/combustion/network_test.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/suse-edge/edge-image-builder/pkg/fileio"
+)
+
```
jdob, line 12: It's unfortunate there isn't a great model to test the Command. I worked around it in the xorriso stuff by returning the command separately from running it. I don't love that either, but IMO it was important to ensure the correct arguments were passed to the command.

Not saying you need to adopt that approach, but is there any reasonably clean way to be able to test the parameters to the generate command?

Dockerfile
```diff
@@ -16,6 +16,11 @@ RUN zypper install -y \
     xorriso  \
     libguestfs kernel-default e2fsprogs parted gptfdisk btrfsprogs
 
+# TODO: Install nmc via zypper once an RPM package is available
```
jdob, line 19: Can you add an issue to make sure we remember to readdress this? I'll start the discussion in chat about issue labels.

jdob (changes requested):
It's a little light on unit tests, but I understand the complications because it's essentially a wrapper on NMC calls. The only one I'd like to see before we merge this is the addition to combustion_test that verifies the behavior when the path isn't found.

pkg/combustion/network.go
atanasdinov: We should probably think about how we want to organise the build directory itself.

Aside from the combustion subdirectory, it'd have a lot of files flying around and once more and more optional components start adding their own .log (and other) files, then the directory itself would become fairly cluttered. I've already seen this where I have around ~10 `iso-build` logs and around ~5 `network-config` logs. And all runs actually seem to override the combustion subdirectory as well.

Perhaps having a `build-<timestamp>` subdirectory which stores all particular files of a single build makes sense. This way we probably would not need timestamps for each log file at all.

Once we figure this out we can look into what package structure to use to unify all of the relevant functionality. Thoughts?

Dockerfile
atanasdinov: We already track this internally in Jira. Do we want a Github issue for it as well?

pkg/combustion/combustion.go
```diff
@@ -50,3 +53,19 @@ func Configure(ctx *image.Context) error {
 
 	return nil
 }
+
+// generateComponentPath verifies whether a component directory exists under the root config directory.
+// Returns the full path to it if it exists e.g. `/config/rpms` or an empty string if not.
+func generateComponentPath(ctx *image.Context, componentDir string) (string, error) {
```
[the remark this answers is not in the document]
atanasdinov: Agree, I split it up.

pkg/combustion/network.go
jdob: I agree on there being no right/wrong answer. I also had a similar thought on the chattier, how it would implicitly teach the user about other options.

My vote is for chattier; we log each component as it's configured and always have some sort of closure to it, either "Finished configuring" or "Skipping since there is no configuration".

Dockerfile
atanasdinov: Raised an issue.

pkg/combustion/network_test.go
atanasdinov: Testing the args themselves is fine the way you did it. Mocking `exec.Command` is another way but it results in higher complexity in tests.

pkg/combustion/network.go
jdob: Yeah, I agree, we can pause on this particular point because it'll get picked up when we step back and look at the build directory as a larger revisit.

jdob (approved):
All nits and random comments, ready to land. Only real actionable item for now is my ask for an issue to revisit the hardcoded download URL.

pkg/network/config_generator.go
```diff
@@ -0,0 +1,29 @@
...
+	return nil
+}
+
+func generateCommand(configDir, outputDir string, output io.Writer) *exec.Cmd {
+	cmd := exec.Command("nmc", "generate",
+		"--config-dir", configDir,
+		"--output-dir", outputDir)
+
+	cmd.Stdout = output
+	cmd.Stderr = output
```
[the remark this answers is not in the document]
atanasdinov: It will definitely work for xorriso. Not sure what the issue you mention has been at the time.

pkg/combustion/network.go
```diff
@@ -0,0 +1,121 @@
...
+	defer func() {
+		if err = logFile.Close(); err != nil {
+			zap.L().Warn("Failed to close network log file properly", zap.Error(err))
+		}
+	}()
+
+	configDir := generateComponentPath(ctx, networkConfigDir)
+	outputDir := filepath.Join(ctx.CombustionDir, networkConfigDir)
+
+	return ctx.NetworkConfigGenerator.GenerateNetworkConfig(configDir, outputDir, logFile)
```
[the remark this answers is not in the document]
atanasdinov: We need to have the option for plugging in providers of different functionality. This way we're able to mock implementations which would not otherwise be easily integrated into tests.

Whether these "providers" live in a type on which we define methods or on an object we share across is a separate question. I can't judge which approach is better until we have more code.

pkg/combustion/combustion.go
```diff
@@ -50,3 +54,29 @@ func Configure(ctx *image.Context) error {
 
 	return nil
 }
+
+func generateComponentPath(ctx *image.Context, componentDir string) string {
+	return filepath.Join(ctx.ImageConfigDir, componentDir)
+}
+
+func isComponentConfigured(ctx *image.Context, componentDir string) bool {
```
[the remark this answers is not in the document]
atanasdinov: We can't have this in the build package because we'd run into circular dependency once again. Maybe a centralised "component" package makes sense for this? Probably in a follow up though.

pkg/network/configurator_installer.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+	defer file.Close()
+
+	if err = file.Chmod(fileio.ExecutablePerms); err != nil {
+		return fmt.Errorf("adjusting permissions: %w", err)
+	}
+
+	downloadURL := fmt.Sprintf(nmcDownloadURL, arch)
+
+	resp, err := http.Get(downloadURL) // nolint: gosec
```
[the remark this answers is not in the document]
atanasdinov: Usually `http.Get()` is not advised (as per the need to ignore the linter here). One would normally use [http.NewRequest](https://pkg.go.dev/net/http#NewRequest) and then pass that to an http client.

I was not convinced that this is the approach we'd be going for anyway so I left it at `http.Get()`. 

Nevertheless, as I've mentioned yesterday, a better idea is to just pre-install both binaries into the container image and use the appropriate one for the target OS image instead of downloading it every time. I pushed the implementation changes in order to reflect this idea.

pkg/network/configurator_installer.go
```diff
@@ -0,0 +1,57 @@
...
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"strings"
+
+	"github.com/suse-edge/edge-image-builder/pkg/fileio"
+)
+
+const nmcDownloadURL = "https://github.com/suse-edge/nm-configurator/releases/download/v0.2.0/nmc-linux-%s"
```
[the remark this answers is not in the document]
atanasdinov: No longer applicable as we now don't download the file during the build process.

In general, yes. It'd be appropriate to move it to the image definition where we parse it from pre-populated .conf file which the user has no access too. But as you've mentioned, this is too much of a hassle.

rdoxenham (approved):

jdob (approved):

Dockerfile
```diff
@@ -16,6 +16,15 @@ RUN zypper install -y \
...
+RUN curl -o nmc-aarch64 -L https://github.com/suse-edge/nm-configurator/releases/download/v0.2.0/nmc-linux-aarch64 && \
+    chmod +x nmc-aarch64
+
+RUN curl -o nmc-x86_64 -L https://github.com/suse-edge/nm-configurator/releases/download/v0.2.0/nmc-linux-x86_64 && \
+    chmod +x nmc-x86_64
+
+RUN cp nmc-$(uname -m) /usr/local/bin/nmc
```
jdob, line 26: Nit: Arguably, these should be a single RUN command to minimize the amount of layers that are created.

pkg/network/config_generator.go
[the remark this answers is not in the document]
jdob: Now that I think about it, this is the solution I went with (except to a log file instead of stdout/stderr). Originally I tried to capture the output using `cmd.Output()` and that didn't behave the way I expected.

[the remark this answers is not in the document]
atanasdinov: You were probably ignoring the output result when there is an error. This is one of the few occasions where a function can return both non-nil result and non-nil error. But then again, if you want to also capture stderr then `CombinedOutput()` is better.

PR #49, opened 2023-11-29
Basic infrastructure for user-facing audit messages

jdob:
Sample output with these changes (ignore my `eib-run` local script):

```
╭─[~/Code/eib]─[]─[]─●
╰─➤ eib-run raw-image-users.yaml
Rpm .......................... [SKIPPED]
Custom Scripts ............... [SUCCESS]
Building RAW image...
Kernel Params ................ [SKIPPED]
Image build complete!

╭─[~/Code/eib]─[]─[]─●
╰─➤ eib-run iso-image.yaml
Users ........................ [SKIPPED]
Rpm .......................... [SKIPPED]
Custom Scripts ............... [SUCCESS]
Building ISO image...
Image build complete!
```


jdob:
I could see the argument for moving the `Audit*` calls to combustion.go and using the returned error (or lack thereof) to have the loop print those messages using the `componentName`. One big issue with that is that it doesn't give us a clean hook for `SKIPPED`, but there may be a way to work around that that doesn't make the API into each component too ugly.

My rationale for this approach is that the component configuration functions themselves are the experts at knowing success v. failed v. skipped and should have full control over when and why those are logged, rather than having the loop in combustion.go do it (also keeping in mind that non-combustion customizations are going to log these messages, so that solution doesn't answer 100% of the cases). 

I'm certainly open to a discussion on it if anyone has a compelling reason to move the audit calls to the combustion loop.

pkg/log/audit.go
```diff
@@ -0,0 +1,50 @@
...
+func AuditComponentFailed(component string) {
+	message := formatComponentStatus(component, messageFailed)
+	Audit(message)
+}
+
+func formatComponentStatus(component, status string) string {
+	// Example output:
+	// Component ... [STATUS]
+
+	name := cases.Title(language.English).String(component)
```
atanasdinov, line 44: A slight inconvenience is that this will not work properly for acronyms e.g. `RPM` is translated to `Rpm` here.

pkg/log/audit_test.go
```diff
@@ -0,0 +1,44 @@
...
+func TestFormatComponentStatus(t *testing.T) {
+	// Tests
+	tests := []struct {
+		testName  string
+		component string
+		status    string
+		expected  string
+	}{
+		{
+			testName:  "Success test",
```
atanasdinov, line 18: Nit: Since these are all tests we don't need to include the word "test" in the names. Something cleaner could be "Successful component" for example.

pkg/combustion/rpm.go
```diff
@@ -22,28 +24,34 @@ var modifyRPMScript string
...
 
 	err = copyRPMs(rpmSourceDir, ctx.CombustionDir, rpmFileNames)
 	if err != nil {
+		log.AuditComponentFailed(rpmComponentName)
 		return nil, fmt.Errorf("copying RPMs over: %w", err)
 	}
 
 	script, err := writeRPMScript(ctx, rpmFileNames)
 	if err != nil {
+		log.AuditComponentFailed(rpmComponentName)
```
atanasdinov, line 50: The repetition is fine for now. Maybe in the future we can extract a `logComponentStatus()` function and defer it for each component. It would check whether the function returns a non-nil error (by naming the returned value) and then either log success or failure.

pkg/build/build.go
```diff
@@ -23,19 +24,32 @@ func New(ctx *image.Context) *Builder {
...
 	if err := b.configureCombustion(b.context); err != nil {
 		return fmt.Errorf("configuring combustion: %w", err)
 	}
 
 	switch b.context.ImageDefinition.Image.ImageType {
 	case image.TypeISO:
-		return b.buildIsoImage()
+		log.Audit("Building ISO image...")
+		if err := b.buildIsoImage(); err != nil {
+			log.Audit("Error building ISO image")
```
atanasdinov, line 37: Just throwing out an idea for this one, let me know if you disagree. The message can be more descriptive e.g. "Building ISO image failed. Check build log file for details"

atanasdinov:
One way to achieve this would be to make the component function either return a third argument (skipped boolean) or introduce a new type (e.g. componentResult) containing the possible outcomes of each configure function.

However, I'm hesitant to suggest either of these approaches until we have a somewhat stable code base where most of the initially planned features are implemented. We still need to introduce many more components and the repo architecture will continue evolving. I vote for keeping the audit calls within the configuring functions at least for now.

atanasdinov (approved):
Formatting looks great!

What we're missing is audit logs for the message component. Feel free to merge once this is addressed, rest of the comments are nitpicks.

pkg/log/audit.go
```diff
@@ -0,0 +1,50 @@
...
+
+	messageSuccess = "SUCCESS"
+	messageSkipped = "SKIPPED"
+	messageFailed  = "FAILED " // leave the trailing space for consistent lengths
+)
+
+// Audit displays a message to the user. This shouldn't be used for debug logging purposes; all
+// messages passed in here should be user-readable.
+func Audit(message string) {
+	fmt.Println(message)
```
atanasdinov, line 22: What do you think about using the log package from the standard library? It includes the timestamp in a user friendly manner:
```
2023/11/30 09:27:04 Generating image customization components...
2023/11/30 09:27:04 Users ........................ [SUCCESS]
2023/11/30 09:27:04 Rpm .......................... [SKIPPED]
2023/11/30 09:27:04 Custom Scripts ............... [SKIPPED]
2023/11/30 09:27:04 Building ISO image...
2023/11/30 09:27:08 Image build complete!
```

jdob:
> One way to achieve this would be to make the component function either return a third argument (skipped boolean) or introduce a new type (e.g. componentResult) containing the possible outcomes of each configure function.
> 
> However, I'm hesitant to suggest either of these approaches until we have a somewhat stable code base where most of the initially planned features are implemented. We still need to introduce many more components and the repo architecture will continue evolving. I vote for keeping the audit calls within the configuring functions at least for now.

100% agree, I have the same hesitation that we'd be chasing an API that isn't quite clear to us by adding all of these sorts of things. We may refactor this out when we have a more clear view of all of the different things happening.

pkg/log/audit.go
jdob: Generally speaking, I don't have strong opinions of how this function does it's thing. The real motivation behind this PR is that we have all of the code doing the same thing. Then we can dork with the details of how this function is implemented all we want.

I'll take a look at the log package. I'm certainly not hell bent on using `Println` and figured that we'd eventually make this at very least configurable to write to different options.

pkg/log/audit.go
jdob: I noticed, it kinda drove me a bit nuts. I thought about adding an optional flag to "skip formatting" the component but felt like overkill.

pkg/log/audit_test.go
jdob: Ahhh, good to know. I'm still not familiar enough with the table testing output to have picked this up on my own, so thanks for pointing it out.

pkg/build/build.go
jdob: This was actually my original intention but I completely forgot when it came to do it. Good call.

PR #50, opened 2023-11-30
Write logs to file instead of stdout

atanasdinov:


cmd/eib/main.go
```diff
@@ -54,29 +54,34 @@ func processArgs() (*image.Context, error) {
...
-		EncoderConfig: encoderCfg,
-		OutputPaths: []string{
-			"stdout",
-		},
-		ErrorOutputPaths: []string{
-			"stderr",
-		},
-	}
+	logConfig := zap.NewProductionConfig()
+	logConfig.Level = zap.NewAtomicLevelAt(logLevel)
```
atanasdinov, line 80: Should we drop the flag and use DEBUG level as a default with the introduction of user facing audits in https://github.com/suse-edge/edge-image-builder/pull/49?

cmd/eib/main.go
```diff
@@ -113,7 +118,8 @@ func validateImageConfigDir(configDir string) error {
 func main() {
 	ctx, err := processArgs()
 	if err != nil {
-		zap.L().Fatal("CLI arguments could not be parsed", zap.Error(err))
+		// use standard logger, zap is not yet configured
+		log.Fatalf("CLI arguments could not be parsed: %s", err)
```
jdob, line 122: Damn, good catch

jdob (approved):
I'm good with this landing as-is if you want to address removing -verbose and the debug v. info in a follow up.

PR #53, opened 2023-12-05
Added support for registering with Elemental

jdob:


pkg/combustion/combustion.go
```diff
@@ -25,18 +25,43 @@ type configureComponent func(context *image.Context) ([]string, error)
...
+	// - Message has no effect on the system, so this can go anywhere
+	// - Custom scripts should be early to allow the most flexibility in the user
+	//   being able to override/preempt the built-in behavior
+	// - Elemental must come after RPMs since (as of Dec 2023) the user must provide the
+	//   elemental RPMs manually
+	type componentWrapper struct {
+		name     string
+		runnable configureComponent
+	}
+	combustionComponents := []componentWrapper{
```
atanasdinov, line 38: Network component is gone, probably it went missing while fixing the conflicts.

pkg/combustion/templates/script-base.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
+# combustion: network
```
atanasdinov, line 2: Do we need this? I'm wondering whether it affects the network component but I can test this at a later point.

pkg/combustion/elemental.go
```diff
@@ -0,0 +1,88 @@
...
+		Elemental image.Elemental
+	}
+	yamlData, err := yaml.Marshal(ElementalWrapper{
+		Elemental: ctx.ImageDefinition.Elemental,
+	})
+	if err != nil {
+		return fmt.Errorf("extracting elemental config: %w", err)
+	}
+
+	if err := os.WriteFile(configFilename, yamlData, fileio.ExecutablePerms); err != nil {
```
atanasdinov, line 64: Correct me if I'm mistaken but this shouldn't require executable perms?

pkg/combustion/elemental.go
```diff
@@ -0,0 +1,88 @@
+package combustion
```
e-minguez, line 1: Shouldn't this check if the rpms are in place as well? (maybe it is and I couldn't find it)

pkg/combustion/elemental.go
```diff
@@ -0,0 +1,88 @@
...
+	elementalComponentName = "elemental"
+	elementalScriptName    = "11-elemental.sh"
+	elementalConfigName    = "elemental_config.yaml"
+)
+
+//go:embed templates/11-elemental-register.sh.tpl
+var elementalScript string
+
+func configureElemental(ctx *image.Context) ([]string, error) {
+	// Even if the "elemental" section is left out of the definition, the full structure will
```
atanasdinov, line 26: What do we mean here? I get that the criteria for the configuration is a non-empty registration URL (until a validation step is implemented), however we don't (and also shouldn't) modify any of the section values.

docs/building-images.md
```diff
@@ -57,6 +57,34 @@ operatingSystem:
...
+
+* `url` - Required; Endpoint URL for registration.
+* `ca-cert` - Required; Certificate for TLS verification.
+
+For information on the TPM configuration, see the [Elemental MachineRegistration documentation](https://elemental.docs.rancher.com/machineregistration-reference/#configelementalregistration).
+
+Additionally, the following RPMs must be included in the RPMs directory as described in the
+Image Configuration Directory section:
+* `elemental-register`
+* `elemental-system-agent`
```
atanasdinov, line 86: I guess this is fine for the initial implementation but I definitely don't feel it makes it a great user experience. Probably pre-installing and detecting image arch (similarly to nmc) would be a much preferred approach but I'd leave this for PM to decide.

pkg/combustion/templates/11-elemental-register.sh.tpl
```diff
@@ -0,0 +1,35 @@
...
+Alias=elemental-system-agent.service
+
+[Service]
+Type=simple
+Restart=always
+RestartSec=5s
+StandardOutput=journal+console
+StandardError=journal+console
+Environment="CATTLE_AGENT_CONFIG=/etc/rancher/elemental/agent/config.yaml"
+ExecStartPre=/bin/sh -c "mkdir -p /var/lib/elemental/agent && cp /etc/rancher/elemental/agent/elemental_connection.json /var/lib/elemental/agent"
```
e-minguez, line 32: It seems you can specify the local plan folder instead of doing this https://github.com/rancher/system-agent/blob/57830e05ec62b3ac10b46b8b9328402f0e30348e/pkg/config/config.go#L17 https://github.com/rancher/system-agent/blob/57830e05ec62b3ac10b46b8b9328402f0e30348e/examples/configuration/README.md?plain=1#L9

pkg/combustion/elemental.go
atanasdinov: You're correct. We can't fail during build time if the components are missing and we need the validation.

atanasdinov (changes requested):
Main issues to resolve before we merge this:
- network component is missing from the combustion flow
- validation that RPM packages is not present

pkg/combustion/elemental.go
jdob: Totally agree we should check. Problem is, it's not that simple. Here are the options I see:

= Option #1: Force the user to name the files according to our convention, so something like `elemental-toolkit.rpm` =
Pros:
- Makes validation simple
- Continues to use the existing RPM install functionality at no extra cost to us
Cons:
- Adds another manual step for the user to do
- Removes version/arch/whatever else information from the RPM name
- Relies on documentation to convey the convention, as compared to Option #3 which is more explicit

= Option #2: Have the validation logic only check for the prefix of the RPM, making sure there are RPMs that start with `elemental-toolkit` and `elemental-system-agent`
Pros:
- Continues to use the existing RPM install functionality at no extra cost to us
- User shouldn't need to do any extra renaming steps in most cases
Cons:
- Validation is a bit on the fragile side
- Relies on documentation to convey the convention, as compared to Option #3 which is more explicit

= Option #3: Have the user specify the names of these files in the image definition, and still adding =
Pros:
- Can still use the existing RPM functionality; the image just won't be built if EIB can't find them by name
Cons:
- Extra data needs to be added to the image definition. _(jdob: I don't hate that; the magic of just dropping it in `rpms` is nice but there is something to be said about having the elemental pieces not work on magic)_

= Option #4: Have the user create an `elemental` directory in the image config dir and include the RPMs in there
Pros:
- Arguably clean, since it makes it obvious what those RPMs are in there for as compared to the catch-all `rpms` directory
- Doesn't use the existing RPM functionality at all, instead injecting the names of the RPMs into the template for elemental registration, thus removing the cross-component dependency
Cons:
- Still need to figure out how to resolve the potential name differences (see Options 1 and 2)
- Potentially confusing why there would be RPMs outside of the `rpms` directory _(jdob: I don't think this is a huge issue)_

Thoughts?

pkg/combustion/templates/11-elemental-register.sh.tpl
jdob: Is there a benefit in changing the approach? Admittedly I copied your scripts largely as is, so I'm all for you making the decision on which approach we take.

pkg/combustion/templates/script-base.sh
jdob: We absolutely need this and is the big blocker that took so much of Danny's time. Without it, it just straight up doesn't work.

pkg/combustion/elemental.go
jdob: You're right, I'll update.

pkg/combustion/elemental.go
jdob: Ok, maybe I need to reword this. It was an explanation for developers potentially not familiar with how the yaml parsing works.

My gut reaction was to check if `Registration` or `Elemental` were nil. But the way the parser works, it will create them with the default values. So it was meant to explain the rationale for checking a specific value instead of just the definition structure itself. Perhaps it's overkill though; maybe it's obvious to new devs that checking for the URL makes sense as the logic.

docs/building-images.md
jdob: I agree, I don't love it either. My concern about embedding it into the EIB container image is that it locks the user into a specific version. To work around that, we'd need to provide a way to override which would look similar to this, so it was a judgement call based on lack of time.

atanasdinov (approved):

pkg/combustion/elemental.go
atanasdinov: Thanks for the comprehensive and thoughtful considerations! I really prefer option 5 but it also implies that we know exactly where to grab the RPMs from which has to be further investigated.

Approving for now and we can introduce the validation in a follow up.

PR #59, opened 2023-12-11
Adding support for SUMA Registration

rdoxenham:


pkg/combustion/templates/12-suma-register.sh.tpl
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+set -euo pipefail
+
+mkdir -p /etc/venv-salt-minion/
+
+{{ if .Get_SSL }}
```
jdob: This brings up an interesting 3.1 discussion about if we want air-gapped as a definition parameter, which would let us have a validation step that tells the user if they enable something that isn't compatible with air-gapped.

docs/building-images.md
```diff
@@ -120,3 +120,19 @@ as follows:
 Additionally, the following RPMs must be included in the RPMs directory as described above:
 * `elemental-register`
 * `elemental-system-agent`
+
+### SUSE Manager (SUMA)
```
jdob, line 124: Love that you added docs. Only request is to move this up under the Image Definition section above (after Operating Systems), since these are definition items (as compared to the directory structure items).

docs/building-images.md
```diff
@@ -120,3 +120,19 @@ as follows:
...
+
+Automatic SUSE Manager registration can be configured for the image. The following items must be defined in the
+configuration file under the `suma` section:
+
+* `host` - this is the FQDN of the SUSE Manager host that the host needs to register against.
```
jdob, line 129: Very small nit: the rest of the docs capitalize the first letter of the description.

docs/building-images.md
```diff
@@ -120,3 +120,19 @@ as follows:
...
+* `activation_key` - this is the activation key that the node uses to register with.
+* `get_ssl` - this specifies whether EIB should download and install the SUMA SSL Certificate (default: false)
+
+The default SSL certificate for the SUSE Manager server can usually be found at
+https://<suma-host>/pub/RHN-ORG-TRUSTED-SSL-CERT
```
jdob, line 134: Nit: Consider marking URLs as monospaced. But I admit that's very much a personal preference.

jdob (changes requested):
Looks really good. Only change is, oddly enough, to the docs. And it's simply moving the SUMA stuff under the right section.

docs/building-images.md
rdoxenham: Good call, I didn't even notice the structure. Will fix.

jdob:
LGTM. I'll leave it to Atanas to add the final approval (I trust his Go more than mine).

pkg/combustion/suma_test.go
```diff
@@ -0,0 +1,75 @@
...
+func TestConfigureSuma_NoConf(t *testing.T) {
+	// Setup
+	ctx, teardown := setupContext(t)
+	defer teardown()
+
+	ctx.ImageDefinition = &image.Definition{
+		OperatingSystem: image.OperatingSystem{
+			Suma: image.Suma{},
+		},
+	}
```
atanasdinov, line 23: You can drop this altogether and the test will still pass since `setupContext` will initialize a non-nil ImageDefinition pointer.

docs/building-images.md
```diff
@@ -67,6 +67,22 @@ operatingSystem:
...
   * `disable` - Optional; List of systemd services to disable.
 
+## SUSE Manager (SUMA)
+
+Automatic SUSE Manager registration can be configured for the image. The following items must be defined in the
+configuration file under the `suma` section:
+
+* `host` - This is the FQDN of the SUSE Manager host that the host needs to register against.
+* `activation_key` - This is the activation key that the node uses to register with.
+* `get_ssl` - This specifies whether EIB should download and install the SUMA SSL Certificate (default: false)
```
atanasdinov, line 77: The only "issue" I find is that both of these are snake_cased which differs from the camelCase we've used for everything else.

atanasdinov (approved):
Congratulations on your first EIB contribution - nice work!

jdob (approved):

PR #74, opened 2024-01-02
SUMA: venv-salt-minion needs to be enabled not restarted

rdoxenham:
In the previous version, at bootup the service was restarted, which wouldn't work properly during combustion, it simply needed to be enabled.

### DIFF
--- a/pkg/combustion/templates/12-suma-register.sh.tpl
+++ b/pkg/combustion/templates/12-suma-register.sh.tpl
@@ -21,4 +21,4 @@ enable_fqdns_grains: False
 
 EOF
 
-systemctl restart venv-salt-minion || true
+systemctl enable venv-salt-minion || true


atanasdinov (approved):

pkg/combustion/templates/12-suma-register.sh.tpl
```diff
@@ -21,4 +21,4 @@ enable_fqdns_grains: False
 
 EOF
 
-systemctl restart venv-salt-minion || true
+systemctl enable venv-salt-minion || true
```
Kristian-ZH, line 24: Do we need `|| true` now?

rdoxenham: @Kristian-ZH arguably not... just playing it safe with this as we don't really want the installation to completely fail if the venv-salt-minion can't start.

jdob (approved):

Issue #75, opened 2024-01-02
Image type in yaml definition file to be case insensitive


dbw7:
```
apiVersion: 1.0
image:
  imageType: iso
  baseImage: slemicro.iso
  outputImageName: eibimage.iso
  ```
Currently, `imageType` is case sensitive, it should be updated to be case insensitive.

dbw7:
Closed by #76

PR #76, opened 2024-01-03
added conversion to lowercase

dbw7:
In reference to #75

### DIFF
--- a/pkg/image/definition.go
+++ b/pkg/image/definition.go
@@ -2,6 +2,7 @@ package image
 
 import (
 	"fmt"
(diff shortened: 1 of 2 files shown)


jdob:
This isn't exactly the best way to do this, but can you change full-valid-example.yaml to have ISO in all caps and have the unit test verify it's lowercase?

jdob (changes requested):
See my comment in the conversation about how we can use the existing valid test to show this.

jdob (approved):

PR #77, opened 2024-01-03
Validation

dbw7:
This is a preliminary version of the image validation handling. There’s still much to discuss and a lot to get feedback on. Additionally, it currently fails the linter and has no tests but this will be updated soon.

### DIFF
--- a/pkg/image/definition.go
+++ b/pkg/image/definition.go
@@ -53,9 +53,5 @@ func ParseDefinition(data []byte) (*Definition, error) {
 	if err := yaml.Unmarshal(data, &definition); err != nil {
 		return nil, fmt.Errorf("could not parse the image definition: %w", err)
 	}
-	err := validate(&definition)
-	if err != nil {
-		return nil, fmt.Errorf("error validating definition file: %w", err)
-	}
 	return &definition, nil
 }
--- a/pkg/image/definition.go
+++ b/pkg/image/definition.go
@@ -53,6 +53,9 @@ func ParseDefinition(data []byte) (*Definition, error) {
 	if err := yaml.Unmarshal(data, &definition); err != nil {
 		return nil, fmt.Errorf("could not parse the image definition: %w", err)
 	}
-
+	err := validate(&definition)
+	if err != nil {
+		return nil, fmt.Errorf("error validating definition file: %w", err)
+	}
 	return &definition, nil
 }
--- a/pkg/image/definition.go
+++ b/pkg/image/definition.go
@@ -55,6 +55,6 @@ func ParseDefinition(data []byte) (*Definition, error) {
 		return nil, fmt.Errorf("could not parse the image definition: %w", err)
 	}
 	definition.Image.ImageType = strings.ToLower(definition.Image.ImageType)
-  
+
 	return &definition, nil
 }
--- a/pkg/image/validation.go
+++ b/pkg/image/validation.go
@@ -5,12 +5,8 @@ import (
 	"strings"
 )
 
-func validate(definition *Definition) error {
-	err := validateAPIVersion(definition)
-	if err != nil {
-		return fmt.Errorf("error validating apiVersion: %w", err)
-	}
-	err = validateImage(definition)
+func ValidateDefinition(definition *Definition) error {
+	err := validateImage(definition)
 	if err != nil {
 		return fmt.Errorf("error validating image: %w", err)
 	}
@@ -22,29 +18,19 @@ func validate(definition *Definition) error {
 	return nil
 }
 
-func validateAPIVersion(definition *Definition) error {
-	// Open to discussion regarding how to validate the else if
-	if definition.APIVersion == (Definition{}.APIVersion) {
-		return fmt.Errorf("apiVersion not defined")
-	} else if len(definition.APIVersion) < 2 {
-		return fmt.Errorf("invalid apiVersion")
-	}
-	return nil
-}
-
 func validateImage(definition *Definition) error {
 	if definition.Image == (Image{}) {
 		return fmt.Errorf("image not defined")
 	}
-	if definition.Image.ImageType == (Definition{}.Image.ImageType) {
+	if definition.Image.ImageType == "" {
 		return fmt.Errorf("imageType not defined")
 	} else if definition.Image.ImageType != "iso" && definition.Image.ImageType != "raw" {
 		return fmt.Errorf("invalid imageType, should be 'iso' or 'raw'")
 	}
-	if definition.Image.BaseImage == (Definition{}.Image.BaseImage) {
+	if definition.Image.BaseImage == "" {
 		return fmt.Errorf("baseImage not defined")
 	}
-	if definition.Image.OutputImageName == (Definition{}.Image.OutputImageName) {
+	if definition.Image.OutputImageName == "" {
 		return fmt.Errorf("outputImageName not defined")
 	}
 	return nil
@@ -54,7 +40,7 @@ func validateOperatingSystem(definition *Definition) error {
 	if checkIfOperatingSystemDefined(&definition.OperatingSystem) {
 		return nil
 	}
-	err := validateKernalArgs(&definition.OperatingSystem)
+	err := validateKernelArgs(&definition.OperatingSystem)
 	if err != nil {
 		return fmt.Errorf("error validating kernal args: %w", err)
 	}
@@ -66,6 +52,10 @@ func validateOperatingSystem(definition *Definition) error {
 	if err != nil {
 		return fmt.Errorf("error validating users: %w", err)
 	}
+	err = validateSuma(&definition.OperatingSystem)
+	if err != nil {
+		return fmt.Errorf("error validating suma: %w", err)
+	}
 	return nil
 }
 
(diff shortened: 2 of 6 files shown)


pkg/image/definition.go
```diff
@@ -53,6 +53,9 @@ func ParseDefinition(data []byte) (*Definition, error) {
 	if err := yaml.Unmarshal(data, &definition); err != nil {
 		return nil, fmt.Errorf("could not parse the image definition: %w", err)
 	}
-
+	err := validate(&definition)
```
atanasdinov, line 56: I'd extract this out of here and instead export a `ValidateDefinition` function. This way we separate the functionality and concerns.

pkg/image/validation.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+	err = validateOperatingSystem(definition)
+	if err != nil {
+		return fmt.Errorf("error validating operating system: %w", err)
+	}
+
+	return nil
+}
+
+func validateAPIVersion(definition *Definition) error {
```
atanasdinov, line 25: We might drop this for now until we have a clear usage for this field.

pkg/image/validation.go
```diff
@@ -0,0 +1,153 @@
...
+		return fmt.Errorf("invalid apiVersion")
+	}
+	return nil
+}
+
+func validateImage(definition *Definition) error {
+	if definition.Image == (Image{}) {
+		return fmt.Errorf("image not defined")
+	}
+	if definition.Image.ImageType == (Definition{}.Image.ImageType) {
```
atanasdinov, line 39: We don't need to create zero-valued structs just for the comparison purposes. The zero value of the most inner field (in this case `string`) is just enough to compare. Applies to all other checks below as well.

```suggestion
	if definition.Image.ImageType ==  "") {
```

pkg/image/validation.go
```diff
@@ -0,0 +1,153 @@
...
+	if definition.APIVersion == (Definition{}.APIVersion) {
+		return fmt.Errorf("apiVersion not defined")
+	} else if len(definition.APIVersion) < 2 {
+		return fmt.Errorf("invalid apiVersion")
+	}
+	return nil
+}
+
+func validateImage(definition *Definition) error {
+	if definition.Image == (Image{}) {
```
atanasdinov, line 36: This is fine and will work for the time being. There's always the opportunity to introduce better checks by using pointers to structs instead of structs but I wouldn't spend the time on such enhancements right now.

pkg/image/validation.go
```diff
@@ -0,0 +1,153 @@
...
+	if err != nil {
+		return fmt.Errorf("error validating users: %w", err)
+	}
+	return nil
+}
+
+func checkIfOperatingSystemDefined(os *OperatingSystem) bool {
+	return len(os.KernelArgs) == 0 &&
+		len(os.Users) == 0 &&
+		len(os.Systemd.Enable) == 0 && len(os.Systemd.Disable) == 0 &&
```
atanasdinov, line 75:
```suggestion
	os.Systemd == (Systemd{}) &&
```

pkg/image/validation.go
```diff
@@ -0,0 +1,153 @@
...
+		os.Suma.Host == "" && os.Suma.ActivationKey == "" && !os.Suma.GetSSL
```
atanasdinov, line 76:
```suggestion
		os.Suma == (Suma{})
```

pkg/image/validation.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+func validateKernalArgs(os *OperatingSystem) error {
```
atanasdinov, line 79:
```suggestion
func validateKernelArgs(os *OperatingSystem) error {
```

pkg/image/validation.go
```diff
@@ -0,0 +1,153 @@
...
+	seenKeys := make(map[string]bool)
+
+	for _, arg := range os.KernelArgs {
+		parts := strings.SplitN(arg, "=", 2)
```
atanasdinov, line 83: Could we please verify if this is always the case for argument layouts? I'm almost fairly certain that there are args which can be just "keys" but I might be wrong.

pkg/image/validation.go
```diff
@@ -0,0 +1,153 @@
...
+			return fmt.Errorf("invalid kernel arg: %s, expected format key=value", arg)
+		}
+
+		key, value := parts[0], parts[1]
+		if value == "" {
+			return fmt.Errorf("kernel arg '%s' has no value", key)
+		}
+
+		if _, exists := seenKeys[key]; exists {
+			return fmt.Errorf("duplicate kernel arg found: %s", key)
```
atanasdinov, line 94: It's nice that we have such check in place.

pkg/image/validation.go
```diff
@@ -0,0 +1,153 @@
...
+			if enableItem == disableItem {
+				return fmt.Errorf("conflict found: '%s' is both enabled and disabled", enableItem)
+			}
+		}
+	}
+
+	return nil
+}
+
+func checkForDuplicates(items []string, listType string) error {
```
atanasdinov, line 122: I'd suggest to remove the `listType` argument and use error formatting in the callers of this function instead.

atanasdinov (commented):
Looks good as a first version!

I'd suggest to also include some basic validation for SUMA (for empty host and activation keys, @rdoxenham might have better ideas) but overall this is a good structure.

At a later point we can always improve by using pointers for the struct values but this is not something I'd recommend doing right now. It's an enhancement that might lead us down a path where the business logic requires too many nil checks in order to avoid panics that we wouldn't run into without pointers.

pkg/image/validation.go
dbw7: For the Systemd comparison, there's this error with the suggested implementation: 
`Invalid operation: os.Systemd == (Systemd{}) (the operator == is not defined on Systemd)` 
So I'll keep the current implementation for now until a better option comes along.

pkg/image/validation.go
dbw7: For my next commit containing tests, I'm gonna operate under the "key=value" assumption and then just change it later if necessary.

rdoxenham: Can confirm, some kernel cmdline arguments are just parameters, without value. See my kernel cmdline for example, where it has "quiet" as a parameter/key, with no required value:

```
rdo@eostre:~% cat /proc/cmdline
BOOT_IMAGE=/boot/vmlinuz-6.6.7-1-default root=UUID=8cfd88e7-a472-4edf-8845-fc7a28b611d5 splash=silent resume=/dev/disk/by-uuid/06b2a538-d922-4c75-a5be-ec0660a185fd mitigations=auto quiet security=selinux selinux=1 enforcing=1
```

pkg/image/validation.go
rdoxenham: Yeah, I'm also a little conflicted on this. Kernel cmdline arguments take the last set values when there's duplicates, but I think it's good that we have a check just incase the customer specifies it twice and wants to be *absolutely* sure that they're setting it correctly.

pkg/image/validation.go
atanasdinov: Ah, yeah.. because slices are not comparable directly with each other and `Enable` and `Disable` are both slices.

pkg/image/validation.go
```diff
@@ -0,0 +1,168 @@
...
+	return nil
+}
+
+func validateOperatingSystem(definition *Definition) error {
+	if checkIfOperatingSystemDefined(&definition.OperatingSystem) {
+		return nil
+	}
+	err := validateKernelArgs(&definition.OperatingSystem)
+	if err != nil {
+		return fmt.Errorf("error validating kernal args: %w", err)
```
atanasdinov, line 45:
```suggestion
		return fmt.Errorf("error validating kernel args: %w", err)
```

pkg/image/validation.go
```diff
@@ -0,0 +1,168 @@
...
+			}
+			if value == "" {
+				return fmt.Errorf("kernel arg '%s' has no value", key)
+			}
+
+			if _, exists := seenKeys[key]; exists {
+				return fmt.Errorf("duplicate kernel arg found: '%s'", key)
+			}
+			seenKeys[key] = true
+		}
```
atanasdinov, line 95: It feels like this can be simplified quite a lot:
```
key := arg

parts := strings.SplitN(arg, "=", 2)
if len(parts) == 2 {
    var value string
    key, value = parts[0], parts[1]
    if key == "" {
	return fmt.Errorf("kernel arg value '%s' has no key", value)
    }
    if value == "" {
	return fmt.Errorf("kernel arg '%s' has no value", key)
    }
}

if _, exists := seenKeys[key]; exists {
    return fmt.Errorf("duplicate kernel arg found: '%s'", key)
}
seenKeys[key] = true
```

pkg/image/validation.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	return nil
+}
+
+func checkForDuplicates(items []string) error {
+	seen := make(map[string]bool)
+	for _, item := range items {
+		if seen[item] {
+			return fmt.Errorf("'%s'", item)
```
atanasdinov, line 125: This doesn't have to be an error - you can just `return item` as a string.

atanasdinov (commented):
I'd strongly suggest to look into implementing table tests for these cases. It's a perfect fit for readability and the assertion issues that I listed as comments wouldn't be there in the first place. There are numerous examples in the repository if you need such e.g. [TestParse](https://github.com/suse-edge/edge-image-builder/blob/main/pkg/template/template_test.go).

atanasdinov (approved):

PR #81, opened 2024-01-05
Add arch to definition

atanasdinov:
- Adds a new `arch` field to the image definition file
- Includes validation for the new field
- Drops image name requirements for network component
- Sets up a graceful recovery from runtime errors / panics

### DIFF
--- a/docs/building-images.md
+++ b/docs/building-images.md
@@ -2,7 +2,7 @@
 
 Two things are necessary to build an image using EIB:
 1. A definition file that describes the image to build
-1. A directory that contains the base SLE Micro image to modify, along with any other custom files that
+2. A directory that contains the base SLE Micro image to modify, along with any other custom files that
    will be included in the built image
 
 ## Image Definition File
@@ -14,6 +14,7 @@ directory may be used to build multiple images by creating multiple definition f
 The following can be used as the minimum configuration required to create an image:
 ```yaml
 apiVersion: 1.0
+arch: x86_64
 image:
   imageType: iso
   baseImage: SLE-Micro.x86_64-5.5.0-Default-SelfInstall-GM.install.iso
@@ -21,6 +22,7 @@ image:
 ```
 
 * `apiVersion` - Indicates the version of the definition file schema for EIB to expect
+* `arch` - Must be either `x86_64` or `aarch64`.
 * `imageType` - Must be either `iso` or `raw`.
 * `baseImage` - Indicates the name of the image file used as the base for the built image. This file must be located
   under the `images` directory of the image configuration directory (see below for more information). This image will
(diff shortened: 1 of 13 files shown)


pkg/image/definition.go
```diff
@@ -11,12 +11,32 @@ const (
 	TypeRAW = "raw"
 )
 
+const (
+	ArchTypeIntel Arch = "x86_64"
```
rdoxenham, line 15: Suggest this is `ArchTypeX86` rather than `ArchTypeIntel`, I mean, it was AMD that came up with x86_64 anyway ;-)

rdoxenham (changes requested):
One minor thing, and if you'd prefer we can leave it as is. Otherwise, this is great. Thank you!

rdoxenham (approved):

pkg/image/testdata/full-valid-example.yaml
```diff
@@ -1,4 +1,5 @@
 apiVersion: 1.0
+arch: x86_64
```
jdob, line 2: My expectation was that this would fall under the `image` section since it's dictating the type of image that's being created/modified. In other words, it feels like it's at the same level the user would be thinking "Raw or ISO?". Is there a reason you put it outside of that?

pkg/image/definition.go
```diff
@@ -11,12 +11,32 @@ const (
...
+
+func (a Arch) Short() string {
+	switch a {
+	case ArchTypeX86:
+		return "amd64"
+	case ArchTypeARM:
+		return "arm64"
+	default:
+		message := fmt.Sprintf("unknown arch: %s", a)
+		panic(message)
```
jdob, line 36: This may be a Go convention (meaning the `Short()` method in general), but it feels like having a potential `panic()` in the middle of our application is a bad thing. I'm sure I'm missing something, but can you explain why both why this would have a panic included, and where it's intended to be used (I don't see any current usages)?

jdob (changes requested):
I set it as "Request Changes", but that's mostly to make sure we at least have the conversation about where in the definition arch lives. Once we talk it out that may end up being the right spot, in which case I'll approve.

docs/building-images.md
```diff
@@ -2,7 +2,7 @@
 
 Two things are necessary to build an image using EIB:
 1. A definition file that describes the image to build
-1. A directory that contains the base SLE Micro image to modify, along with any other custom files that
+2. A directory that contains the base SLE Micro image to modify, along with any other custom files that
```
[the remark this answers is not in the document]
atanasdinov: Weird. The IDE highlighted it for me. Will revert.

pkg/image/testdata/full-valid-example.yaml
atanasdinov: Fair point. I actually misread the EIB plan document. Will move it under `image`.

pkg/image/definition.go
atanasdinov: It makes sense for the user to input the `x86_64` and `aarch64` values as I believe these are the most "common" ones to differentiate the instruction sets. However, there are also a lot of places where `amd64` and `arm64` are used in their places instead so I added this `Short()` function. 

Frankly speaking, I don't think it's the best way to name it but I also didn't find anything else suitable. I'm open to ideas though.

And lastly, this is going to be used for the Kubernetes RKE2 and K3S artefacts. Those include the "shortened" versions e.g. `rke2.linux-amd64`, `rke2-images-core.linux-arm64.tar.zst`, etc.

Having the `panic()` here is fine since:
- our validation will ensure that unsupported `arch` is not possible
- `Short()` can easily be used without having to handle an error case
- `recover()` is set in place in order to gracefully recover from the unexpected case of this or anything else running into a runtime error

jdob: Ahhh, ok, that makes sense given it's usage. I'd suggest a better name but I can't think of one :)

I agree that the panic is preventable, so if you're comfortable with it, I am too.

atanasdinov: Yes, I'd say this is indeed one of the few cases where a `panic()` is more convenient to have. We basically want to limit the `Arch` type to an enum of the two supported values `x86_64` and `aarch64` and their _shorter_ representations.

jdob (approved):

PR #82, opened 2024-01-05
Add file downloader

atanasdinov:
- Introduces an http package and the option to download files
- It will later be used in Kubernetes components to download the necessary artefacts for airgapped installs


dbw7 (approved):
Looks good!

atanasdinov:
I'd want to explore a little better [Github's API](https://github.com/google/go-github) before merging this.

pkg/http/download.go
```diff
@@ -0,0 +1,43 @@
...
+// DownloadFile downloads a file from the specified URL and stores it to the given path.
+func DownloadFile(ctx context.Context, url, path string) error {
+	zap.S().Infof("Downloading file from '%s' to '%s'...", url, path)
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, http.NoBody)
+	if err != nil {
+		return fmt.Errorf("creating request: %w", err)
+	}
+
+	resp, err := http.DefaultClient.Do(req)
```
jdob, line 22: No surprise, I've never used this library in Go. Do we want to explicitly add some sort of timeout to this call (assuming it's even supported)? I worry about the possibility of EIB "hanging" because it's trying to download and for whatever reason, the connection is taking forever.

pkg/http/download_test.go
```diff
@@ -0,0 +1,66 @@
...
+		{
+			name:        "Unexpected status",
+			ctx:         context.Background(),
+			url:         "https://github.com/suse-edge/eib",
+			expectedErr: "unexpected status code: 404",
+		},
+		{
+			name:        "Error creating file",
+			ctx:         context.Background(),
+			url:         "https://github.com/suse-edge/edge-image-builder",
```
jdob, line 39: This is tricky since I'm generally not a fan of unit tests making a network call. Is there any sort of way in Go to differentiate between "unit" and "integration" tests, such that we can only have this test run in CI but not by default locally? The answer may be that I need to do more than `go test ./...` locally. Just curious if you had any experience with this sort of test differentiation.

pkg/http/download.go
atanasdinov: This is what the `ctx context.Context` field is included. The caller will determine whether or how it'd want to manage the download.

You can refer to the [package documentation](https://pkg.go.dev/context) or this [blog post](https://go.dev/blog/context) for further explanation and examples.

pkg/http/download_test.go
atanasdinov: I considered these lightweight enough not to label as integration tests. However, it is indeed possible and actually very easy to implement the separation via "build tags". Here is a short [blog post](https://mickey.dev/posts/go-build-tags-testing/) showcasing how it's done.

I'll come back to this once I figure out how much of the functionality here will be used for the Kubernetes combustion components.

pkg/http/download.go
atanasdinov: Short example of how this will work for us:

```
ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
defer cancel()
			
err := DownloadFile(ctx, <url>, <path>)
// Error handling
```

If the `http.Do(req)` call (in `DownloadFile`) takes longer than 10 seconds, it will automatically fail with a `context deadline exceeded` error.

atanasdinov:
> I'd want to explore a little better [Github's API](https://github.com/google/go-github) before merging this.

I decided against using the Github API for now. It is fairly straightforward to list the release assets using [GetReleaseByTag](https://github.com/google/go-github/blob/v57.0.0/github/repos_releases.go#L138) and downloading those using [DownloadReleaseAsset](https://github.com/google/go-github/blob/v57.0.0/github/repos_releases.go#L343).

However, by relying on those, we still need to filter only the necessary assets for a particular build and using the API can very easily run into [rate limiting](https://github.com/google/go-github/tree/v57.0.0?tab=readme-ov-file#rate-limiting) errors. In order to resolve these, then we'd also have to introduce GitHub authentication and given the time constraints we have, this is not something which will provide us with a huge benefit.

We'll instead rely on using the download functionality introduced in this PR against the so called _web download urls_ (e.g. `https://github.com/rancher/rke2/releases/download/<version>/<asset>`) to download the relevant artefacts.

pkg/http/download.go
jdob: Ahhh, ok, so there is a hook for it. Thanks for the clarification.

pkg/http/download_test.go
jdob: Works for me 👍🏼

jdob (approved):

PR #84, opened 2024-01-08
Air-grapped Workloads Definition

dbw7:
This is the first in a set of PRs for the implementation of air-gapped workloads. This first adds the Hauler definition, updates the example YAML file, and adds validation and tests for the Hauler definition.


pkg/image/definition.go
```diff
@@ -16,6 +16,7 @@ type Definition struct {
 	APIVersion      string          `yaml:"apiVersion"`
 	Image           Image           `yaml:"image"`
 	OperatingSystem OperatingSystem `yaml:"operatingSystem"`
+	Hauler          Hauler          `yaml:"hauler"`
```
atanasdinov, line 19: Hauler is the implementation - not the API contract we want to expose to the user. This should be `workloads` or something similar where it's clear what goes into it.

pkg/image/definition.go
```diff
@@ -48,6 +49,28 @@ type Suma struct {
...
 
+type Hauler struct {
+	ContainerImages []ContainerImage `yaml:"images"`
+	HelmCharts      []HelmChart      `yaml:"charts"`
+	Files           []File           `yaml:"files"`
+}
+
+type ContainerImage struct {
+	Name string `yaml:"name"`
+	Key  string `yaml:"key"`
```
atanasdinov, line 60: What type of key do we specify here?

pkg/image/definition.go
```diff
@@ -48,6 +49,28 @@ type Suma struct {
...
+
+type HelmChart struct {
+	Name    string `yaml:"name"`
+	RepoURL string `yaml:"repoURL"`
+	Version string `yaml:"version"`
+}
+
+type File struct {
+	Name string `yaml:"name"`
+	Path string `yaml:"path"`
```
atanasdinov, line 71: Can this be a local path or is it always an HTTP URL?

pkg/image/validation.go
```diff
@@ -159,5 +165,95 @@ func validateSuma(os *OperatingSystem) error {
...
+
+func validateHelmCharts(charts []HelmChart) error {
+	seenCharts := make(map[string]bool)
+
+	for _, chart := range charts {
+		if chart.Name == "" {
+			return fmt.Errorf("no chart name defined")
+		}
+		if chart.RepoURL == "" {
+			return fmt.Errorf("no chart repository URL defined")
```
atanasdinov, line 223: Can we include the name of the chart in this wrapper?

pkg/image/validation.go
```diff
@@ -159,5 +165,95 @@ func validateSuma(os *OperatingSystem) error {
...
+		}
+		if chart.Version == "" {
+			return fmt.Errorf("no chart version defined")
```
atanasdinov, line 226: Can we include the name of the chart in this wrapper?

pkg/image/validation.go
```diff
@@ -159,5 +165,95 @@ func validateSuma(os *OperatingSystem) error {
...
+
+func validateFiles(files []File) error {
+	seenFiles := make(map[string]bool)
+
+	for _, file := range files {
+		if file.Name == "" {
+			return fmt.Errorf("no file name defined")
+		}
+		if file.Path == "" {
+			return fmt.Errorf("no file path defined")
```
atanasdinov, line 249: Can we include the name of the file in this wrapper?

pkg/image/validation_test.go
```diff
@@ -533,3 +533,266 @@ func TestValidateOperatingSystemSumaMissingActivationKey(t *testing.T) {
...
+				{
+					Name: "hello-world:latest",
+					Key:  "",
+				},
+				{
+					Name: "rgcrprod.azurecr.us/longhornio/longhorn-ui:v1.5.1",
+					Key:  "carbide-key.pub",
+				},
+				{
+					Name: "fake-image:latest",
```
atanasdinov, line 590: This is not necessary since it is identical to the first case.

pkg/image/validation_test.go
```diff
@@ -533,3 +533,266 @@ func TestValidateOperatingSystemSumaMissingActivationKey(t *testing.T) {
...
+				},
+				{
+					Name: "fake-image2-with-dupe-key:latest",
```
atanasdinov, line 593: This is not necessary since it is identical to the second case.

pkg/image/validation_test.go
```diff
@@ -533,3 +533,266 @@ func TestValidateOperatingSystemSumaMissingActivationKey(t *testing.T) {
...
+		},
+		{
+			name: "No Image Name Defined",
+			images: []ContainerImage{
+				{
+					Name: "",
+					Key:  "",
+				},
+				{
+					Key: "",
```
atanasdinov, line 606: Identical to the previous case.

pkg/image/validation_test.go
```diff
@@ -533,3 +533,266 @@ func TestValidateOperatingSystemSumaMissingActivationKey(t *testing.T) {
...
+				{
+					Name: "hello-world:latest",
+					Key:  "carbide-key.pub",
+				},
+			},
+			expectedErr: "duplicate container image found: 'hello-world:latest'",
+		},
+	}
+
+	// Test
```
atanasdinov, line 627: I don't understand the purpose of these comments. What does it say which we don't already see?

atanasdinov (changes requested):
We must use an alternative name for the section in the definition file. The rest looks fine to me. Happy to approve after this is resolved.

pkg/image/definition.go
rdoxenham: Exactly this ^^

pkg/image/definition.go
dbw7: resolving as a result of the removal of hauler file handling

pkg/image/validation.go
dbw7: resolving as a result of the removal of hauler file handling

pkg/image/definition.go
dbw7: Putting what we discussed here for visibility:
https://rancherfederal.github.io/hauler-docs/docs/introduction/quickstart#using-the-command-line
This key is for supply chain artifacts as shown in the link above.
Will rename `Key` to `SupplyChainKey`

atanasdinov (approved):
I'm good with this landing after the `hauler` references are completely dropped. Good work!

rdoxenham (approved):
LGTM, thanks!

pkg/image/validation.go
```diff
@@ -159,5 +165,69 @@ func validateSuma(os *OperatingSystem) error {
...
+	err = validateHelmCharts(definition.EmbeddedArtifactRegistry.HelmCharts)
+	if err != nil {
+		return fmt.Errorf("error validating helm charts: %w", err)
+	}
+
+	return nil
+}
+
+func checkIfEmbeddedArtifactRegistryDefined(definition *Definition) bool {
+	return len(definition.EmbeddedArtifactRegistry.HelmCharts) == 0 && len(definition.EmbeddedArtifactRegistry.ContainerImages) == 0
```
jdob, line 189: I may totally be misreading this, but this equates to "If both of these are zero, return that yes, embedded artifact registry is defined". It feels like the logic is backwards from the name. Am I just undercaffeinated?

jdob (changes requested):
I'll flip to approve if it turns out I'm completely wrong on the check method

pkg/image/validation.go
dbw7: nah you're right, your logic makes more sense

jdob (approved):

PR #85, opened 2024-01-09
Introduce copy file content by byte chunks logic

ipetrov117:
Introduce the copy file content by byte chunks logic introduced in #64 in a separate PR in order to reduce complexity of #64

### DIFF
--- a/pkg/fileio/file_io.go
+++ b/pkg/fileio/file_io.go
@@ -1,6 +1,7 @@
 package fileio
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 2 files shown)


pkg/fileio/file_io.go
```diff
@@ -22,21 +23,54 @@ func CopyFile(src string, dest string, perms os.FileMode) error {
...
+}
+
+func createFileWithPerms(dest string, perms os.FileMode) (*os.File, error) {
+	file, err := os.Create(dest)
+	if err != nil {
+		return nil, fmt.Errorf("creating file: %w", err)
+	}
+
+	if err = file.Chmod(perms); err != nil {
+		return nil, fmt.Errorf("setting up permissions: %w", err)
```
atanasdinov, line 72: We need to close the file before returning, otherwise it will remain open.

pkg/fileio/file_io_test.go
```diff
@@ -74,3 +76,76 @@ func TestCopyFile(t *testing.T) {
...
+
+	tmpDir, err := os.MkdirTemp("", destDirPrefix)
+	require.NoError(t, err)
+	defer os.RemoveAll(tmpDir)
+
+	tmpFile, err := os.CreateTemp(tmpDir, srcFilePrefix)
+	require.NoError(t, err)
+	defer func() {
+		require.NoError(t, tmpFile.Close())
+		require.NoError(t, os.Remove(tmpFile.Name()))
```
atanasdinov, line 95: This is already covered by the `defer os.RemoveAll(tmpDir)` call above.

atanasdinov, line 95: Nit: We don't need a file in the first place since `CopyFileN()` accepts an `io.Reader`. This can be simplified by passing `bytes.NewBufferString(srcFileContent)` to each test case and just asserting the string literal.

atanasdinov (commented):
Thanks for extracting this out! My only concern is the file not being closed if permission changing fails. Happy to approve after this is resolved!

pkg/fileio/file_io_test.go
```diff
@@ -74,3 +76,76 @@ func TestCopyFile(t *testing.T) {
...
+			name:        "File is successfully copied",
+			source:      tmpFile,
+			destination: fmt.Sprintf("%s/copy", tmpDir),
+			perms:       NonExecutablePerms,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			err := CopyFileN(test.source, test.destination, test.perms, 4096)
```
atanasdinov, line 130: Can we also make the chunk size 1 byte only? Using 4096 means that the contents are read in one go.

atanasdinov (approved):

PR #86, opened 2024-01-09
Introduce 'podman' library in EIB project

ipetrov117:
Introduces the 'podman' library present in #64 in a separate PR in order to reduce #64 complexity


pkg/podman/listener.go
```diff
@@ -0,0 +1,46 @@
...
+		return fmt.Errorf("configuring the podman system serice command: %w", err)
+	}
+	defer logfile.Close()
+
+	err = cmd.Start()
+	if err != nil {
+		return fmt.Errorf("error running podman system service: %w", err)
+	}
+
+	return err
```
atanasdinov, line 31: This is always `nil`.

pkg/podman/listener.go
```diff
@@ -0,0 +1,46 @@
...
+const (
+	podmanArgsBase        = "--log-level=debug system service -t 0"
+	podmanExec            = "/usr/bin/podman"
+	podmanListenerLogFile = "podman-system-service.log"
+)
+
+// creates a listening service that answers API calls for Podman (https://docs.podman.io/en/v4.8.2/markdown/podman-system-service.1.html)
+// only way to start the service from within a container - https://github.com/containers/podman/tree/v4.8.2/pkg/bindings#starting-the-service-manually
+func setupAPIListener(out string) error {
+	cmd, logfile, err := preparePodmanCommand(out)
```
atanasdinov, line 20: The problem with this is that it's much harder to test the arguments of the command. It's more flexible to create the log file and only then run `preparePodmanCommand` with an `io.Writer` argument.

pkg/podman/listener.go
```diff
@@ -0,0 +1,46 @@
...
+		return fmt.Errorf("error running podman system service: %w", err)
+	}
+
+	return err
+}
+
+func preparePodmanCommand(out string) (*exec.Cmd, *os.File, error) {
+	logFile, err := os.Create(filepath.Join(out, podmanListenerLogFile))
+	if err != nil {
+		return nil, nil, fmt.Errorf("generating podman lister log file: %w", err)
```
atanasdinov, line 37:
```suggestion
		return nil, nil, fmt.Errorf("creating podman listener log file: %w", err)
```

go.mod
```diff
@@ -3,14 +3,137 @@ module github.com/suse-edge/edge-image-builder
 go 1.21.1
 
 require (
+	// version should match buildah version in the
+	// podman mod file https://github.com/containers/podman/blob/v4.8.2/go.mod#L14
+	github.com/containers/buildah v1.33.2
+	github.com/containers/podman/v4 v4.8.2
```
atanasdinov, line 9: There's a newer patch version out now (`4.8.3`). Should we upgrade?

pkg/podman/podman.go
```diff
@@ -0,0 +1,188 @@
...
+const (
+	podmanSock         = "unix:///run/podman/podman.sock"
+	dockerfile         = "Dockerfile"
+	podmanDirName      = "podman"
+	podmanBuildLogFile = "podman-image-build.log"
+)
+
+type Podman struct {
+	context context.Context
+	socket  string
```
atanasdinov, line 32: This field is unused. Can we remove it?

pkg/podman/podman.go
```diff
@@ -0,0 +1,188 @@
...
+	out     string
+}
+
+// New setups a podman listening service and returns a connected podman client.
+//
+// Parameters:
+//   - out - location for podman to output any logs created as a result of podman commands
+func New(out string) (*Podman, error) {
+	podmanDirPath := filepath.Join(out, podmanDirName)
```
atanasdinov, line 41: I'm not entirely sure if we want this. After the recent build directory restructuring all log files are now stored at the root `BuildDir`. This makes it very easy to follow and I'm afraid if we start introducing subdirectories only for the sake of log files then we will make troubleshooting more confusing. After all, everything would also need to be covered by tests and docs. What are your thoughts?

pkg/podman/podman.go
```diff
@@ -0,0 +1,188 @@
...
+	}, nil
+}
+
+// Import imports a tarball and saves it as a filesystem image
+//
+// Parameters:
+//   - tarball - path to the tarball to be imported
+//   - ref 	  - name for the image that will be created from the tarball
+func (p *Podman) Import(tarball, ref string) error {
+	zap.L().Sugar().Infof("Importing image '%s' from tarball...", ref)
```
atanasdinov, line 68: Nit: If you want to use the suggared logger, you can just call `S()` instead. Also applies to the other occurrences below.
```suggestion
	zap.S().Infof("Importing image '%s' from tarball...", ref)
```

pkg/podman/podman.go
```diff
@@ -0,0 +1,188 @@
...
+	if err != nil {
+		return fmt.Errorf("building image from context %s: %w", imageContext, err)
+	}
+
+	return nil
+}
+
+// Create creates a container from the given image. Returns the id of the container.
+func (p *Podman) Create(img string) (string, error) {
+	zap.L().Sugar().Infof("Running container from %s image...", img)
```
atanasdinov, line 111:
```suggestion
	zap.L().Sugar().Infof("Creating container from %s image...", img)
```

pkg/podman/podman.go
```diff
@@ -0,0 +1,188 @@
...
+			return fmt.Errorf("illegal file path: %w", err)
+		}
+
+		switch header.Typeflag {
+		case tar.TypeDir:
+			if err = os.MkdirAll(path, os.FileMode(header.Mode)); err != nil {
+				return fmt.Errorf("creating directory %s: %w", path, err)
+			}
+		case tar.TypeReg:
+			if err = fileio.CopyFileN(tarReader, path, os.FileMode(header.Mode), 4096); err != nil {
```
atanasdinov, line 171: Can we extract a constant for the chunk size?

pkg/podman/podman.go
```diff
@@ -0,0 +1,188 @@
...
+	reader, writer := io.Pipe()
+	defer reader.Close()
+
+	copyFunc, err := containers.CopyToArchive(p.context, id, src, writer)
+	if err != nil {
+		return fmt.Errorf("creating copy function for archive from %s: %w", src, err)
+	}
+
+	go func() {
+		writer.CloseWithError(copyFunc())
```
atanasdinov, line 139: I'd suggest to extract the `copyFunc` call out of this, log the error and only then call `writer.CloseWithError()`.

I'm also unsure whether `writer.Close()` is not the better approach here. The reason being is that there are several `CopyToArchive` error results which are not using the writer at all.

atanasdinov (changes requested):
The integration looks good overall. It's only when we start using the package that it'll be more clear whether splitting up the various logic flows (e.g. `Import()`, `Build()`...) is better than a wrapper around those. The latter would technically make it easier to mock but let's see how this turns out.

There's nothing critical but I'd love to hear your thoughts on some of the questions below.

pkg/podman/listener.go
ipetrov117: Agreed, I was using the [xorriso](https://github.com/suse-edge/edge-image-builder/blob/main/pkg/build/iso.go#L48) prepare command function for inspiration, but I definitely see the value in your suggestion.

pkg/podman/podman.go
ipetrov117: Agreed, having all log files in a single place would make it easier for the user to debug possible problems. Also the user would not have to worry if he missed any log files.

go.mod
```diff
@@ -3,14 +3,137 @@ module github.com/suse-edge/edge-image-builder
 go 1.21.1
 
 require (
+	// version should match buildah version in the
+	// podman mod file https://github.com/containers/podman/blob/v4.8.2/go.mod#L14
```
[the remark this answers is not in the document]
atanasdinov: We'll be keeping this as discussed offline.

pkg/podman/podman.go
```diff
@@ -0,0 +1,188 @@
...
+
+const (
+	podmanSock         = "unix:///run/podman/podman.sock"
+	dockerfile         = "Dockerfile"
+	podmanDirName      = "podman"
+	podmanBuildLogFile = "podman-image-build.log"
+)
+
+type Podman struct {
+	context context.Context
```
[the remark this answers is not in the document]
ipetrov117: Yea a GitHub issue would need to be raised for this. Essentially we would probably have to do an integration test here as opposed to doing a unit test. Mainly this is true because in order to test the functionality, we need a [podman-system-service](https://docs.podman.io/en/latest/markdown/podman-system-service.1.html) started in order to be able to communicate with podman and execute commands, which will be very hard to mock as a unit test. IMO an integration test would suit this better.

pkg/podman/podman.go
```diff
@@ -0,0 +1,190 @@
...
+
+	copyFunc, err := containers.CopyToArchive(p.context, id, src, writer)
+	if err != nil {
+		return fmt.Errorf("creating copy function for archive from %s: %w", src, err)
+	}
+
+	go func() {
+		err := copyFunc()
+		if err != nil {
+			zap.S().Errorf("copying %w to %w: %w", src, dest, err)
```
atanasdinov, line 134: This is where we log the outcome so the %w formatter is not appropriate: 
```suggestion
			zap.S().Errorf("Copying %s to %s failed: %s", src, dest, err)
```

atanasdinov (approved):

jdob (approved):

PR #89, opened 2024-01-09
Add kubernetes definition

atanasdinov:
- Introduces kubernetes definition and validation
- Possible configurations are kept to a minimum
- Follow ups might introduce more options e.g. SELinux

### DIFF
--- a/pkg/image/testdata/full-valid-example.yaml
+++ b/pkg/image/testdata/full-valid-example.yaml
@@ -27,3 +27,9 @@ operatingSystem:
     host: suma.edge.suse.com
     activationKey: slemicro55
     getSSL: false
+kubernetes:
+  version: v1.29.0+rke2r1
+  nodeType: server
+  cni: cilium
+  multus: true
+  vSphere: false
--- a/pkg/image/validation.go
+++ b/pkg/image/validation.go
@@ -2,17 +2,23 @@ package image
 
 import (
 	"fmt"
+	"slices"
 	"strings"
+
+	"github.com/suse-edge/edge-image-builder/pkg/log"
 )
 
 func ValidateDefinition(definition *Definition) error {
-	err := validateImage(definition)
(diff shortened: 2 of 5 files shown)


pkg/image/validation.go
```diff
@@ -64,6 +70,31 @@ func validateOperatingSystem(definition *Definition) error {
...
+func validateKubernetes(definition *Definition) error {
+	if definition.Kubernetes.Version == "" {
+		// Not configured
+		return nil
+	}
+
+	supportedCNIs := []string{CNITypeNone, CNITypeCanal, CNITypeCalico, CNITypeCilium}
+
+	if definition.Kubernetes.CNI == "" {
+		log.Audit("Kubernetes CNI was not specified. Please set \"cni: none\" if you intend to use your own")
```
atanasdinov, line 82: This is an edge case which we can technically log to the file, however, the message will be written before writing the following error which might be counterintuitive. I settled for auditing it but I'm also fine with revisiting this if it doesn't make sense.

jdob: Not 100% sure I follow, but I'm fine with this approach until we see how it plays out during usage.

jdob (approved):
Approved even with the question about node type. I have a feeling you have a bigger picture in mind and I trust your longer term direction here.

pkg/image/testdata/full-valid-example.yaml
```diff
@@ -27,3 +27,9 @@ operatingSystem:
     host: suma.edge.suse.com
     activationKey: slemicro55
     getSSL: false
+kubernetes:
+  version: v1.29.0+rke2r1
+  nodeType: server
```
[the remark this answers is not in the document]
atanasdinov: Yes, this will most likely change. I wouldn't want to play around with unknown variables at this point, though. It's an easy change once we need it.

PR #90, opened 2024-01-09
Added logic to unsquash/resquash the raw image inside of the ISO

jdob:
This has no net new functionality, but lays the groundwork to refactor the raw image modifications to be done for ISOs as well.

I do want to (eventually) add a test ISO (obviously a custom built one and not a GB-sized one) to actually run the scripts against, but for now the tests prove that the correct values are passed into the templates.

### DIFF
--- a/pkg/build/iso.go
+++ b/pkg/build/iso.go
@@ -56,7 +56,7 @@ func (b *Builder) extractIso() error {
 	}
 	defer func() {
 		if err = extractLog.Close(); err != nil {
-			zap.L().Warn("failed to close ISO extraction log file properly", zap.Error(err))
+			zap.S().Warn("failed to close ISO extraction log file properly", zap.Error(err))
 		}
 	}()
 
@@ -78,7 +78,7 @@ func (b *Builder) rebuildIso() error {
 	}
 	defer func() {
 		if err = rebuildLog.Close(); err != nil {
-			zap.L().Warn("failed to close ISO rebuild log file properly", zap.Error(err))
+			zap.S().Warn("failed to close ISO rebuild log file properly", zap.Error(err))
 		}
 	}()
 
--- a/pkg/build/iso.go
+++ b/pkg/build/iso.go
@@ -1,39 +1,94 @@
 package build
 
 import (
+	_ "embed"
 	"fmt"
 	"os"
 	"os/exec"
 	"path/filepath"
-	"strings"
 
+	"github.com/suse-edge/edge-image-builder/pkg/fileio"
+	"github.com/suse-edge/edge-image-builder/pkg/template"
 	"go.uber.org/zap"
 )
 
 const (
-	xorrisoArgsBase = "-indev %s -outdev %s -map %s /combustion -boot_image any replay -changes_pending yes"
-	xorrisoExec     = "/usr/bin/xorriso"
-	xorrisoLogFile  = "iso-build.log"
+	isoExtractDir        = "iso-extract"
+	rawExtractDir        = "raw-extract"
+	extractIsoScriptName = "iso-extract.sh"
+	extractIsoLogFile    = "iso-extract.log"
+	rebuildIsoScriptName = "iso-build.sh"
+	rebuildIsoLogFile    = "iso-build.log"
 )
 
+//go:embed templates/extract-iso.sh.tpl
+var extractIsoTemplate string
+
+//go:embed templates/rebuild-iso.sh.tpl
+var rebuildIsoTemplate string
+
 func (b *Builder) buildIsoImage() error {
-	err := b.deleteExistingOutputIso()
-	if err != nil {
+	if err := b.deleteExistingOutputIso(); err != nil {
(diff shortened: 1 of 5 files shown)


pkg/build/iso_test.go
```diff
@@ -65,42 +93,68 @@ func TestDeleteExistingImage(t *testing.T) {
...
+	assert.Contains(t, found, fmt.Sprintf("ISO_EXTRACT_DIR=%s", expectedIsoExtractDir))
+
+	expectedRawExtractDir := filepath.Join(ctx.BuildDir, rawExtractDir)
+	assert.Contains(t, found, fmt.Sprintf("RAW_EXTRACT_DIR=%s", expectedRawExtractDir))
+
+	expectedIsoPath := builder.generateBaseImageFilename()
+	assert.Contains(t, found, fmt.Sprintf("ISO_SOURCE=%s", expectedIsoPath))
+}
+
+func TestWriteIsoScript_Rebuild(t *testing.T) {
```
jdob, line 126: Before you yell at me and my distaste for table tests, I did consider them here. Ultimately though, we're going to further modify the rebuild script for other functionality and I suspect the testing for those changes will further distinguish the two tests. That said, if those changes aren't as invasive as I fear, I'll refactor to table tests.

pkg/build/templates/extract-iso.sh.tpl
```diff
@@ -0,0 +1,22 @@
...
+
+#  Template Fields
+#  IsoExtractDir - Full path to the directory (under the build directory) where the ISO should be extracted
+#  RawExtractDir - Full path to the directory (under the build directory) where the RAW image should be
+#                  unsquashed
+#  IsoSource - Full path to the ISO to extract
+
+ISO_EXTRACT_DIR={{.IsoExtractDir}}
+RAW_EXTRACT_DIR={{.RawExtractDir}}
+ISO_SOURCE={{.IsoSource}}
```
atanasdinov, line 12: Is there a benefit in using bash variables instead of the templated ones? Seems like a repetition to me which is making the script unnecessarily longer.

pkg/build/templates/rebuild-iso.sh.tpl
```diff
@@ -0,0 +1,39 @@
...
+#  RawExtractDir - Full path to the directory where the RAW image was extracted
+#  IsoSource - Full path to the original ISO that was extracted
+#  OutputImageFilename - Full path and name of the ISO to create
+#  CombustionDir - Full path to the combustion directory to include in the new ISO
+
+ISO_EXTRACT_DIR={{.IsoExtractDir}}
+RAW_EXTRACT_DIR={{.RawExtractDir}}
+ISO_SOURCE={{.IsoSource}}
+OUTPUT_IMAGE={{.OutputImageFilename}}
+COMBUSTION_DIR={{.CombustionDir}}
```
atanasdinov, line 15: Same as above.

pkg/build/iso.go
```diff
@@ -1,39 +1,92 @@
...
+	}
+
+	cmd, rebuildLog, err := b.createIsoCommand(rebuildIsoLogFile, rebuildIsoScriptName)
 	if err != nil {
-		return fmt.Errorf("error running xorriso: %w", err)
+		return fmt.Errorf("preparing to build the new ISO: %w", err)
+	}
+	defer func() {
+		if err = rebuildLog.Close(); err != nil {
+			zap.L().Warn("failed to close ISO rebuild log file properly", zap.Error(err))
```
atanasdinov, line 81: The current format of the log file is text and not JSON so using structured logging looks weird in the output.
```suggestion
			zap.S().Warnf("Failed to close ISO rebuild log file properly: %s", err)
```

pkg/build/iso.go
```diff
@@ -1,39 +1,92 @@
...
+	}
+
+	cmd, extractLog, err := b.createIsoCommand(extractIsoLogFile, extractIsoScriptName)
 	if err != nil {
-		return fmt.Errorf("configuring the ISO build command: %w", err)
+		return fmt.Errorf("preparing to extract the contents of the ISO: %w", err)
+	}
+	defer func() {
+		if err = extractLog.Close(); err != nil {
+			zap.L().Warn("failed to close ISO extraction log file properly", zap.Error(err))
```
atanasdinov, line 59: Same as above.

atanasdinov (approved):

pkg/build/templates/extract-iso.sh.tpl
jdob: I went back and forth on this. Doing it this way helps separate the tests verifying the template is correctly filled out from tests that the script itself works. The division here is that testing this function only cares that the correct values are passed into the template, not that the script itself is correct.

pkg/build/iso.go
jdob: Man, I really haven't looked at the zap APIs at all since it was set up. I didn't realize this distinction. I'll fix both mentioned instances.

PR #91, opened 2024-01-09
Updated all uses of zap to use the sugared logger for consistency

jdob:


pkg/combustion/network.go
```diff
@@ -88,7 +88,7 @@ func generateNetworkConfig(ctx *image.Context) error {
 
 	defer func() {
 		if err = logFile.Close(); err != nil {
-			zap.L().Warn("Failed to close network log file properly", zap.Error(err))
+			zap.S().Warn("Failed to close network log file properly: %s", err)
```
atanasdinov, line 91:
```suggestion
			zap.S().Warnf("Failed to close network log file properly: %s", err)
```

atanasdinov (approved):

PR #92, opened 2024-01-10
Introduce 'package' definition and validation

ipetrov117:
Introduces the 'package' definition and validation that #64 requires for its RPM resolution logic. This is introduced in a separate PR in order to reduce the complexity of #64.

### DIFF
--- a/pkg/image/definition.go
+++ b/pkg/image/definition.go
@@ -64,9 +64,9 @@ type OperatingSystem struct {
 }
 
 type Packages struct {
-	PKGList  []string `yaml:"packageList"`
-	AddRepos []string `yaml:"additionalRepos"`
-	RegCode  string   `yaml:"registrationCode"`
+	PKGList         []string `yaml:"packageList"`
+	AdditionalRepos []string `yaml:"additionalRepos"`
+	RegCode         string   `yaml:"registrationCode"`
 }
 
 type OperatingSystemUser struct {
--- a/pkg/image/definition.go
+++ b/pkg/image/definition.go
@@ -60,6 +60,13 @@ type OperatingSystem struct {
 	Users      []OperatingSystemUser `yaml:"users"`
 	Systemd    Systemd               `yaml:"systemd"`
 	Suma       Suma                  `yaml:"suma"`
+	Packages   Packages              `yaml:"packages"`
+}
+
+type Packages struct {
+	PKGList  []string `yaml:"packageList"`
+	AddRepos []string `yaml:"additionalRepos"`
+	RegCode  string   `yaml:"registrationCode"`
 }
 
 type OperatingSystemUser struct {
--- a/pkg/image/definition_test.go
+++ b/pkg/image/definition_test.go
@@ -64,6 +64,26 @@ func TestParse(t *testing.T) {
 	assert.Equal(t, "slemicro55", suma.ActivationKey)
 	assert.Equal(t, false, suma.GetSSL)
 
+	// Operating System -> Packages
+	pkgConfig := definition.OperatingSystem.Packages
+	require.Len(t, pkgConfig.PKGList, 6)
+	require.Len(t, pkgConfig.AddRepos, 2)
+	expectedPKGList := []string{
+		"wget2",
+		"dpdk22",
+		"dpdk22-tools",
+		"libdpdk-23",
+		"libatomic1",
+		"libbpf0",
+	}
+	assert.Equal(t, expectedPKGList, pkgConfig.PKGList)
+	expectedAddRepos := []string{
+		"https://download.nvidia.com/suse/sle15sp5/",
+		"https://developer.download.nvidia.com/compute/cuda/repos/sles15/x86_64/",
+	}
+	assert.Equal(t, expectedAddRepos, pkgConfig.AddRepos)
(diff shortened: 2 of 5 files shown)


ipetrov117:
@jdob as discussed [here](https://github.com/suse-edge/edge-image-builder/pull/64#discussion_r1439627457) the encryption/decryption of the `RegCode` will happen in a different PR after #64.

Would it also be a good idea to handle the `AddRepo` enhancement described [here](https://github.com/suse-edge/edge-image-builder/pull/64#discussion_r1439624823) in a different PR also after #64, so that we can have a working base on which we can build on top of?

pkg/image/definition.go
```diff
@@ -60,6 +60,13 @@ type OperatingSystem struct {
 	Users      []OperatingSystemUser `yaml:"users"`
 	Systemd    Systemd               `yaml:"systemd"`
 	Suma       Suma                  `yaml:"suma"`
+	Packages   Packages              `yaml:"packages"`
+}
+
+type Packages struct {
+	PKGList  []string `yaml:"packageList"`
+	AddRepos []string `yaml:"additionalRepos"`
+	RegCode  string   `yaml:"registrationCode"`
```
atanasdinov, line 69: I'm not sure if this is the best structure to have. The more I look at it, the more I find the repetition (`packages: -> packageList`) weird.

The only alternative I can think of is something like this:
```
  rpm:
    packages:
    - package1
    - package2
    additionalRepositories:
    - repo1
    - repo2
    registrationCode: INTERNAL-USE-ONLY-foo-bar
```

Perhaps I'm nitpicking though so I'd love if @rdoxenham and @jdob also look at this.

pkg/image/definition.go
```diff
@@ -60,6 +60,13 @@ type OperatingSystem struct {
 	Users      []OperatingSystemUser `yaml:"users"`
 	Systemd    Systemd               `yaml:"systemd"`
 	Suma       Suma                  `yaml:"suma"`
+	Packages   Packages              `yaml:"packages"`
+}
+
+type Packages struct {
+	PKGList  []string `yaml:"packageList"`
+	AddRepos []string `yaml:"additionalRepos"`
```
atanasdinov, line 68: The `AddRepos` variable name sounds like a boolean flag so I'd suggest to rename it to `AdditionalRepos` or `AdditionalRepositories`.

rdoxenham: `AdditionalRepos` gets my vote here. I think the `AddRepos` thing comes from the `zypper` syntax, but I agree that we should clarify this.

pkg/image/definition.go
rdoxenham: Yeah, I see what you're saying @atanasdinov, but I still like the `operatingSystem.packages` bit. I agree that `packages --> packageList` seems a little strange/repetitive, but I'd also like to avoid the use of `rpms` as we have a dedicated `rpms/` directory and that may also cause confusion.

rdoxenham (changes requested):
Very small nit following @atanasdinov proposal. Once fixed I'll LGTM. Thank you!

pkg/image/definition.go
ipetrov117: I also like the `packages` bit, mainly for the reasons that @rdoxenham described. However, in order to avoid the `packages --> packageList` confusion, we could perhaps do something like this:

```yaml
packages:
  install:
  - package1
  - package2
  additionalRepositories:
  - repo1
  - repo2
  registrationCode: INTERNAL-USE-ONLY-foo-bar
```
@atanasdinov WDYT?

atanasdinov: @rdoxenham You're right that `rpm` potentially collides with the `rpm` directory, however, we also have to keep in mind that we still want to side load RPMs, right? So it is more confusing (to me at least) if you have different names for the different options of the same functionality.

@ipetrov117 The issue with `install` is that it is a verb which doesn't quite sit right with `additionalRepositories` and `registrationCode` on the side... I don't have any better ideas though so I will approve to unblock.

atanasdinov (approved):

rdoxenham (approved):
LGTM, thank you!

pkg/image/definition.go
rdoxenham: Yeah... verb aside, I think the flow does still make sense...

1. This is the definition of the desired state of the operating system packages
2. These are the packages I want to list to install
3. Here's where you can potentially pull these packages from, or help solve for dependencies
4. Here's my registration code so you can pull from SLE Micro & PackageHub repo's

We should likely let this one go :)

jdob:
> @jdob as discussed [here](https://github.com/suse-edge/edge-image-builder/pull/64#discussion_r1439627457) there will be an issue for the `RegCode` being included as plain text which will be handled by a different PR in the future.
> 
> Would it also be a good idea to handle the `AddRepo` enhancement described [here](https://github.com/suse-edge/edge-image-builder/pull/64#discussion_r1439624823) in a different PR also after #64, so that we can have a working base on which we can build on top of?

Agreed, we can revisit the repo definition thing, no need to block this PR.

jdob (approved):

ipetrov117:
> Agreed, we can revisit the repo definition thing, no need to block this PR.

Okay, I will directly open a PR with the repo definition logic once we review and merge #64

PR #96, opened 2024-01-11
Registry Implementation

dbw7:
Preliminary version of the registry implementation for feedback and updating before merging.
First version is missing tests.


pkg/combustion/templates/13-embedded-registry.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+set -euo pipefail
+
+mkdir -p /etc/hauler
+
+mv haul.tar.zst /etc/hauler/haul.tar.zst
```
rdoxenham, line 6: I don't think that `/etc/hauler/` is the right place to store the Hauler archive. Can we consider moving this to `/var/lib/` please? That's where the majority of the other mutable data lives, and is not subject to btrfs snapshot rollbacks on SLE Micro.

pkg/combustion/templates/13-embedded-registry.sh
```diff
@@ -0,0 +1,26 @@
...
+
+cat <<- EOF > /etc/systemd/system/registry.service
+  [Unit]
+  Description=Hauler Load and Serve
+  After=network.target
+
+  [Service]
+  Type=simple
+  User=root
+  WorkingDirectory=/etc/hauler
```
rdoxenham, line 17: Same here if this is used for the archive location.

pkg/combustion/templates/13-embedded-registry.sh
```diff
@@ -0,0 +1,26 @@
...
+  ExecStartPre=/usr/bin/hauler store load haul.tar.zst
+  ExecStart=/usr/bin/hauler store serve
+  Restart=on-failure
+
+  [Install]
+  WantedBy=multi-user.target
+EOF
+
+systemctl enable registry.service
```
rdoxenham, line 26: Can we call this "local-registry" or something? I think "registry" is a little vague.

pkg/combustion/templates/13-embedded-registry.sh
dbw7: Will update 👍

dbw7: After testing, I think this won't work as `/var/lib/` doesn't seem to be available during combustion

rdoxenham (approved):
LGTM, thanks!

pkg/combustion/templates/13-embedded-registry.sh
atanasdinov: What about `/opt`? @rdoxenham

Dockerfile
```diff
@@ -21,7 +21,13 @@ RUN curl -o nmc-aarch64 -L https://github.com/suse-edge/nm-configurator/releases
...
     curl -o nmc-x86_64 -L https://github.com/suse-edge/nm-configurator/releases/download/v0.2.0/nmc-linux-x86_64 && \
     chmod +x nmc-x86_64 && \
-    cp nmc-$(uname -m) /usr/local/bin/nmc
+    cp nmc-$(uname -m) /usr/local/bin/nmc && \
+    curl -o hauler-amd64.tar -L https://github.com/rancherfederal/hauler/releases/download/v0.4.2/hauler_0.4.2_linux_amd64.tar.gz && \
+    tar -xf hauler-amd64.tar && \
+    mv hauler /usr/bin/hauler-amd64 && \
+    curl -o hauler-arm64.tar -L https://github.com/rancherfederal/hauler/releases/download/v0.4.2/hauler_0.4.2_linux_arm64.tar.gz && \
+    tar -xf hauler-arm64.tar && \
+    mv hauler /usr/bin/hauler-arm64
```
atanasdinov, line 30: I'd suggest to move this to a separate layer since it's different from the network configurations and the thought of constantly adding new dependencies and invalidating existing layers seems off since this can grow exponentially. I know @jdob is not generally a fan of many layers though. Thoughts?

pkg/combustion/templates/13-embedded-registry.sh
atanasdinov: Ignore this, you've already done it.

rdoxenham: Yep @dbw7 and I spoke about this offline. `/opt/hauler/` works well.

pkg/combustion/registry.go
```diff
@@ -0,0 +1,129 @@
...
+
+func writeHaulerManifest(ctx *image.Context) error {
+	haulerManifestYamlFile := filepath.Join(ctx.BuildDir, haulerManifestYamlName)
+
+	data, err := template.Parse(haulerManifestYamlName, haulerManifest, ctx.ImageDefinition.EmbeddedArtifactRegistry)
+	if err != nil {
+		return fmt.Errorf("applying template to %s: %w", haulerManifestYamlName, err)
+	}
+
+	if err := os.WriteFile(haulerManifestYamlFile, []byte(data), fileio.ExecutablePerms); err != nil {
```
atanasdinov, line 78: Does it have to be executable?

pkg/combustion/registry.go
```diff
@@ -0,0 +1,129 @@
...
+	return nil
+}
+
+func populateHaulerStore(ctx *image.Context, haulerExecutable string) error {
+	haulerManifestPath := filepath.Join(ctx.BuildDir, haulerManifestYamlName)
+
+	cmd := exec.Command(haulerExecutable, "store", "sync", "--files", haulerManifestPath)
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("populating hauler store: %w: %s", err, string(output))
```
atanasdinov, line 91: Embedding a presumably very large output is not something we want into the error itself. Can we create a log file for it and use it for the command's stdout and stderr?

pkg/combustion/registry.go
```diff
@@ -0,0 +1,129 @@
...
+	return nil
+}
+
+func generateHaulerTar(ctx *image.Context, haulerExecutable string) error {
+	haulerTarDest := filepath.Join(ctx.CombustionDir, haulerTarName)
+
+	cmd := exec.Command(haulerExecutable, "store", "save", "--filename", haulerTarDest)
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("creating hauler registry tar: %w: %s", err, string(output))
```
atanasdinov, line 103: Same as above.

pkg/combustion/templates/13-embedded-registry.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+set -euo pipefail
+
+mkdir /opt/hauler
+
+mv haul.tar.zst /opt/hauler/haul.tar.zst
```
atanasdinov, line 6: It's better to template the input since a change to the constant will completely invalidate the script.

pkg/combustion/registry.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/suse-edge/edge-image-builder/pkg/fileio"
+	"github.com/suse-edge/edge-image-builder/pkg/image"
+	"github.com/suse-edge/edge-image-builder/pkg/log"
+	"github.com/suse-edge/edge-image-builder/pkg/template"
+)
+
+const (
+	haulerManifestYamlName = "hauler-manifest.yaml"
+	registryScriptName     = "13-embedded-registry.sh"
+	haulerTarName          = "haul.tar.zst"
```
atanasdinov, line 19: Can we rename this to `container-registry.tar.zst` or something similar instead?

atanasdinov (changes requested):
Looks good for the most part but there are few things which need to address.

The main issue with having the implementation in the combustion package is that it is not testable at all. We discussed having a separate package for it and plugging in an interface. Was there a problem with this approach?

pkg/combustion/templates/13-embedded-registry.sh
jdob: That's a good call. I'd go a step further and make sure we include some sort of designation that shows it came from EIB, since `local-registry` still won't give people a ton of hints where it came from.

Dockerfile
jdob: I actually had the same thought as Atanas when I saw this. I think layers being separated by functionality makes sense.

pkg/combustion/registry.go
dbw7: No it doesn't, not sure why I selected that originally, will update.

pkg/combustion/templates/13-embedded-registry.sh
rdoxenham: Yeah I'm a little hesitant to suggest that we have "eib" in there, but I'll go with the majority here if we think it needs to have it.

pkg/combustion/registry.go
```diff
@@ -0,0 +1,173 @@
...
+	if err != nil {
+		return "", fmt.Errorf("writing registry script: %w", err)
+	}
+
+	return registryScriptName, nil
+}
+
+func createRegistryCommand(ctx *image.Context, commandName string, args []string) (*exec.Cmd, *os.File, error) {
+	fullLogFilename := filepath.Join(ctx.BuildDir, registryLogFileName)
+	logFile, err := os.OpenFile(fullLogFilename, os.O_APPEND|os.O_CREATE|os.O_WRONLY, fileio.NonExecutablePerms)
```
atanasdinov, line 163: I'm a little bit conflicted on this one. Not sure how large the output is but if we reuse the same file then it is harder for us to track down the issues once something fails. What do you think?

pkg/combustion/registry_test.go
```diff
@@ -0,0 +1,155 @@
...
+	require.NoError(t, err)
+
+	foundBytes, err := os.ReadFile(manifestFileName)
+	require.NoError(t, err)
+	found := string(foundBytes)
+	assert.Contains(t, found, "rgcrprod.azurecr.us/longhornio/longhorn-ui:v1.5.1")
+	assert.Contains(t, found, "https://releases.rancher.com/server-charts/stable")
+}
+
+func TestWriteHaulerManifestEmptyManifest(t *testing.T) {
```
atanasdinov, line 57: Is this even valid? Have we tested what happens when the file doesn't contain any images and / or charts?

```
apiVersion: content.hauler.cattle.io/v1alpha1
kind: Images
metadata:
  name: Embedded-Registry-Images
spec:
  images:
---
apiVersion: content.hauler.cattle.io/v1alpha1
kind: Charts
metadata:
  name: Embedded-Registry-Charts
spec:
  charts:
```

pkg/combustion/registry_test.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+func TestCopyHaulerBinary(t *testing.T) {
+	// Setup
+	ctx, teardown := setupContext(t)
+	defer teardown()
+
+	haulerExecutable := filepath.Join(ctx.BuildDir, "hauler")
+	err := os.WriteFile(haulerExecutable, []byte(""), fileio.ExecutablePerms)
+	assert.NoError(t, err)
```
atanasdinov, line 133: We can't continue if this fails.
```suggestion
	require.NoError(t, err)
```

pkg/combustion/templates/hauler-manifest.yaml.tpl
```diff
@@ -0,0 +1,24 @@
+apiVersion: content.hauler.cattle.io/v1alpha1
+kind: Images
+metadata:
+  name: Embedded-Registry-Images
```
atanasdinov, line 4: Can we lower case the name? It feels weird this way.

pkg/combustion/templates/hauler-manifest.yaml.tpl
```diff
@@ -0,0 +1,24 @@
...
+    - name: {{ .Name }}
+      {{- if .SupplyChainKey }}
+      key: {{ .SupplyChainKey }}
+      {{- end }}
+    {{- end }}
+---
+apiVersion: content.hauler.cattle.io/v1alpha1
+kind: Charts
+metadata:
+  name: Embedded-Registry-Charts
```
atanasdinov, line 17: Same as above.

pkg/combustion/registry.go
dbw7: I was also debating the same thing, but ultimately decided that it might be too much to have 2 different log files, but honestly I’m fine with that if it’s gonna be easier in the long run when it comes to debugging.

pkg/combustion/registry_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+	manifestFileName := filepath.Join(ctx.BuildDir, haulerManifestYamlName)
+	_, err = os.Stat(manifestFileName)
+	require.NoError(t, err)
+
+	foundBytes, err := os.ReadFile(manifestFileName)
+	require.NoError(t, err)
+	found := string(foundBytes)
+	assert.Contains(t, found, "rgcrprod.azurecr.us/longhornio/longhorn-ui:v1.5.1")
+	assert.Contains(t, found, "https://releases.rancher.com/server-charts/stable")
```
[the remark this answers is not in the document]
dbw7: Will update 👍

pkg/combustion/registry_test.go
dbw7: I was doing this just to show how the function would behave with an empty input, but in reality, it should never actually reach this step in that state. Should I remove the test?

pkg/combustion/registry_test.go
```diff
@@ -0,0 +1,155 @@
...
+	require.NoError(t, err)
+
+	registryScriptPath := filepath.Join(ctx.CombustionDir, "13-embedded-registry.sh")
+	_, err = os.Stat(registryScriptPath)
+	require.NoError(t, err)
+
+	foundBytes, err := os.ReadFile(registryScriptPath)
+	require.NoError(t, err)
+	found := string(foundBytes)
+	assert.Contains(t, found, registryTarName)
```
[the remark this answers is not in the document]
dbw7: Will do 👍

pkg/combustion/registry.go
atanasdinov: Feel free to include an example of what it currently looks like so we can decide based on that.

pkg/combustion/registry.go
```diff
@@ -0,0 +1,173 @@
...
+
+//go:embed templates/13-embedded-registry.sh.tpl
+var registryScript string
+
+func configureRegistry(ctx *image.Context) ([]string, error) {
+	if image.IsEmbeddedArtifactRegistryEmpty(ctx.ImageDefinition.EmbeddedArtifactRegistry) {
+		log.AuditComponentSkipped(registryComponentName)
+		return nil, nil
+	}
+	haulerExecutable := filepath.Join("usr", "bin", fmt.Sprintf("hauler-%s", ctx.ImageDefinition.Image.Arch.Short()))
```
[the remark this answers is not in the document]
dbw7: That was actually an issue I was thinking about, I’d like to talk with you to see your thoughts regarding solutions.

pkg/combustion/registry_test.go
dbw7: Just tested: if either value is empty, it is just skipped, so I'll remove the test.

dbw7:
I added `haulerBinaryPath string` as a parameter to `copyHaulerBinary` because testing it without being able to specify where to look was unreliable.

pkg/combustion/templates/13-embedded-registry.sh.tpl
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+set -euo pipefail
+
+mkdir /opt/hauler
+
+mv {{ .EmbeddedRegistryTar }} /opt/hauler/{{ .EmbeddedRegistryTar }}
+mv hauler /usr/bin/hauler
```
atanasdinov, line 7: @rdoxenham is it better to have this under `/usr/bin` or `/usr/local/bin`?

atanasdinov (approved):

atanasdinov:
> I added `haulerBinaryPath string` as a parameter to `copyHaulerBinary` because testing it without being able to specify where to look was unreliable.

This is also the reason why extracting the functionality in a separate package is better. Can be done in a follow up, though.

pkg/combustion/templates/13-embedded-registry.sh.tpl
rdoxenham: `/usr/local/bin` is better here as it's excluded from rollbacks, i.e. it's part of the persistent/mutable parts of the OS. We may need to mount it via combustion though, so we should test this.

Dockerfile
```diff
@@ -31,8 +31,18 @@ RUN curl -o nmc-aarch64 -L https://github.com/suse-edge/nm-configurator/releases
     chmod +x nmc-x86_64 && \
     cp nmc-$(uname -m) /usr/local/bin/nmc
 
+
+RUN curl -o hauler-amd64.tar -L https://github.com/rancherfederal/hauler/releases/download/v0.4.2/hauler_0.4.2_linux_amd64.tar.gz && \
```
agracey, line 35: I was just chatting with Craig about sbux. I think the risk that we have is IBM coming in a selling them on the same solution they've developed for

agracey, line 35: This means that we will also need to build hauler in OBS/IBS if we want to get EIB into the suse registry.

atanasdinov (approved):

PR #98, opened 2024-01-14
Adding unattended mode and disk selection

rdoxenham:
This PR adds two new features:

1.  Unattended mode, where we skip the GRUB selection of install or boot from disk and automatically run the installation after 3 seconds (we may want to make this larger to enable users to override this). The default behaviour is to wait at the GRUB prompt to avoid data loss. For fully automated installs we need a way of overriding this.

2. Disk selection, where the user can tell the installer which disk to target (needs to be block special, e.g. /dev/sda), which has two primary benefits - firstly, if the system has more than one disk it allows the user to make this choice outside at image build time, not at boot time, and secondly, if the user has selected a disk to install to it forces the confirmation of wiping anything found on the target disk.

With these two options we can enable fully unattended and automated provisioning via EIB, this requires no user intervention.  Important note, (1) above works today with the default SLE Micro SelfInstall ISO, but (2) requires https://github.com/SUSE/kiwi_sle15/pull/6 to land first. However, this code can land without impacting existing images without this patch, as the kernel cmdline argument will just be ignored.

### DIFF
--- a/docs/building-images.md
+++ b/docs/building-images.md
@@ -37,6 +37,8 @@ The operating system configuration section is entirely optional.
 The following describes the possible options for the operating system section:
 ```yaml
 operatingSystem:
+  installDevice: /path/to/disk
+  unattended: false
   kernelArgs:
   - arg1
   - arg2
@@ -56,6 +58,11 @@ operatingSystem:
       - serviceX
 ```
 
+* `installDevice` - Optional; only for ISO images - specifies the disk that should be used as the install
+  device. This needs to be block special, and will default to automatically wipe any data found on the disk.
+* `unattended` - Optional; only for ISO images - forces GRUB override to automatically install the operating
+  system rather than prompting user to begin the installation. In combination with `installDevice` can create
+  a fully unattended and automated install. Beware of creating boot loops and data loss with these options.
 * `kernelArgs` - Optional; Provides a list of flags that should be passed to the kernel on boot.
 * `users` - Optional; Defines a list of operating system users to be created. Each entry is made up of
   the following fields:
--- a/docs/building-images.md
+++ b/docs/building-images.md
@@ -60,9 +60,11 @@ operatingSystem:
 
 * `installDevice` - Optional; only for ISO images - specifies the disk that should be used as the install
   device. This needs to be block special, and will default to automatically wipe any data found on the disk.
+  If left omitted, the user will still have to select the disk to install to (if >1 found) and confirm wipe.
 * `unattended` - Optional; only for ISO images - forces GRUB override to automatically install the operating
   system rather than prompting user to begin the installation. In combination with `installDevice` can create
   a fully unattended and automated install. Beware of creating boot loops and data loss with these options.
+  If left omiitted (or set to `false`) the user will still have to choose to install via the GRUB menu.
 * `kernelArgs` - Optional; Provides a list of flags that should be passed to the kernel on boot.
 * `users` - Optional; Defines a list of operating system users to be created. Each entry is made up of
   the following fields:
--- a/pkg/build/iso_test.go
+++ b/pkg/build/iso_test.go
@@ -129,6 +129,13 @@ func TestWriteIsoScript_Rebuild(t *testing.T) {
 	defer teardown()
 	builder := Builder{context: ctx}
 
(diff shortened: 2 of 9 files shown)


rdoxenham:
I will follow this up with some tests in the next few days.

pkg/image/testdata/full-valid-example.yaml
```diff
@@ -5,6 +5,8 @@ image:
   baseImage: slemicro5.5.iso
   outputImageName: eibimage.iso
 operatingSystem:
+  installDevice: /dev/sda
+  unattended: true
```
atanasdinov, line 9: Can we update the [config parsing test](https://github.com/suse-edge/edge-image-builder/blob/main/pkg/image/definition_test.go#L11) here? It's what this definition file is used for.

docs/building-images.md
```diff
@@ -56,6 +58,11 @@ operatingSystem:
       - serviceX
 ```
 
+* `installDevice` - Optional; only for ISO images - specifies the disk that should be used as the install
+  device. This needs to be block special, and will default to automatically wipe any data found on the disk.
+* `unattended` - Optional; only for ISO images - forces GRUB override to automatically install the operating
```
atanasdinov, line 63: Aside from these two being ISO only, do we want to validate anything else? Should we raise a GitHub issue to add the validation steps or you prefer to include those with the tests you've mentioned you've planned?

atanasdinov (approved):

docs/building-images.md
rdoxenham: I've just added them... if the user now selects "raw" then neither of these options are allowed.

pkg/image/testdata/full-valid-example.yaml
rdoxenham: Done!

atanasdinov (approved):
LGTM!

e-minguez:
As a suggestion maybe we can have another parameter (or not, just do it if the unattended flag is set) that ejects the /dev/cdrom device at combustion time (simply eject /dev/cdrom) if it is unattended so it prevents infinite reinstallation loops in case of a reboot

rdoxenham:
> As a suggestion maybe we can have another parameter (or not, just do it if the unattended flag is set) that ejects the /dev/cdrom device at combustion time (simply eject /dev/cdrom) if it is unattended so it prevents infinite reinstallation loops in case of a reboot

Thanks @e-minguez, I'm not sure how this could work with anything other than a real CDROM? I mean, does `eject` work for virtual media? I don't think it would have any impact to a USB stick, right?

pkg/build/iso_test.go
```diff
@@ -157,6 +164,15 @@ func TestWriteIsoScript_Rebuild(t *testing.T) {
 
 	expectedCombustionDir := ctx.CombustionDir
 	assert.Contains(t, found, fmt.Sprintf("COMBUSTION_DIR=%s", expectedCombustionDir))
+
+	// Make sure that unattended mode is configured for GRUB
+	assert.Contains(t, found, "set timeout=")
```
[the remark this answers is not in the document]
rdoxenham: Thanks @atanasdinov - I've fixed these, but will leave this comment unresolved for others to follow the example that you've suggested.

e-minguez:
> > As a suggestion maybe we can have another parameter (or not, just do it if the unattended flag is set) that ejects the /dev/cdrom device at combustion time (simply eject /dev/cdrom) if it is unattended so it prevents infinite reinstallation loops in case of a reboot
> 
> Thanks @e-minguez, I'm not sure how this could work with anything other than a real CDROM? I mean, does `eject` work for virtual media? I don't think it would have any impact to a USB stick, right?

You are right, it would be useless for USB sticks but IIRC it would work for virtual media.

docs/building-images.md
jdob, line 63: Can you add in what the default is when it's not specified?

docs/building-images.md
```diff
@@ -56,6 +58,11 @@ operatingSystem:
       - serviceX
 ```
 
+* `installDevice` - Optional; only for ISO images - specifies the disk that should be used as the install
```
jdob, line 61: Same here on the default, can you describe the behavior when this is omitted?

jdob (approved):

PR #99, opened 2024-01-15
Adding support for setting timezone and chrony sources

rdoxenham:
This PR adds the ability to specify the `timezone` you wish your system to be setup with (see `timedatectl list-timezones`) and also allows the user to specify a list of `chronyPools` and `chronyServers`, which will populate a new file `/etc/chrony.d/eib-sources.conf` (removing the out of the box default SUSE NTP pool).

### DIFF
--- a/docs/building-images.md
+++ b/docs/building-images.md
@@ -37,6 +37,8 @@ The operating system configuration section is entirely optional.
 The following describes the possible options for the operating system section:
 ```yaml
 operatingSystem:
+  installDevice: /path/to/disk
+  unattended: false
   kernelArgs:
   - arg1
   - arg2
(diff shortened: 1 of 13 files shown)


docs/building-images.md
```diff
@@ -37,6 +37,13 @@ The operating system configuration section is entirely optional.
 The following describes the possible options for the operating system section:
 ```yaml
 operatingSystem:
+  time:
+    timezone: Europe/London
+  chronyPools:
```
jdob, line 42: My gut feeling is that the two chrony* sections should be under the umbrella `time`.

pkg/combustion/time.go
```diff
@@ -0,0 +1,51 @@
...
+
+	"github.com/suse-edge/edge-image-builder/pkg/fileio"
+	"github.com/suse-edge/edge-image-builder/pkg/image"
+	"github.com/suse-edge/edge-image-builder/pkg/log"
+	"github.com/suse-edge/edge-image-builder/pkg/template"
+)
+
+const (
+	timeComponentName = "time"
+	timeScriptName    = "09-time-setup.sh"
```
jdob, line 17: Is there a reason you put this so early in the script execution? I noticed it in `combustion.go` and now here with the prefix.

docs/building-images.md
jdob: Looking at the code, the implementation looks correct, just not this example.

jdob (changes requested):

rdoxenham:
Well and truly broke this PR. Will resubmit, and also buy a book on git ;-)

PR #100, opened 2024-01-15
Split kubernetes artefacts into installers and images

atanasdinov:
- Splits the `kubernetes/` directory into `kubernetes/install` and `kubernetes/images`
- `kubernetes/install` will be used during combustion for the `INSTALL_RKE2_ARTIFACT_PATH` env variable necessary for airgap installs
- `kubernetes/images` will be used to transfer the airgap images to `/var/lib/rancher/rke2/agent/images/`

### DIFF
--- a/pkg/image/context.go
+++ b/pkg/image/context.go
@@ -17,7 +17,7 @@ type kubernetesScriptInstaller interface {
 }
 
 type kubernetesArtefactDownloader interface {
-	DownloadArtefacts(kubernetes Kubernetes, arch Arch, destinationPath string) error
+	DownloadArtefacts(kubernetes Kubernetes, arch Arch, destinationPath string) (string, string, error)
 }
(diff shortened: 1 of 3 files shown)


pkg/image/context.go
```diff
@@ -17,7 +17,7 @@ type kubernetesScriptInstaller interface {
 }
 
 type kubernetesArtefactDownloader interface {
-	DownloadArtefacts(kubernetes Kubernetes, arch Arch, destinationPath string) error
+	DownloadArtefacts(kubernetes Kubernetes, arch Arch, destinationPath string) (string, string, error)
```
jdob, line 20: We talked about this in chat. Noting it here that they aren't named in the interface but are in the implementation. Your call if you want to change it.

jdob (approved):

PR #101, opened 2024-01-15
Hook up the ISO flow to the raw modifications

jdob:


pkg/build/templates/modify-raw-image.sh.tpl
```diff
@@ -2,31 +2,36 @@
...
-guestfish --format=raw --rw -a {{.OutputImage}} -i <<'EOF'
+guestfish --format=raw --rw -a {{.ImagePath}} -i <<'EOF'
   # Enables write access to the read only filesystem
   sh "btrfs property set / ro false"
 
   {{ if ne .ConfigureGRUB "" }}
-      # GRUB configuration
-      {{ .ConfigureGRUB }}
+  {{ .ConfigureGRUB }}
   {{ end }}
```
atanasdinov, line 21: If we drop the comment, do we even need the if statement?

pkg/build/raw.go
```diff
@@ -24,13 +24,22 @@ var modifyRawImageTemplate string
...
+	if err := b.modifyRawImage(b.generateOutputImageFilename(), true, true); err != nil {
+		// modifyRawImage will wrap the error, simply return it here
+		return err
+	}
+
+	return nil
+}
+
+func (b *Builder) modifyRawImage(imagePath string, includeCombustion, renameFilesystem bool) error {
+
```
atanasdinov, line 41: Nit: Leaving an empty line at the start of a function is considered bad practice.

pkg/build/raw.go
```diff
@@ -24,13 +24,22 @@ var modifyRawImageTemplate string
...
 			b.context.ImageDefinition.Image.BaseImage, b.generateOutputImageFilename(), err)
 	}
 
-	if err = b.writeModifyScript(); err != nil {
+	if err := b.modifyRawImage(b.generateOutputImageFilename(), true, true); err != nil {
+		// modifyRawImage will wrap the error, simply return it here
+		return err
+	}
+
+	return nil
```
atanasdinov, line 37: We can just `return b.modifyRawImage(b.generateOutputImageFilename(), true, true)` if wrapping is not necessary.

pkg/build/iso.go
```diff
@@ -146,3 +153,23 @@ func (b *Builder) createIsoCommand(logFilename, scriptName string) (*exec.Cmd, *
...
+		}
+		return nil
+	}
+
+	rawExtractPath := filepath.Join(b.context.BuildDir, rawExtractDir)
+	if err := filepath.Walk(rawExtractPath, f); err != nil {
+		return "", fmt.Errorf("traversing raw extract directory %s: %w", rawExtractPath, err)
+	}
+
+	return filepath.Join(rawExtractPath, foundFile), nil
```
atanasdinov, line 174: What if `foundFile` is still an empty string here?

pkg/build/raw_test.go
```diff
@@ -43,60 +36,91 @@ func TestCreateRawImageCopyCommand(t *testing.T) {
...
 func TestCreateModifyCommand(t *testing.T) {
 	// Setup
-	builder := Builder{
-		context: &image.Context{
-			BuildDir: "build-dir",
-		},
-	}
+	ctx, teardown := setupContext(t)
+	defer teardown()
+	builder := Builder{context: ctx}
```
atanasdinov, line 115: This was intentionally not creating a fully populated context since it is not needed and not affecting the scenario we wanted to test. Is there a reason for changing it?

pkg/build/raw_test.go
```diff
@@ -43,60 +36,91 @@ func TestCreateRawImageCopyCommand(t *testing.T) {
...
-				},
+	ctx, teardown := setupContext(t)
+	defer teardown()
+	ctx.ImageDefinition = &image.Definition{
+		Image: image.Image{
+			OutputImageName: "output-image",
+		},
+		OperatingSystem: image.OperatingSystem{
+			KernelArgs: []string{"alpha", "beta"},
+		},
```
atanasdinov, line 47: These seem to be obsolete.

pkg/build/raw_test.go
```diff
@@ -14,17 +15,9 @@ import (
...
-				Image: image.Image{
-					BaseImage:       "base-image",
-					OutputImageName: "build-image",
-				},
-			},
-		},
-	}
+	ctx, teardown := setupContext(t)
+	defer teardown()
+	builder := Builder{context: ctx}
```
atanasdinov, line 20: Same comment as one of the above. If we don't use the directories, why do we need to build a context? The main reason behind not using `setupContext()` is that it will eventually be easier to get rid of once we figure out whether the structure of the `build` package is what we want it to be.

pkg/build/templates/modify-raw-image.sh.tpl
jdob: Fair, it doesn't technically need to be there. I like it for the readability; it keys in new developers to the idea that GRUB won't always be configured.

pkg/build/iso.go
jdob: Good pickup.

pkg/build/raw.go
jdob: Fixed; you've pointed it out before and I normally check for it.

pkg/build/raw_test.go
jdob: We do use the directories; `builder.generateBaseImageFilename()` and `builder.generateOutputImageFilename()` both use them.

atanasdinov: Those use string concatenation - I see nothing wrong with hardcoding these values in the tests.

pkg/build/raw_test.go
jdob: Nope, was just changing for consistency. I can revert if you'd like.

pkg/build/iso.go
```diff
@@ -146,3 +153,27 @@ func (b *Builder) createIsoCommand(logFilename, scriptName string) (*exec.Cmd, *
...
+		return nil
+	}
+
+	rawExtractPath := filepath.Join(b.context.BuildDir, rawExtractDir)
+	if err := filepath.Walk(rawExtractPath, f); err != nil {
+		return "", fmt.Errorf("traversing raw extract directory %s: %w", rawExtractPath, err)
+	}
+
+	if foundFile == "" {
+		return "", fmt.Errorf("unable to find a raw image in: %s", rawExtractDir)
```
atanasdinov, line 175: Can we add a test case for this?

jdob: Yeah, that's my bad. I initially thought I'd have to come back when I got around to building a test ISO, but I can test this without the need for that.

atanasdinov (approved):

PR #102, opened 2024-01-15
Adding support for setting timezone and chrony sources

rdoxenham:
This PR adds the ability to specify the timezone you wish your system to be setup with (see `timedatectl list-timezones`) and also allows the user to specify a list of `chronyPools` and `chronyServers`, which will populate a new file `/etc/chrony.d/eib-sources.conf` (removing the out of the box default SUSE NTP pool).

The setting of the timezone and chrony sources is executed early on in the boot process to ensure there's no issues with SSL, where later stages, e.g. SUMA/Elemental registration may break.

### DIFF
--- a/pkg/combustion/templates/09-time-setup.sh.tpl
+++ b/pkg/combustion/templates/09-time-setup.sh.tpl
@@ -1,0 +1,18 @@
+#!/bin/bash
+set -euo pipefail
+
+{{ if .Timezone -}}
+ln -sf /usr/share/zoneinfo/{{ .Timezone }} /etc/localtime
+{{ end -}}
+
+{{ if or (gt (len .ChronyPools) 0) (gt (len .ChronyServers) 0) }}
+rm -f /etc/chrony.d/pool.conf
+{{ end -}}
+
+{{ range .ChronyPools -}}
+echo "pool {{ . }} iburst" >> /etc/chrony.d/eib-sources.conf
+{{ end -}}
+
+{{ range .ChronyServers -}}
+echo "server {{ . }} iburst" >> /etc/chrony.d/eib-sources.conf
+{{ end -}}
--- a/pkg/combustion/time_test.go
+++ b/pkg/combustion/time_test.go
@@ -1,0 +1,78 @@
+package combustion
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/suse-edge/edge-image-builder/pkg/fileio"
+	"github.com/suse-edge/edge-image-builder/pkg/image"
+)
+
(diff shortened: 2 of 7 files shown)


pkg/combustion/time_test.go
```diff
@@ -0,0 +1,78 @@
...
+func TestConfigureTime_NoConf(t *testing.T) {
+	// Setup
+	ctx, teardown := setupContext(t)
+	defer teardown()
+
+	ctx.ImageDefinition = &image.Definition{
+		OperatingSystem: image.OperatingSystem{
+			Time: image.Time{},
+		},
+	}
```
atanasdinov, line 23: We can just define an empty context with `var ctx image.Context` and skip the directory setup included in `setupContext()`.

pkg/combustion/templates/09-time-setup.sh.tpl
```diff
@@ -0,0 +1,18 @@
...
+{{ if .Timezone -}}
+ln -sf /usr/share/zoneinfo/{{ .Timezone }} /etc/localtime
+{{ end -}}
+
+{{ if or (gt (len .ChronyPools) 0) (gt (len .ChronyServers) 0) }}
+rm -f /etc/chrony.d/pool.conf
+{{ end -}}
+
+{{ range .ChronyPools -}}
+echo "pool {{ . }} iburst" >> /etc/chrony.d/eib-sources.conf
```
atanasdinov, line 13: Is having a new file all that's needed for this?

atanasdinov (approved):

pkg/combustion/templates/09-time-setup.sh.tpl
rdoxenham: Yep, if you look at `/etc/chrony.conf` it has an `includes` section so anything we throw in here, works:

```
localhost:~ # grep include /etc/chrony.conf
# Also include any directives found in configuration files in /etc/chrony.d
include /etc/chrony.d/*.conf

localhost:~ # cat /etc/chrony.d/eib-sources.conf
pool 2.suse.pool.ntp.org iburst
```

Chrony will then automatically pick up this pool:

```
localhost:~ # chronyc sources
MS Name/IP address         Stratum Poll Reach LastRx Last sample
===============================================================================
^+ time.cloudflare.com           3  10   377   341   +836us[ +993us] +/- 9231us
^* pool.ntp0.cam.ac.uk           2  10   377   328    -64us[  +93us] +/- 7829us
^- time.shf.uk.as44574.net       4  10   377   342   +449us[ +606us] +/-   28ms
^- ntp1.as200552.net             2   9   377   342   -973us[ -816us] +/-   46ms
```

atanasdinov: Thanks!

atanasdinov (approved):

jdob (approved):

PR #104, opened 2024-01-15
Introduced new validation runner and ported image validations

jdob:
This is currently not in use; the existing validation is still what is run.

This isn't complete. I'm not sure we'll keep error and/or component in `FailedValidation`. I'll make that call once I port the rest over and see what it looks like when I replace the existing call. But I wanted to submit the PRs in shorter pieces to make the actual review process less painful.

### DIFF
--- a/pkg/image/validation/image.go
+++ b/pkg/image/validation/image.go
@@ -1,0 +1,85 @@
+package validation
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"slices"
+	"strings"
+
+	"github.com/suse-edge/edge-image-builder/pkg/image"
+)
+
+const (
+	imageComponent = "image"
+)
+
+var validImageTypes = []string{image.TypeISO, image.TypeRAW}
+var validArchTypes = []string{string(image.ArchTypeARM), string(image.ArchTypeX86)}
+
+func validateImage(ctx *image.Context) []FailedValidation {
+	def := ctx.ImageDefinition
+
+	var failures []FailedValidation
+
+	if def.Image.ImageType == "" {
+		failures = append(failures, FailedValidation{
+			userMessage: "The 'imageType' field is required in the 'image' section.",
+			component:   imageComponent,
+		})
+	} else if !slices.Contains(validImageTypes, def.Image.ImageType) {
+		msg := fmt.Sprintf("The 'imageType' field must be one of: %s", strings.Join(validImageTypes, ", "))
+		failures = append(failures, FailedValidation{
+			userMessage: msg,
+			component:   imageComponent,
+		})
+	}
+
+	if def.Image.Arch == "" {
+		failures = append(failures, FailedValidation{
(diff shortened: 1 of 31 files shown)


pkg/image/validation/validation.go
```diff
@@ -0,0 +1,27 @@
+package validation
+
+import (
+	"github.com/suse-edge/edge-image-builder/pkg/image"
+)
+
+type FailedValidation struct {
```
atanasdinov, line 7: Nit: `ValidationFailure` sounds much better in what the structure is conveying in my opinion.

pkg/image/validation/image_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+	testBaseImageFilename := filepath.Join(testImagesDir, "base-image.iso")
+	_, err = os.Create(testBaseImageFilename)
+	require.NoError(t, err)
+
+	tests := map[string]struct {
+		ImageDefinition        image.Definition
+		ExpectedFailedMessages []string
+	}{
+		`complete valid definition`: {
```
atanasdinov, line 32: Nit: Using `` creates string literals (i.e. constants) and not string variables (i.e. ""). Is there a reason to prefer the former over the latter?

pkg/image/validation/image.go
```diff
@@ -0,0 +1,85 @@
...
+
+	"github.com/suse-edge/edge-image-builder/pkg/image"
+)
+
+const (
+	imageComponent = "image"
+)
+
+var validImageTypes = []string{image.TypeISO, image.TypeRAW}
+var validArchTypes = []string{string(image.ArchTypeARM), string(image.ArchTypeX86)}
```
atanasdinov, line 18: Can we move these to the `validateImage()` function or do we expect to reuse them?

atanasdinov (approved):

pkg/image/validation/image_test.go
jdob: Nah, I don't have a preference. I literally just saw this approach and wanted to try it out. But I can also see the argument for code consistency across the project, so if we want to standardize on the existing model of a name field in the struct, I'm completely fine with that.

pkg/image/validation/validation.go
jdob: I went back and forth. I liked this one since in my head I kept thinking of the list of them as "a list of failed validations" but I could just as easily wrap my head around "a list of validation failures".

pkg/image/validation/image.go
jdob: I don't have aspirations of reusing them. My mentality is that I tend to make lists like this outside of the code path to prevent re-initializing the list on each pass through. However, seeing as there's only going to be one pass through the validation, that concern isn't an issue. I'll move them because I like the idea of limiting the scope of them.

jdob:
Closing out to resubmit because I suck at git.

PR #105, opened 2024-01-15
Adding support for setting the system-wide proxy servers

rdoxenham:
This PR adds support for setting the system-wide proxy servers. It takes care of the following in `/etc/sysconfig/proxy`:

1. If either `httpProxy` or `httpsProxy` are set, it forces `PROXY_ENABLED="yes"`
2. If `httpProxy` is set, it will setup `HTTP_PROXY="<httpProxy>"`
3. If `httpsProxy` is set, it will setup `HTTPS_PROXY="<httpsProxy>"`
4. If `noProxy` is set, it will setup `NO_PROXY="<noProxy>"` (if left omitted it will leave default, `localhost, 127.0.0.1`).

This was tested via Squid and works exactly as required. This likely needs to be setup very early in the boot-up procedure as many items may require proxy.

### DIFF
--- a/docs/building-images.md
+++ b/docs/building-images.md
@@ -46,6 +46,10 @@ operatingSystem:
     chronyServers:
       - 10.0.0.1
       - 10.0.0.2
+  proxy:
+    httpProxy: http://10.0.0.1:3128
+    httpsProxy: http://10.0.0.1:3128
+    noProxy: localhost, 127.0.0.1, edge.suse.com
   kernelArgs:
   - arg1
   - arg2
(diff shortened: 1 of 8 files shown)


docs/building-images.md
```diff
@@ -46,6 +46,10 @@ operatingSystem:
     chronyServers:
       - 10.0.0.1
       - 10.0.0.2
+  proxy:
+    httpProxy: http://10.0.0.1:3128
+    httpsProxy: http://10.0.0.1:3128
+    noProxy: localhost, 127.0.0.1, edge.suse.com
```
atanasdinov, line 52: As you've mentioned offline, I'm not sure whether this being a single string is better than a list. Are these values usually copy pasted on a single line?

rdoxenham: I think this is fine as it is and we can adapt later if users would prefer a list.

jdob: Just keep in mind that adapting later is much more complicated for something like the image definition, which represents our API. Not impossible, but I wanted to at least say that the definition is one area where we shouldn't necessarily push off too many decisions until later.

jdob (approved):

PR #110, opened 2024-01-16
Add Kubernetes component

atanasdinov:
- Bootstraps Kubernetes component
- Introduces single node fully airgapped RKE2 setup
- Follow ups will include:
   - Revisiting kubernetes definition i.e. dropping fields such as CNI and multus which can be extracted from the config file
   - HA support
   - k3s support
   - Documentation 😬

### DIFF
--- a/pkg/combustion/kubernetes.go
+++ b/pkg/combustion/kubernetes.go
@@ -1,0 +1,27 @@
+package combustion
+
+import (
+	"fmt"
+
+	"github.com/suse-edge/edge-image-builder/pkg/image"
+	"github.com/suse-edge/edge-image-builder/pkg/log"
+)
+
+const (
+	k8sComponentName = "kubernetes"
+)
+
+func configureKubernetes(ctx *image.Context) ([]string, error) {
+	switch ctx.ImageDefinition.Kubernetes {
+	case "":
+		log.AuditComponentSkipped(k8sComponentName)
+		return nil, nil
+	case image.KubernetesTypeRKE2:
+		panic("implement me")
+	case image.KubernetesTypeK3s:
+		panic("implement me")
+	default:
+		log.AuditComponentFailed(k8sComponentName)
+		return nil, fmt.Errorf("unexpected k8s distro: %s", ctx.ImageDefinition.Kubernetes)
+	}
+}
--- a/pkg/combustion/kubernetes.go
+++ b/pkg/combustion/kubernetes.go
@@ -1,27 +1,149 @@
 package combustion
 
 import (
+	_ "embed"
+	"errors"
 	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"strings"
 
+	"github.com/suse-edge/edge-image-builder/pkg/fileio"
 	"github.com/suse-edge/edge-image-builder/pkg/image"
 	"github.com/suse-edge/edge-image-builder/pkg/log"
+	"github.com/suse-edge/edge-image-builder/pkg/template"
+	"go.uber.org/zap"
 )
 
 const (
-	k8sComponentName = "kubernetes"
+	k8sComponentName  = "kubernetes"
+	k8sConfigDir      = "kubernetes"
+	k8sConfigFile     = "config.yaml"
+	rke2InstallScript = "15-rke2-install.sh"
+)
+
+var (
+	//go:embed templates/15-rke2-installer.sh.tpl
+	rke2InstallerScript string
 )
 
 func configureKubernetes(ctx *image.Context) ([]string, error) {
-	switch ctx.ImageDefinition.Kubernetes {
-	case "":
+	version := ctx.ImageDefinition.Kubernetes.Version
+
+	if version == "" {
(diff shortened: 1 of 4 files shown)


pkg/combustion/templates/15-rke2-installer.sh.tpl
```diff
@@ -0,0 +1,26 @@
...
+export INSTALL_RKE2_TYPE={{ .NodeType }}
+{{- end }}
+
+export INSTALL_RKE2_TAR_PREFIX=/opt/rke2
+export INSTALL_RKE2_ARTIFACT_PATH={{ .InstallPath }}
+
+./rke2_installer.sh
+
+echo "export KUBECONFIG=/etc/rancher/rke2/rke2.yaml" >> ~/.bashrc
+echo "export PATH=${PATH}:/var/lib/rancher/rke2/bin" >> ~/.bashrc
```
jdob, line 24: What's the purpose for adding this to `PATH`? Is it just to get access to `kubectl` when you log into the box?

If this isn't needed for RKE2 to actually run in the first place, I think we should at least consider removing this. Generally speaking, production machines are locked down to all hell, so having this in `PATH` may be frowned upon by the more anal-retentive security admins.

Adding @rdoxenham for both his input based on his Linux experience and ultimately his recommendation as PM.

pkg/combustion/kubernetes.go
```diff
@@ -0,0 +1,149 @@
...
+		zap.S().Info("Kubernetes config file not provided")
+		return "", nil
+	}
+
+	configDir := generateComponentPath(ctx, k8sConfigDir)
+	configFile := filepath.Join(configDir, k8sConfigFile)
+
+	_, err := os.Stat(configFile)
+	if err != nil {
+		if errors.Is(err, fs.ErrNotExist) {
```
jdob, line 136: Is there a reason you manually used `errors.Is` instead of `os.IsNotExist`? I'm sure it's the same thing under the covers (too lazy to check), but feels safer to rely on the `os` package since that's where `Stat?()` came from in the first place.

pkg/combustion/kubernetes_test.go
```diff
@@ -0,0 +1,251 @@
...
+type mockKubernetesScriptInstaller struct {
+	installScript func(distribution, sourcePath, destPath string) error
+}
+
+func (m mockKubernetesScriptInstaller) InstallScript(distribution, sourcePath, destPath string) error {
+	if m.installScript != nil {
+		return m.installScript(distribution, sourcePath, destPath)
+	}
+
+	panic("not implemented")
```
jdob, line 24: I'm personally not a fan of any sort of panic/exit in the middle of unit tests; I think they should always run to completion. It's similar in my head to using `assert` instead of `require` where it makes sense. But it's not a big enough problem to warrant blocking the merge.

jdob (commented):
Minor comments. The only one stopping me from approving is the conversation around if the PATH entry is an issue.

pkg/combustion/kubernetes.go
atanasdinov: If you take a look at the documentation of `os.IsNotExist`, it says:
```
// IsNotExist returns a boolean indicating whether the error is known to
// report that a file or directory does not exist. It is satisfied by
// ErrNotExist as well as some syscall errors.
//
// This function predates errors.Is. It only supports errors returned by
// the os package. New code should use errors.Is(err, fs.ErrNotExist).
```

This is also the reason why I might have changed such usages in the past.

pkg/combustion/kubernetes_test.go
atanasdinov: It is in fact a sanity check which prevents a misuse of the mock itself. We often rely on the returned error as a value to determine whether or not the mock operation is successful so distinguishing intentional errors with a panic is usually the way to go in these cases.

pkg/combustion/kubernetes.go
jdob: Good to know :)

pkg/combustion/templates/15-rke2-installer.sh.tpl
atanasdinov: I indeed use it for the `kubectl` usages and I guess we can live without it. Makes the development stage just slightly easier though. I'm fine with dropping it.

agracey: I would suggest removing it from the code but adding the steps in a "setups for development" best practices doc

rdoxenham: Agreed, we should drop this from EIB. Interestingly, the k3s installer does this automagically, but rke2 doesn't. Either way, this should be a post-deployment action as/when/if required, not something we do by default.

agracey: I've asked about it before and was told K3s does this by default because it's more likely used in local dev environment.

rdoxenham (approved):
LGTM and tested. Thanks!

jdob (approved):
Glad we came to a quick decision on the PATH thing. Approved.

PR #111, opened 2024-01-17
Use pigz instead of gzip for base image archive

ipetrov117:
Move away from using `gzip` when creating the SLE Micro archive when running RPM resolution. Currently we are using `gzip` to create an archive for the data passed by [virt-tar-out](https://github.com/suse-edge/edge-image-builder/blob/main/pkg/rpm/resolver/templates/prepare-base.sh.tpl#L36).

The problem here is that `gzip` is working on a single CPU thread which makes it very slow when archiving large data.
[pigz](https://zlib.net/pigz/) ([doc](https://zlib.net/pigz/pigz.pdf)) offers a `gzip` that can run in parallel. 

Using `pigz` dramatically decreases the time needed to create the SLE Micro archive which is used as a base image for the RPM resolver.

### DIFF
--- a/pkg/rpm/resolver/templates/prepare-base.sh.tpl
+++ b/pkg/rpm/resolver/templates/prepare-base.sh.tpl
@@ -33,7 +33,7 @@ if [ `wc -w <<< $RAW_FILE` -ne 1 ]; then
 	exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


rdoxenham (approved):
Thanks for doing this. Confirming that this works perfectly! Cheers.

atanasdinov (approved):

jdob (approved):

PR #112, opened 2024-01-17
Drop Kubernetes options from the EIB definition in favour of RKE2 config file

atanasdinov:
- Removes CNI and multus configurations from EIB definition
- Uses a Kubernetes (RKE2) configuration file instead
- Uses Cilium as a default CNI if one is not provided
- Drops vSphere support

### DIFF
--- a/pkg/combustion/kubernetes.go
+++ b/pkg/combustion/kubernetes.go
@@ -14,6 +14,7 @@ import (
 	"github.com/suse-edge/edge-image-builder/pkg/log"
 	"github.com/suse-edge/edge-image-builder/pkg/template"
 	"go.uber.org/zap"
+	"gopkg.in/yaml.v3"
 )
 
 const (
@@ -21,6 +22,9 @@ const (
 	k8sConfigDir      = "kubernetes"
 	k8sConfigFile     = "config.yaml"
 	rke2InstallScript = "15-rke2-install.sh"
+
+	cniKey          = "cni"
+	cniDefaultValue = image.CNITypeCilium
 )
 
 var (
@@ -83,14 +87,26 @@ func configureRKE2(ctx *image.Context) (string, error) {
 		return "", fmt.Errorf("copying RKE2 installer script: %w", err)
 	}
 
-	configFile, err := copyKubernetesConfig(ctx, image.KubernetesDistroRKE2)
+	config, err := parseKubernetesConfig(ctx)
 	if err != nil {
-		return "", fmt.Errorf("copying RKE2 config: %w", err)
+		return "", fmt.Errorf("parsing RKE2 config: %w", err)
+	}
+
+	cni, multusEnabled, err := extractCNI(config)
+	if err != nil {
+		return "", fmt.Errorf("extracting CNI from RKE2 config: %w", err)
+	}
+
+	configFile, err := storeKubernetesConfig(ctx, config, image.KubernetesDistroRKE2)
+	if err != nil {
+		return "", fmt.Errorf("storing RKE2 config file: %w", err)
 	}
 
 	installPath, imagesPath, err := ctx.KubernetesArtefactDownloader.DownloadArtefacts(
-		ctx.ImageDefinition.Kubernetes,
 		ctx.ImageDefinition.Image.Arch,
+		ctx.ImageDefinition.Kubernetes.Version,
+		cni,
+		multusEnabled,
 		ctx.CombustionDir,
 	)
 	if err != nil {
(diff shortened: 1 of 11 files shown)


pkg/combustion/kubernetes.go
```diff
@@ -122,28 +138,116 @@ func configureRKE2(ctx *image.Context) (string, error) {
...
+		}
+
+		return parseCNIs(cnis)
+
+	default:
+		return "", false, fmt.Errorf("invalid cni: %v", configuredCNI)
+	}
+}
+
+func parseCNIs(cnis []string) (cni string, multusEnabled bool, err error) {
```
atanasdinov, line 216: The validations here basically replicate the [RKE2 behaviour](https://github.com/rancher/rke2/blob/master/pkg/cli/cmds/server.go#L169). We discussed this with @rdoxenham and we'd want to establish the same rules that the particular distribution enforces and not silently fix mistakes on behalf of the user.

I believe we should be able to extract this out to the new validation framework eventually.

pkg/combustion/kubernetes.go
```diff
@@ -122,28 +138,116 @@ func configureRKE2(ctx *image.Context) (string, error) {
...
+		}
+	default:
+		return "", false, fmt.Errorf("invalid cni value: %v", cnis)
+	}
+
+	return cni, multusEnabled, nil
+}
+
+func storeKubernetesConfig(ctx *image.Context, config map[string]any, distribution string) (string, error) {
+	data, err := yaml.Marshal(config)
```
jdob, line 240: What made you change the user-specified config by explicitly entering the default? I could see an argument for doing this, just curious what the rationale was. Feels a bit wrong to modify their files, so I at least want to be able to say we've had the discussion on why it's needed.

pkg/combustion/templates/15-rke2-installer.sh.tpl
```diff
@@ -6,10 +6,8 @@ mkdir -p /var/lib/rancher/rke2/agent/images/
 cp {{ .ImagesPath }}/* /var/lib/rancher/rke2/agent/images/
 umount /var
 
-{{- if .ConfigFile }}
```
jdob, line 9: I imagine most users will want to do _some_ sort of customization to their RKE2 instance and not be comfortable with the install out of box, but why is it now a requirement?

jdob (approved):
Approved as long as my questions didn't make you think "Wait, maybe we need to change this."

pkg/combustion/templates/15-rke2-installer.sh.tpl
atanasdinov: It is a requirement because even if a file is not provided, we store one with a default value. Currently it is only `cni: cilium` in these cases but it can easily grow.

pkg/combustion/kubernetes.go
```diff
@@ -122,28 +138,116 @@ func configureRKE2(ctx *image.Context) (string, error) {
...
 
 	configDir := generateComponentPath(ctx, k8sConfigDir)
 	configFile := filepath.Join(configDir, k8sConfigFile)
 
-	_, err := os.Stat(configFile)
+	b, err := os.ReadFile(configFile)
 	if err != nil {
 		if errors.Is(err, fs.ErrNotExist) {
-			return "", fmt.Errorf("kubernetes component directory exists but does not contain config.yaml")
+			return nil, fmt.Errorf("kubernetes component directory exists but does not contain config.yaml")
```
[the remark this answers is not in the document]
atanasdinov: You're absolutely correct. This is most likely going to change even in the next iteration where we'd have to better organise multiple configuration files (for HA architectures). Such check will eventually be plugged into the validation framework if it's still necessary.

pkg/combustion/kubernetes.go
```diff
@@ -122,28 +138,116 @@ func configureRKE2(ctx *image.Context) (string, error) {
 	return rke2InstallScript, nil
 }
 
-func copyKubernetesConfig(ctx *image.Context, distro string) (string, error) {
+func parseKubernetesConfig(ctx *image.Context) (map[string]any, error) {
+	auditDefaultCNI := func() {
+		auditMessage := fmt.Sprintf("Kubernetes CNI not explicitly set, defaulting to: %s", cniDefaultValue)
+		log.Audit(auditMessage)
```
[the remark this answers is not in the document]
atanasdinov: This is one of those which felt "critical" enough to show. We no longer will fail if a CNI is not provided so it felt important to show it to the user. Maybe it's just me though - I'm fine with removing it.

pkg/combustion/kubernetes.go
atanasdinov: If a user does not provide a configuration file or provides one without a `cni` configuration,  we have to modify the configuration before saving it. The default one in those cases would be Canal, however, @rdoxenham suggested that for Edge deployments we should always go for Cilium if something else is not explicitly set.

pkg/combustion/kubernetes.go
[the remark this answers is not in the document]
rdoxenham: I agree that this is probably a critical one to show... CNI selection is something that customers will usually have a strong preference about, and if they've failed to have it in their configuration then I think it makes sense to call them out on it.

pkg/combustion/kubernetes.go
rdoxenham: Whilst we're trying to be flexible, we should also consider being quite prescriptive with our desired configuration that customers are using. Cilium is a strong choice here and if we can standardise on this as the default, I think that makes sense.

rdoxenham (approved):
LGTM. Thank you!

pkg/combustion/kubernetes.go
jdob: Works for me 👍🏼

pkg/combustion/kubernetes.go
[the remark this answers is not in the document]
jdob: Ok, I see the rationale. How about if we move it to the validation step (can be in the follow up related to the validation comment below) and show it then? I'm envisioning some sort of specific callout that shows "We're running a validation step, here are the results" where we mention both things that straight up fail and warnings like this.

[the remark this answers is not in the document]
atanasdinov: Completely agree with this. We can move it once we move the component validation to where it is more appropriate.

Issue #61, opened 2023-12-12
Verify network configuration on different Combustion stages


atanasdinov:
Elemental support in https://github.com/suse-edge/edge-image-builder/pull/53 introduced the `# combustion: network` marker to the base combustion script.

Verify whether nmc is still working as expected at that stage and whether we can use the [prepare](https://github.com/openSUSE/combustion#perform-modifications-in-the-initrd-environment) marker to plug it in earlier instead.

PR #115, opened 2024-01-18
Resolve network configuration issues

atanasdinov:
- Utilizes Combustion's prepare marker for static configurations
- Sets a high execution priority for the network script
- Defaults to DHCP in cases where nmc fails

Closes:
- https://github.com/suse-edge/edge-image-builder/issues/61
- https://github.com/suse-edge/nm-configurator/issues/20


pkg/combustion/network.go
```diff
@@ -19,10 +19,10 @@ const (
 	// Used for both input component source and
 	// output configurations subdirectory under combustion.
 	networkConfigDir        = "network"
-	networkConfigScriptName = "configure-network.sh"
+	networkConfigScriptName = "03-configure-network.sh"
```
atanasdinov, line 22: This will now always be the first script we execute. I intentionally didn't go for 01 or 02 in case we'd want to plug in something more important before this but I'm open to changing it as well.

rdoxenham: Good to leave 01/02 free... 03 is good as we really need to think about bringing up the network really early in the process given the other network-related items that are configured.

rdoxenham (approved):
LGTM. Have tested successfully 🚀 thanks!

pkg/combustion/network.go
jdob: Agreed that we want to leave some way to inject things at the very start. I still think we need a more comprehensive numbering review once we get close to MVP, though I don't see this one moving from the start.

pkg/combustion/templates/script-base.sh.tpl
```diff
@@ -1,7 +1,17 @@
 #!/bin/bash
-# combustion: network
 set -euo pipefail
 
+{{ if .NetworkScript -}}
```
jdob, line 4: Surprised we made it this long without needing to templatize the base combustion script.

pkg/combustion/templates/script-base.sh.tpl
```diff
@@ -1,7 +1,17 @@
...
+# combustion: prepare network
+
+if [ "${1-}" = "--prepare" ]; then
+    ./{{ .NetworkScript }}
+    exit 0
+fi
+{{- else -}}
+# combustion: network
```
jdob, line 12: This is so funny to me since it looks like an else block that simply lays down a comment, but the implications are enormous.

jdob (approved):

PR #117, opened 2024-01-18
Migrated artifact registry into new validation framework

jdob:
- Migrated the isEmbeddedArtifactRegistry check into the combustion file for the registry work since it's primarily used by that component


pkg/image/validation/registry.go
```diff
@@ -0,0 +1,97 @@
...
+
+	"github.com/suse-edge/edge-image-builder/pkg/combustion"
+	"github.com/suse-edge/edge-image-builder/pkg/image"
+)
+
+const (
+	registryComponent = "Artifact Registry"
+)
+
+func validateEmbeddedArtefactRegistry(ctx *image.Context) []FailedValidation {
```
dbw7, line 15: This is a nit but `Artifact` is only spelled with an "e" once in the whole file.

pkg/image/validation/registry_test.go
```diff
@@ -0,0 +1,238 @@
+package validation
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/suse-edge/edge-image-builder/pkg/image"
+)
+
+func TestValidateEmbeddedArtefactRegistry(t *testing.T) {
```
dbw7, line 10: Same comment as before, shows up here and on line 65, but only for `2/17` instances.

pkg/image/validation/registry_test.go
```diff
@@ -0,0 +1,238 @@
...
+						Version: "31",
+					},
+				},
+			},
+			ExpectedFailedMessages: []string{
+				"The 'name' field is required for each entry in 'images'.",
+				"The 'name' field is required for each entry in 'charts'.",
+			},
+		},
+	}
```
dbw7, line 55: Another potential nit, I think this coverage is fine, but I think I'd also like to see individual failure cases.

pkg/image/validation/registry_test.go
```diff
@@ -0,0 +1,238 @@
...
+						RepoURL: "example.com",
+						Version: "1.0",
+					},
+				},
+			},
+			ExpectedFailedMessages: []string{
+				"The 'repoURL' field must begin with either 'http://' or 'https://'.",
+			},
+		},
+	}
```
dbw7, line 219: Same nit suggestion regarding individual failure cases (which is probably not necessary to be honest). Though I will say that I really like the way you concatenate the error messages, very cool.

dbw7 (commented):
This is the first time I’ve gotten to see the new validation changes, I really like it! None of the suggestions are urgent, very minor nits.

pkg/image/validation/registry.go
```diff
@@ -0,0 +1,97 @@
...
+	var failures []FailedValidation
+	def := ctx.ImageDefinition
+
+	if combustion.IsEmbeddedArtifactRegistryEmpty(def.EmbeddedArtifactRegistry) {
+		return failures
+	}
```
atanasdinov, line 21: Can we drop this check? Relying on the combustion package for the validation itself seems weird and trying to iterate over nil slices should not lead to errors below.

atanasdinov (approved):

pkg/image/validation/registry.go
jdob: I know, I was trying to keep consistent with the other places in the code that use Artefact, forgetting it was part of the definition. I'll revert these for consistency within the file.

pkg/image/validation/registry_test.go
jdob: The individual failure cases are covered in the respective tests for validating container images and helm charts. This test simply shows that both of those are called from this runner function.

pkg/image/validation/registry.go
jdob: I didn't love this dependency either; it was mostly because the check is used at both runtime and validation time. But since this is the only instance where that happens (so far) and like you said it doesn't affect the behavior, I agree with dropping it.

PR #122, opened 2024-01-19
Migrated the CLI to use the new validation framework

jdob:
Previous output, which only displayed the first validation failure encountered:
```
2024/01/19 19:53:41 CLI arguments could not be parsed: parsing image definition file iso-image.yaml: error validating definition file: error validating image: arch not defined
```

New output:
```
  Image
    The 'imageType' field must be one of: iso, raw
    The 'arch' field is required in the 'image' section.
    The 'outputImageName' field is required in the 'image' section.
    The specified base image 'oops.iso' cannot be found.
  Operating System
    Duplicate kernel argument found: foo
    Systemd enable list contains duplicate entries: bar
    Systemd conflict found, 'foo' is both enabled and disabled.
    User 'jdob' must have either a password or SSH key.
    The suma 'host' field may not contain 'http://' or 'https://'
    The 'activationKey' field is required for the 'suma' section.
    The 'unattended' field can only be used when 'imageType' is 'iso'.
    The 'installDevice' field can only be used when 'imageType' is 'iso'.
  Artifact Registry
    Duplicate image name 'hello-world:latest' found in the 'images' section.
    The 'repoURL' field is required for each entry in 'charts'.
    The 'version' field is required for each entry in 'charts'.
    The 'repoURL' field is required for each entry in 'charts'.
    Duplicate chart name 'My Charts' found in the 'charts' section.
    The 'repoURL' field is required for each entry in 'charts'.
    The 'version' field is required for each entry in 'charts'.
  Kubernetes
    The 'apiVIP' field is required in the 'network' section when defining entries under 'nodes'.
    The 'apiHost' field is required in the 'network' section when defining entries under 'nodes'.
    The 'hostname' field is required for entries in the 'nodes' section.
```

Notes:
* The previous validation is not deleted as part of this PR. Once this lands, I'll submit a second one to remove the old code.
* For this PR, please focus on the formatting of the messages (inclusion of sections instead of a single list, whether or not I should also include the section's YAML key, indentation, etc.). If there are concerns with the phrasing of any of the individual validation messages, please make sure to mark them as such. I'll create a follow up Issue & PR that addresses the usefulness of the message contents separately (I don't want this PR to get bogged down debating the phrasing of the messages themselves, just focusing on the formatting specifically).
  * To that point, I'm debating if I want to include more specifics (where possible) about the actual entry that caused the issue. That might get ugly.
  * I'm also debating removing repeated messages (see the 'repoUrl' warnings under the registry). Those current appear once per failure; in this case, I defined multiple charts that missed the field to trigger such a situation.

### DIFF
--- a/cmd/eib/main.go
+++ b/cmd/eib/main.go
@@ -6,6 +6,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 
 	"github.com/suse-edge/edge-image-builder/pkg/build"
@@ -77,14 +78,27 @@ func processArgs() (*image.Context, error) {
 
 		logMessageBuilder := strings.Builder{}
 
-		for componentName, failures := range failedValidations {
+		orderedComponentNames := make([]string, 0, len(failedValidations))
+		for c := range failedValidations{
+			orderedComponentNames = append(orderedComponentNames, c)
+		}
+		sort.Sort(sort.StringSlice(orderedComponentNames))
+
+		for _, componentName := range orderedComponentNames {
+			failures := failedValidations[componentName]
 			audit.Audit(fmt.Sprintf("  %s", componentName))
 			for _, cf := range failures {
 				audit.Audit(fmt.Sprintf("    %s", cf.UserMessage))
+				logMessageBuilder.WriteString(cf.UserMessage + "\n")
+				if cf.Error != nil {
+					logMessageBuilder.WriteString("\t" + cf.Error.Error() + "\n")
+				}
 			}
 		}
 
-		zap.S().Fatalf(logMessageBuilder.String())
+		if s := logMessageBuilder.String(); s != "" {
+			zap.S().Fatalf("Image definition validation failures:\n%s", s)
+		}
 	}
 
 	if validate {
--- a/cmd/eib/main.go
+++ b/cmd/eib/main.go
@@ -75,19 +75,9 @@ func processArgs() (*image.Context, error) {
 	if len(failedValidations) > 0 {
 		audit.Audit("Image definition validation found the following errors:")
 
-		failuresByComponent := map[string][]validation.FailedValidation{}
 		logMessageBuilder := strings.Builder{}
 
-		for _, fv := range failedValidations {
-			failuresByComponent[fv.Component] = append(failuresByComponent[fv.Component], fv)
-
-			logMessageBuilder.WriteString(fv.UserMessage)
-			if fv.Error != nil {
-				logMessageBuilder.WriteString(fv.Error.Error())
-			}
-		}
-
-		for componentName, failures := range failuresByComponent {
+		for componentName, failures := range failedValidations {
 			audit.Audit(fmt.Sprintf("  %s", componentName))
 			for _, cf := range failures {
 				audit.Audit(fmt.Sprintf("    %s", cf.UserMessage))
--- a/cmd/eib/main.go
+++ b/cmd/eib/main.go
@@ -6,10 +6,12 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/suse-edge/edge-image-builder/pkg/build"
 	"github.com/suse-edge/edge-image-builder/pkg/combustion"
 	"github.com/suse-edge/edge-image-builder/pkg/image"
+	"github.com/suse-edge/edge-image-builder/pkg/image/validation"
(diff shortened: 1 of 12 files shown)


cmd/eib/main.go
```diff
@@ -66,6 +71,37 @@ func processArgs() (*image.Context, error) {
...
 
+	failedValidations := validation.ValidateDefinition(ctx)
+	if len(failedValidations) > 0 {
+		audit.Audit("Image definition validation found the following errors:")
+
+		failuresByComponent := map[string][]validation.FailedValidation{}
+		logMessageBuilder := strings.Builder{}
+
+		for _, fv := range failedValidations {
+			failuresByComponent[fv.Component] = append(failuresByComponent[fv.Component], fv)
```
atanasdinov, line 82: The repetitive nature of this makes me believe that `ValidateDefinition()` returning a `map[string][]validation.FailedValidation` is the way to go here. We'd be dropping the component initialisation for each struct and we'd be completely dropping the loop here in favour of the one below.

jdob: Yeah, I agree. This sort of translation like this is a very Python thing to do so I hadn't even stepped back to consider that `ValidateDefinition` should return them by component.

jdob:
> On the message repetition topic: it does make sense to only return the message once if it's a generic one not applicable to a specific field e.g. `The 'repoURL' field is required for each entry in 'charts'.`
> 
> The question in this case is then probably "Can we always tie a failure to a specific entry?" for which I'm not able to provide a good enough answer.
> 
> Happy to approve once the comments below are addressed.

Yeah, the more I thought about it over the weekend, the more I realized we don't have a clean way of mapping the error to its "entity", for lack of a better word. For instance, if a package name is missing, it's difficult to figure out how to say "The package name is missing for the, uh, unnamed package."

I'm gonna leave the repetition as-is and we can see how things shake out during QA.

jdob:
New log output (not the user-facing messages) as of my next commit:
```╭─[~/Code/eib]─[]─[]─●
╰─➤ cat _build/build-Jan22_16-36-51/eib-build.log                                                                                        1 ↵
2024-01-22T16:36:51.738Z	FATAL	eib/main.go:92	Image definition validation failures:
The 'imageType' field must be one of: iso, raw
The 'arch' field is required in the 'image' section.
The 'outputImageName' field is required in the 'image' section.
The specified base image 'oops.iso' cannot be found.
	stat /eib/images/oops.iso: no such file or directory
Duplicate kernel argument found: foo
Systemd enable list contains duplicate entries: bar
Systemd conflict found, 'foo' is both enabled and disabled.
User 'jdob' must have either a password or SSH key.
The suma 'host' field may not contain 'http://' or 'https://'
The 'activationKey' field is required for the 'suma' section.
The 'unattended' field can only be used when 'imageType' is 'iso'.
The 'installDevice' field can only be used when 'imageType' is 'iso'.
Duplicate image name 'hello-world:latest' found in the 'images' section.
The 'repoURL' field is required for each entry in 'charts'.
The 'version' field is required for each entry in 'charts'.
The 'repoURL' field is required for each entry in 'charts'.
Duplicate chart name 'My Charts' found in the 'charts' section.
The 'repoURL' field is required for each entry in 'charts'.
The 'version' field is required for each entry in 'charts'.
The 'apiVIP' field is required in the 'network' section when defining entries under 'nodes'.
The 'apiHost' field is required in the 'network' section when defining entries under 'nodes'.
The 'hostname' field is required for entries in the 'nodes' section.

main.processArgs
	/src/cmd/eib/main.go:92
main.main
	/src/cmd/eib/main.go:159
runtime.main
```

pkg/image/validation/validation.go
```diff
@@ -23,7 +23,10 @@ func ValidateDefinition(ctx *image.Context) []FailedValidation {
 	}
 	for _, v := range validations {
 		componentFailures := v(ctx)
-		failures = append(failures, componentFailures...)
+
+		if len(componentFailures) > 0 {
+			failures[componentFailures[0].Component] = componentFailures
```
atanasdinov, line 28: This is not what I imagined. I thought that `validations` would map a section / component to its validation function and use that key in order to store the failures. This way you don't have to set `Component` for each object and missing it will not lead into strange behaviour.

jdob: I don't think it makes sense for each `validate*` method to return a map though. Each will only ever validate things in one component, so it feels weird to return a map with a single key from each of those methods.

atanasdinov (approved):

PR #123, opened 2024-01-19
Manifest Handling

dbw7:
Putting this up as a draft so that multiple eyes can take a look at it. At the moment the main handling is untested and, to be honest, there are many questions about how this implementation should actually work. Admittedly, this code is very messy but it outlines the general manifest handling flow.


pkg/manifests/manifests.go
```diff
@@ -0,0 +1,193 @@
...
+	"github.com/suse-edge/edge-image-builder/pkg/image"
+	"go.uber.org/zap"
+	"gopkg.in/yaml.v3"
+)
+
+func configureManifests(ctx *image.Context) error {
+	var downloadedManifests []string
+	var combinedManifestPaths []string
+	var extractedImages []string
+	var err error // created here to avoid scoping issues with "downloadedManifests" when using := on line 29
```
atanasdinov, line 21: There's no need to have the comment and comments listing a particular (or multiple) LoC are not generally a good idea since the implementation can easily change but the comment can remain untouched.

pkg/manifests/manifests.go
```diff
@@ -0,0 +1,193 @@
...
+	"strings"
+
+	"github.com/suse-edge/edge-image-builder/pkg/fileio"
+	"github.com/suse-edge/edge-image-builder/pkg/http"
+	"github.com/suse-edge/edge-image-builder/pkg/image"
+	"go.uber.org/zap"
+	"gopkg.in/yaml.v3"
+)
+
+func configureManifests(ctx *image.Context) error {
```
atanasdinov, line 17: I'd argue this should be part of the Kubernetes combustion component (and package). What is the purpose behind extracting it separately?

pkg/manifests/manifests.go
```diff
@@ -0,0 +1,193 @@
...
+			return fmt.Errorf("creating %s dir: %w", downloadDestination, err)
+		}
+
+		downloadedManifests, err = downloadManifests(ctx, downloadDestination)
+		if err != nil {
+			return fmt.Errorf("error downloading manifests: %w", err)
+		}
+	}
+
+	localManifestSrcDir := filepath.Join(ctx.ImageConfigDir, "manifests")
```
atanasdinov, line 35: This should instead be under the kubernetes directory:
```
kubernetes/
├── config
│   ├── agent.yaml
│   └── server.yaml
└── manifests
    └── manifest1.yaml
```

pkg/manifests/manifests.go
```diff
@@ -0,0 +1,193 @@
...
+)
+
+func configureManifests(ctx *image.Context) error {
+	var downloadedManifests []string
+	var combinedManifestPaths []string
+	var extractedImages []string
+	var err error // created here to avoid scoping issues with "downloadedManifests" when using := on line 29
+
+	if len(ctx.ImageDefinition.Kubernetes.Manifests.URLs) != 0 {
+		downloadDestination := filepath.Join(ctx.CombustionDir, "downloaded-manifests")
```
atanasdinov, line 24: Just FYI, if I have the time I'd like to introduce a better structure of the Combustion directory where everything related to Kubernetes falls under `/combustion/kubernetes`. Until then, this is fine.

pkg/manifests/manifests.go
```diff
@@ -0,0 +1,193 @@
...
+
+	return manifestPaths, nil
+}
+
+func downloadManifests(ctx *image.Context, destPath string) ([]string, error) {
+	manifests := ctx.ImageDefinition.Kubernetes.Manifests.URLs
+	var manifestPaths []string
+
+	for index, manifestURL := range manifests {
+		filePath := filepath.Join(destPath, fmt.Sprintf("manifest-%d.yaml", index+1))
```
atanasdinov, line 150: Is this naming the final decision we set for?

pkg/manifests/manifests.go
```diff
@@ -0,0 +1,193 @@
...
+		}
+
+		foundImages, err := findImagesInManifest(manifestData)
+		if err != nil {
+			return fmt.Errorf("error finding images in manifest %w", err)
+		}
+		extractedImages = append(extractedImages, foundImages...)
+	}
+
+	extractedImages = removeDuplicateImages(extractedImages)
```
atanasdinov, line 57: Seems obsolete if `addImagesToRegistryDefinition` already has a check for it. `findImagesInManifest` also lists the images in a map first and then goes through a slice and then through a map again. Can we reuse a data structure?

pkg/manifests/manifests.go
```diff
@@ -0,0 +1,193 @@
...
+	return formattedImages
+}
+
+func addImagesToRegistryDefinition(ctx *image.Context, manifestImages []string) {
+	for _, imageName := range manifestImages {
+		if !isInRegistry(ctx, imageName) {
+			containerImageDef := image.ContainerImage{
+				Name: imageName,
+			}
+			ctx.ImageDefinition.EmbeddedArtifactRegistry.ContainerImages = append(ctx.ImageDefinition.EmbeddedArtifactRegistry.ContainerImages, containerImageDef)
```
atanasdinov, line 180: This is the part which is the biggest problem, in my opinion. I don't think we should be modifying what the image definition itself contains - it's much better to have a slice which contains all of the `EmbeddedArtifactRegistry.ContainerImages` and then append the downloaded ones to that instead.

pkg/manifests/manifests.go
dbw7: No, that's up for discussion, just putting that temporarily.

pkg/manifests/manifests.go
dbw7: 👍

pkg/manifests/manifests.go
dbw7: This has now been changed into a package that will have uses in both the registry and the kubernetes components.

pkg/manifests/manifests.go
dbw7: 👍

pkg/manifests/manifests.go
dbw7: I've restructured it to be a much cleaner flow

dbw7:
This is still missing a lot of tests for the manifests functions, however, using a manifest such as:
```
apiVersion: 1.0
image:
  imageType: ISO
  baseImage: slemicro.iso
  outputImageName: eibimage.iso
  arch: x86_64
operatingSystem:
  users:
    - username: root
      encryptedPassword: $6$harL4dKSOzhzprzV$Kquzh7WwOSQhivefg/KVC2XRyCpxG0gOoj1c97ySOPolfRWmvtY5fTQFWrTxjuEmUKYinB7A1HAH1cf3oYjjv0
embeddedArtifactRegistry:
  images:
    - name: quay.io/podman/hello
kubernetes:
  version: v1.29.0+rke2r1
  manifests:
    urls:
      - https://k8s.io/examples/application/nginx-app.yaml
  charts:
    - name: rancher
      repoURL: https://releases.rancher.com/server-charts/latest
      version: 2.8.0
```
results in a valid image that has all the necessary images and helm charts in the internal registry.
Additionally, all that is missing is clean-up + copying the manifests to combustion and then actually deploying the manifests.

pkg/combustion/templates/13-embedded-registry.sh.tpl
```diff
@@ -17,12 +17,13 @@ cat <<- EOF > /etc/systemd/system/eib-embedded-registry.service
   Type=simple
   User=root
   WorkingDirectory=/opt/hauler
-  ExecStartPre=/usr/local/bin/hauler store load {{ .EmbeddedRegistryTar }}
-  ExecStart=/usr/local/bin/hauler store serve
+  ExecStartPre=/bin/bash -c 'for file in /opt/hauler/*.tar.zst; do /usr/local/bin/hauler store load \$file; done'
```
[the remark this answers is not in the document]
dbw7: We did but this same implementation works for both, though I suppose it might be a bit confusing to keep it like that, will switch it back.

pkg/manifests/manifests.go
```diff
@@ -0,0 +1,190 @@
...
+	"go.uber.org/zap"
+	"gopkg.in/yaml.v3"
+)
+
+func GetAllImagesAndCharts(ctx *image.Context) ([]image.ContainerImage, []image.HelmChart, error) {
+	var downloadedManifestPaths []string
+	var combinedManifestPaths []string
+	var extractedImagesSet = make(map[string]string)
+	var err error
+	helmCharts := ctx.ImageDefinition.Kubernetes.HelmCharts
```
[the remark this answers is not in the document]
dbw7: Yeah I think I was trying to make it more explicit, but it is redundant.

pkg/manifests/manifests.go
```diff
@@ -0,0 +1,190 @@
+package manifests
```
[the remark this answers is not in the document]
dbw7: My thinking is having it as a package purely for manifest handling that is utilized in multiple places (both registry and k8s) but I'm not particularly attached to that.

pkg/combustion/registry.go
```diff
@@ -10,6 +10,7 @@ import (
 	"github.com/suse-edge/edge-image-builder/pkg/fileio"
 	"github.com/suse-edge/edge-image-builder/pkg/image"
 	"github.com/suse-edge/edge-image-builder/pkg/log"
+	registry "github.com/suse-edge/edge-image-builder/pkg/manifests"
```
atanasdinov, line 13: Package name and directories must be the same. This is why you're required to use an alias here and you wouldn't be.

pkg/combustion/registry.go
```diff
@@ -30,12 +33,25 @@ var haulerManifest string
...
+	err := os.Mkdir(registriesDir, os.ModePerm)
+	if err != nil {
+		log.AuditComponentFailed(registryComponentName)
+		return nil, fmt.Errorf("creating registry dir: %w", err)
+	}
+
+	containerImages, err := registry.GetAllImages(ctx)
+	if err != nil {
+		log.AuditComponentFailed(registryComponentName)
+		return nil, fmt.Errorf("getting all container images and helm charts: %w", err)
```
atanasdinov, line 51:
```suggestion
		return nil, fmt.Errorf("getting all container images: %w", err)
```

pkg/combustion/templates/13-embedded-registry.sh.tpl
```diff
@@ -4,7 +4,7 @@ set -euo pipefail
 mkdir /opt/hauler
 
 mount /usr/local
-mv {{ .EmbeddedRegistryTar }} /opt/hauler/{{ .EmbeddedRegistryTar }}
+mv ./{{ .RegistryDir }}/{{ .EmbeddedRegistryTar }} /opt/hauler/
```
atanasdinov, line 7: Nit: Do we need the `./` part?

pkg/combustion/templates/13-embedded-registry.sh.tpl
```diff
@@ -18,11 +18,12 @@ cat <<- EOF > /etc/systemd/system/eib-embedded-registry.service
...
   ExecStartPre=/usr/local/bin/hauler store load {{ .EmbeddedRegistryTar }}
-  ExecStart=/usr/local/bin/hauler store serve
+  ExecStart=/usr/local/bin/hauler store serve -p {{ .Port }}
   Restart=on-failure
 
   [Install]
   WantedBy=multi-user.target
 EOF
 
+
```
atanasdinov, line 28: Is this line intended?

pkg/manifests/manifests_test.go
```diff
@@ -0,0 +1,135 @@
...
+
+	// Test
+	storeManifestImageNames(manifestData, extractedImagesSet)
+	allImages := make([]string, 0, len(extractedImagesSet))
+	for uniqueImage := range extractedImagesSet {
+		allImages = append(allImages, uniqueImage)
+	}
+
+	// Verify
+	assert.Equal(t, []string{}, allImages)
```
atanasdinov, line 99: We can assert an empty map here.

pkg/manifests/manifests.go
```diff
@@ -0,0 +1,150 @@
...
+		}
+
+		sourcePath := filepath.Join(src, manifest.Name())
+		manifestPaths = append(manifestPaths, sourcePath)
+	}
+
+	return manifestPaths, nil
+}
+
+func downloadManifests(ctx *image.Context, destPath string) ([]string, error) {
```
atanasdinov, line 136: This function doesn't require a full context. If it accepts just the slice of manifests, then all the `setupContext()` calls will not be necessary in the tests.

pkg/manifests/manifests.go
```diff
@@ -0,0 +1,150 @@
...
+	err = yaml.Unmarshal(manifestData, &manifest)
+	if err != nil {
+		return nil, fmt.Errorf("error unmarshalling manifest yaml '%s': %w", manifestPath, err)
+	}
+
+	return manifest, nil
+}
+
+func storeManifestImageNames(data any, imageSet map[string]string) {
+
```
atanasdinov, line 87: Functions shouldn't start with an empty line.

pkg/manifests/manifests.go
```diff
@@ -0,0 +1,150 @@
...
+		downloadedManifestPaths, err := downloadManifests(ctx, downloadDestination)
+		if err != nil {
+			return nil, fmt.Errorf("error downloading manifests: %w", err)
+		}
+
+		combinedManifestPaths = append(combinedManifestPaths, downloadedManifestPaths...)
+	}
+
+	localManifestSrcDir := filepath.Join(ctx.ImageConfigDir, "kubernetes", "manifests")
+	localManifestPaths, err := getLocalManifestPaths(localManifestSrcDir)
```
atanasdinov, line 35: Does this even work if the directory doesn't exist or is empty?

pkg/manifests/manifest_integration_test.go
```diff
@@ -0,0 +1,216 @@
...
+	ctx.ImageDefinition.Kubernetes.Manifests.URLs = []string{
+		"https://k8s.io/examples/application/nginx-app.yaml",
+	}
+
+	// Test
+	manifestPaths, err := downloadManifests(ctx, destPath)
+
+	// Verify
+	require.NoError(t, err)
+	assert.FileExists(t, expectedFilePath)
```
atanasdinov, line 60: Nice!

pkg/manifests/manifest_integration_test.go
```diff
@@ -0,0 +1,216 @@
...
+		},
+		{
+			Name: "custom-api:1.2.3",
+		},
+		{
+			Name:           "quay.io/podman/hello",
+			SupplyChainKey: "sample-key",
+		},
+	}
+	sort.Slice(expectedContainerImages, func(i, j int) bool {
```
atanasdinov, line 129: Can't we manually put them in order? This only adds a layer of complexity (although small).

pkg/manifests/manifest_integration_test.go
```diff
@@ -0,0 +1,216 @@
...
+
+	ctx.ImageDefinition.Kubernetes.Manifests.URLs = []string{
+		"k8s.io/examples/application/nginx-app.yaml",
+	}
+
+	// Test
+	_, err := GetAllImages(ctx)
+
+	// Verify
+	require.ErrorContains(t, err, "error downloading manifests: downloading manifest")
```
atanasdinov, line 203: Can we include more of the error? It's not quite clear why it fails aside from the name of the test itself.

pkg/manifests/manifest_integration_test.go
```diff
@@ -0,0 +1,216 @@
+//go:build integration
```
atanasdinov, line 1: This file must only contain tests which perform actual downloading from the internet. All other cases belong to `manifests_test.go` instead. Also please either rename this file to plural "manifests" or the others to singular "manifest" for consistency.

atanasdinov (changes requested):

pkg/combustion/templates/13-embedded-registry.sh.tpl
dbw7: Will test, probably not

pkg/combustion/templates/13-embedded-registry.sh.tpl
dbw7: No that was a mistake, thank you!

pkg/manifests/manifests.go
dbw7: Updated this based on our conversation.

pkg/manifests/manifest_integration_test.go
dbw7: Just tested, unfortunately the order comes back different every time (I assume cause it's a map at first)

pkg/manifests/manifest_integration_test.go
dbw7: Just tested, unfortunately the order comes back different every time (I assume cause it's a map at first)

pkg/combustion/registry.go
```diff
@@ -173,6 +222,19 @@ func createRegistryCommand(ctx *image.Context, commandName string, args []string
...
 
-func IsEmbeddedArtifactRegistryEmpty(registry image.EmbeddedArtifactRegistry) bool {
-	return len(registry.HelmCharts) == 0 && len(registry.ContainerImages) == 0
+func IsEmbeddedArtifactRegistryConfigured(ctx *image.Context) bool {
+	return len(ctx.ImageDefinition.Kubernetes.HelmCharts) != 0 ||
+		len(ctx.ImageDefinition.EmbeddedArtifactRegistry.ContainerImages) != 0 ||
+		len(ctx.ImageDefinition.Kubernetes.Manifests.URLs) != 0
+}
+
+func localManifestsConfigured(localManifestSrcDir string) (bool, error) {
```
atanasdinov, line 231: Can we reuse `isComponentConfigured` instead?

pkg/registry/manifests_integration_test.go
```diff
@@ -0,0 +1,105 @@
...
+		{
+			Name:           "quay.io/podman/hello",
+			SupplyChainKey: "sample-key",
+		},
+	}
+	manifestURLs := []string{"https://k8s.io/examples/application/nginx-app.yaml"}
+
+	// Test
+	containerImages, err := GetAllImages(embeddedContainerImages, manifestURLs, localManifestSrcDir, manifestDownloadDest)
+	sort.Slice(containerImages, func(i, j int) bool {
```
atanasdinov, line 98: This has to happen after `require.NoError()` below.

pkg/registry/manifests_integration_test.go
```diff
@@ -0,0 +1,105 @@
+package registry
```
atanasdinov, line 1: We're missing the build tag.

pkg/registry/manifests_test.go
```diff
@@ -0,0 +1,241 @@
...
+	defer func() {
+		require.NoError(t, os.RemoveAll("downloaded-manifests"))
+	}()
+
+	// Test
+	manifestPaths, err := getLocalManifestPaths("downloaded-manifests")
+
+	// Verify
+	require.NoError(t, err)
+	assert.Equal(t, []string(nil), manifestPaths)
```
atanasdinov, line 147:
```suggestion
	assert.Nil(t, manifestPaths)
```

pkg/registry/manifests_test.go
```diff
@@ -0,0 +1,241 @@
...
+	// Verify
+	require.ErrorContains(t, err, "manifest download destination directory not defined")
+}
+
+func TestGetAllImagesLocalManifestDirNotDefined(t *testing.T) {
+	// Test
+	_, err := GetAllImages(nil, nil, "", "")
+
+	// Verify
+	require.NoError(t, err)
```
atanasdinov, line 185: Let's assert that the result is an empty slice.

atanasdinov (approved):

pkg/combustion/registry.go
dbw7: Yes, I forgot about that

PR #124, opened 2024-01-22
Implement HA RKE2 architecture

atanasdinov:
- Splits RKE2 configuration based on number of nodes
- Introduces a MetalLB + Endpoint Copier Operator manifest in order to create a VIP address for the cluster
- Establishes sane default values for servers and agents in both single and multinode setup
- Generating a cluster token to be done in a follow up


pkg/image/validation/kubernetes.go
```diff
@@ -48,15 +48,9 @@ func validateNodes(k8s *image.Kubernetes) []FailedValidation {
 		})
 	}
 
-	if k8s.Network.APIHost == "" {
-		failures = append(failures, FailedValidation{
-			UserMessage: "The 'apiHost' field is required in the 'network' section when defining entries under 'nodes'.",
```
atanasdinov, line 53: We decided on having this field as optional.

pkg/combustion/kubernetes.go
```diff
@@ -22,15 +22,25 @@ const (
...
 
 var (
 	//go:embed templates/15-rke2-single-node-installer.sh.tpl
 	rke2SingleNodeInstaller string
+
+	//go:embed templates/15-rke2-multi-node-installer.sh.tpl
+	rke2MultiNodeInstaller string
+
+	//go:embed templates/rke2-vip.yaml.tpl
+	rke2VIPManifest string
```
rdoxenham, line 43: We may want to move this to `k8s-vip.yaml.tpl` and `k8sVIPManifest` respectively so we can adapt to k3s too; k3s will have the same requirements for the API VIP as RKE2.

pkg/combustion/kubernetes.go
```diff
@@ -88,45 +98,146 @@ func configureRKE2(ctx *image.Context) (string, error) {
 		return "", fmt.Errorf("copying RKE2 installer script: %w", err)
 	}
 
-	config, err := parseKubernetesConfig(ctx)
-	if err != nil {
-		return "", fmt.Errorf("parsing RKE2 config: %w", err)
+	if len(ctx.ImageDefinition.Kubernetes.Nodes) > 1 {
+		return configureMultiNodeRKE2(ctx)
```
rdoxenham, line 102: Not something we have to solve for now, but if this was 2 this code path would still execute, whereas it wouldn't be advised to implement two-nodes without NATS integration, for example. No big deal, but maybe a warning would be good about two-nodes with etcd not actually giving you any form of HA (as if one etcd node goes down, the other will too, as you lose quorum).

pkg/combustion/kubernetes.go
```diff
@@ -88,45 +98,146 @@ func configureRKE2(ctx *image.Context) (string, error) {
...
+
+	installPath, imagesPath, err := downloadRKE2Artefacts(ctx, serverConfig)
 	if err != nil {
 		return "", fmt.Errorf("downloading RKE2 artefacts: %w", err)
 	}
 
+	var vipManifest string
+
+	if ctx.ImageDefinition.Kubernetes.Network.APIVIP == "" {
+		zap.S().Info("Virtual IP address for RKE2 cluster is not provided and will not be configured")
```
rdoxenham, line 138:
```suggestion
		zap.S().Info("Virtual IP address for RKE2 cluster is not provided and will not be configured.")
```

pkg/combustion/kubernetes_test.go
```diff
@@ -161,15 +165,15 @@ func TestConfigureKubernetes_SuccessfulRKE2Server(t *testing.T) {
 	require.NoError(t, err)
 
 	contents := string(b)
-	assert.NotContains(t, contents, "export INSTALL_RKE2_TYPE=server",
-		"INSTALL_RKE2_TYPE is set when the definition file does not explicitly set it")
 	assert.Contains(t, contents, "cp server-images/* /var/lib/rancher/rke2/agent/images/")
-	assert.Contains(t, contents, "cp rke2_config.yaml /etc/rancher/rke2/config.yaml")
+	assert.Contains(t, contents, "cp server.yaml /etc/rancher/rke2/config.yaml")
+	assert.Contains(t, contents, "cp rke2-vip.yaml /var/lib/rancher/rke2/server/manifests/rke2-vip.yaml")
```
rdoxenham, line 170: As before, I suspect this manifest will be identical for k3s too, so perhaps it makes sense to rename it, although we can always do that when we land k3s work.

pkg/combustion/kubernetes_test.go
```diff
@@ -184,6 +188,129 @@ func TestConfigureKubernetes_SuccessfulRKE2Server(t *testing.T) {
...
+	b, err = os.ReadFile(configPath)
+	require.NoError(t, err)
+
+	configContents = map[string]any{} // clear the map
+	require.NoError(t, yaml.Unmarshal(b, configContents))
+
+	assert.Equal(t, "canal", configContents["cni"])
+	assert.Equal(t, "123", configContents["token"])
+	assert.Equal(t, nil, configContents["server"])
+	assert.Equal(t, []any{"192.168.122.100.sslip.io", "192.168.122.100", "api.cluster01.hosted.on.edge.suse.com"}, configContents["tls-san"])
```
rdoxenham, line 299: Same here

pkg/combustion/kubernetes_test.go
```diff
@@ -184,6 +188,129 @@ func TestConfigureKubernetes_SuccessfulRKE2Server(t *testing.T) {
...
+	b, err = os.ReadFile(configPath)
+	require.NoError(t, err)
+
+	configContents = map[string]any{} // clear the map
+	require.NoError(t, yaml.Unmarshal(b, configContents))
+
+	assert.Equal(t, "canal", configContents["cni"])
+	assert.Equal(t, "123", configContents["token"])
+	assert.Equal(t, "https://192.168.122.100:9345", configContents["server"])
+	assert.Equal(t, []any{"192.168.122.100.sslip.io", "192.168.122.100", "api.cluster01.hosted.on.edge.suse.com"}, configContents["tls-san"])
```
rdoxenham, line 313: And here

pkg/combustion/templates/15-rke2-multi-node-installer.sh.tpl
```diff
@@ -0,0 +1,49 @@
...
+
+{{- range .Nodes }}
+hosts[{{ .Hostname }}]={{ .Type }}
+{{- end }}
+
+HOSTNAME=$(cat /etc/hostname)
+if [ ! "$HOSTNAME" ]; then
+    echo "ERROR: Could not identify whether the host is an RKE2 server or agent due to missing hostname"
+    exit 1
+fi
```
rdoxenham, line 14: Not sure I've tested this, but isn't `/etc/hostname` by default `localhost.localdomain` ?

rdoxenham (approved):
Some minor comments, especially related to k3s inclusion work, but otherwise this is great. It was tested already 🚀

pkg/combustion/templates/15-rke2-multi-node-installer.sh.tpl
atanasdinov: It will default to `localhost.localdomain` but at combustion time the value is an empty string for some reason. You can verify it by configuring multi node cluster and using a NIC not known by nmc.

pkg/combustion/kubernetes.go
atanasdinov: Punctuation is (usually) not something we use for log file statements.

pkg/combustion/kubernetes.go
atanasdinov: Do we actively discourage an architecture of 1 server and 1 agent? It has nothing to do with a HA as usually defined but it's a topology that users might want to use for some reason. Thoughts?

pkg/combustion/kubernetes.go
rdoxenham: Good call out, you're entirely right. I guess we either have a specific file for rke2, or we modify the template to check if it's rke2/k3s.

pkg/combustion/kubernetes.go
rdoxenham: Oh! I was just thinking of multi-server nodes. You're right to think about one server and one agent use-cases, this is actually something we know that customers would be interested in. Where they need the capacity but not necessarily availability of the control plane. I think the check would be for two *servers*.

pkg/combustion/templates/15-rke2-multi-node-installer.sh.tpl
rdoxenham: No worries. So long as this check works, I think we're good.

pkg/combustion/kubernetes.go
atanasdinov: Makes sense but this is something we should do during the validation step. I'll address it.

pkg/image/definition.go
```diff
@@ -136,9 +136,9 @@ type Network struct {
 }
 
 type Node struct {
-	Hostname string `yaml:"hostname"`
-	Type     string `yaml:"type"`
-	First    bool   `yaml:"firstNode"`
+	Hostname    string `yaml:"hostname"`
+	Type        string `yaml:"type"`
+	Initialiser bool   `yaml:"initialiser"`
```
jdob, line 141: It's not worth blocking the PR for this, but we should discuss the spelling of "initializer" in the image definition (I'm actually being serious this time, not just giving you crap). I think it's potentially confusing that `embeddedArtifactRegistry` uses the Americanized spelling while this uses whatever you call the other spelling version (I know, I know, your answer is going to be "correct"). In the code I don't care if we mix and match, but for a public API I think it looks bad if it's not consistent. We can talk about this in the definition review next week.

pkg/kubernetes/cluster.go
```diff
@@ -0,0 +1,223 @@
...
+	cniKey          = "cni"
+	cniDefaultValue = image.CNITypeCilium
+	serverKey       = "server"
+	tlsSANKey       = "tls-san"
+)
+
+type Cluster struct {
+	// Initialiser is the hostname of the initialiser node.
+	// Defaults to the first configured server if not explicitly selected.
+	Initialiser string
```
jdob, line 31: Nit: In the definition, `Initialiser` is a boolean, but here it's the hostname. There shouldn't be any technical confusion since they are different types, though it might be a bit more clear to new devs if this was `InitialiserHostname` (esp since it matches the pattern of `InitialiserConfig` below).

pkg/kubernetes/cluster.go
```diff
@@ -0,0 +1,223 @@
...
+	zap.S().Infof("Generated cluster token: %s", token)
+	config[tokenKey] = token
+}
+
+func setClusterCNI(config map[string]any) {
+	if _, ok := config[cniKey]; ok {
+		return
+	}
+
+	auditMessage := fmt.Sprintf("Kubernetes CNI not explicitly set, defaulting to: %s", cniDefaultValue)
```
jdob, line 164: Nit: The user-facing messages have been closer to full sentences in many places, so we may want to consider:

"The Kubernetes CNI is not explicitly set, defaulting to '%s'."

jdob (approved):
Nothing blocking this from landing. I do think we should discuss the definition spelling, but that can come next week. I still haven't seen this actually working, but it a.) doesn't actively break anything and b.) looks sound from a coding perspective, so I'm cool with landing this and debugging my issue next week.

PR #125, opened 2024-01-22
Refactor "AdditionalRepos" field to a struct

ipetrov117:
Refactoring `AdditionalRepos` from a `[]string` to a struct was discussed in #64. 
By adopting this structure we get the flexibility to easily add additional repo configurations whenever the need for such configuration arises.

Note: This PR is a prerequisite for the RPM resolver GPG validation logic PR that is currently being implemented.

### DIFF
--- a/pkg/image/validation/os.go
+++ b/pkg/image/validation/os.go
@@ -166,12 +166,27 @@ func validatePackages(os *image.OperatingSystem) []FailedValidation {
 		})
 	}
 
-	if duplicates := findDuplicates(os.Packages.AdditionalRepos); len(duplicates) > 0 {
-		duplicateValues := strings.Join(duplicates, ", ")
-		msg := fmt.Sprintf("The 'additionalRepos' field contains duplicate repos: %s", duplicateValues)
-		failures = append(failures, FailedValidation{
-			UserMessage: msg,
-		})
+	if len(os.Packages.AdditionalRepos) > 0 {
+		urlSlice := []string{}
+
+		for _, repo := range os.Packages.AdditionalRepos {
+			if repo.URL == "" {
+				msg := "Additional repository list contains an entry with empty 'url' field."
+				failures = append(failures, FailedValidation{
+					UserMessage: msg,
+				})
+			}
+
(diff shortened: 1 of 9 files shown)


pkg/rpm/resolver/resolver.go
```diff
@@ -158,6 +158,14 @@ func (r *Resolver) writeDockerfile(localPackagesPath string, packages *image.Pac
 	return nil
 }
 
+func (r *Resolver) generateAddRepoList(repos []image.AddRepo) (repoList []string) {
```
atanasdinov, line 161: Does it make sense to return the string which is being used instead?

pkg/image/validation/os.go
```diff
@@ -178,13 +178,29 @@ func validatePackages(os *image.OperatingSystem) []FailedValidation {
...
-		failures = append(failures, FailedValidation{
-			UserMessage: msg,
-			Component:   osComponent,
-		})
+	if len(os.Packages.AdditionalRepos) > 0 {
+		urlSlice := []string{}
+
+		for _, repo := range os.Packages.AdditionalRepos {
+			if repo.URL == "" {
+				msg := "additional repository list contains an entry with empty 'url' field"
```
atanasdinov, line 186: User messages are capitalised and usually end with a dot.

atanasdinov (approved):

rdoxenham (approved):
LGTM. I assume that the other repo types will be file or OCI?

ipetrov117:
> LGTM. I assume that the other repo types will be file or OCI?

We were discussing both local RPM repositories + `.rpm` files, but realistically I do not think we will land the support for these until the code freeze for `3.0`, so it will probably be for `3.1`. I have created #128 so that we can track this.

pkg/rpm/resolver/resolver.go
ipetrov117: It makes sense, yes. In the `generateAddRepoStr` we will concatenate a string that will be passed to the resolver Dockerfile.

atanasdinov (approved):

PR #126, opened 2024-01-22
Introduce copy multiple files logic to fileio package

ipetrov117:
This PR introduces logic that allows us to:

1. Copy multiple files from a single directory
2. Copy files with a specific file extension from a single directory
3. Copy full directory structure along with all files that the structure holds
4. Copy full directory structure along with files that have a specific file extension

Note: This PR is a prerequisite for the RPM resolver GPG validation logic PR that is currently being implemented.

### DIFF
--- a/pkg/fileio/file_io.go
+++ b/pkg/fileio/file_io.go
@@ -5,6 +5,9 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
+
+	"go.uber.org/zap"
 )
 
 const (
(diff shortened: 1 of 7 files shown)


pkg/fileio/file_io.go
```diff
@@ -62,6 +65,58 @@ func CopyFileN(src io.Reader, dest string, perms os.FileMode, n int64) error {
...
+//
+// If `copySubDir` is set to false, copies files only from 'src' directory
+// and does not iterate over sub-directories.
+//
+// If `copySubDir` is set to true, iterates through all sub-directories
+// and copies the directory tree along with all the files.
+//
+// If `copySubDir` is used with 'ext', iterates through all sub-directories
+// and only copies files with the specified extention.
+func CopyFiles(src, dest, ext string, copySubDir bool) error {
```
atanasdinov, line 80: My only concern is that this function is attempting to do a lot. Do we need all the use cases?

atanasdinov (approved):

pkg/fileio/file_io.go
ipetrov117: The idea is for the full functionality for this function to be utilised.
Functionality utilisations are as follows:

1. Copying all files with specific extension from the root directory (without recursively going in sub dirs) is going to be used for copying local RPMs and GPGs (will come with the GPG PR)
2. The copy full directory logic is introduced with #128 in mind, where it will be used to copy the full local RPM repository that the user provides. 

Granted #128 will not be part of the GPG PR, but will be added later-on. Still I did not see the value of adding a function that partially does something, when we will soon need to rewrite it with the full functionality that the current implementation gives.

On the note of splitting this function on several functions. IMO this will offer more complexity, since the recursion logic will become too complex and will not be as easy to follow.

atanasdinov: Thanks for the explanation. I'm good with this landing as is.

jdob (approved):

PR #127, opened 2024-01-23
Registry logic restructuring

dbw7:
This PR contains a lot of restructuring based on the conversations we had. A lot of the tests need to be updated and this implementation is not complete but I’m putting it up as a draft to get a second set of eyes on it. There also may be some lingering parts that don’t necessarily make sense at the moment.

To this though, I’ve tested the current implementation and this is deploying exactly what was being deployed before the restructure without any change to the image definition file.

### DIFF
--- a/pkg/combustion/registry.go
+++ b/pkg/combustion/registry.go
@@ -15,12 +15,10 @@ import (
 )
 
 const (
-	haulerManifestYamlName = "hauler-manifest.yaml"
-	registryScriptName     = "13-embedded-registry.sh"
-	registryTarName        = "embedded-registry.tar.zst"
-	registryComponentName  = "embedded artifact registry"
-	registryLogFileName    = "embedded-registry.log"
-	hauler                 = "hauler"
+	registryScriptName    = "13-embedded-registry.sh"
+	registryComponentName = "embedded artifact registry"
+	registryLogFileName   = "embedded-registry.log"
+	hauler                = "hauler"
 )
 
 //go:embed templates/hauler-manifest.yaml.tpl
@@ -30,31 +28,13 @@ var haulerManifest string
 var registryScript string
 
 func configureRegistry(ctx *image.Context) ([]string, error) {
-	if IsEmbeddedArtifactRegistryEmpty(ctx.ImageDefinition.EmbeddedArtifactRegistry) {
+	if IsEmbeddedArtifactRegistryAndKubernetesManifestsEmpty(ctx) {
 		log.AuditComponentSkipped(registryComponentName)
 		return nil, nil
 	}
 
-	err := writeHaulerManifest(ctx)
-	if err != nil {
-		log.AuditComponentFailed(registryComponentName)
-		return nil, fmt.Errorf("writing hauler manifest: %w", err)
-	}
-
-	err = populateHaulerStore(ctx)
-	if err != nil {
-		log.AuditComponentFailed(registryComponentName)
-		return nil, fmt.Errorf("populating hauler store: %w", err)
-	}
-
-	err = generateRegistryTar(ctx)
-	if err != nil {
-		log.AuditComponentFailed(registryComponentName)
-		return nil, fmt.Errorf("generating hauler store tar: %w", err)
-	}
-
 	haulerBinaryPath := fmt.Sprintf("hauler-%s", string(ctx.ImageDefinition.Image.Arch))
-	err = copyHaulerBinary(ctx, haulerBinaryPath)
+	err := copyHaulerBinary(ctx, haulerBinaryPath)
 	if err != nil {
 		log.AuditComponentFailed(registryComponentName)
 		return nil, fmt.Errorf("copying hauler binary: %w", err)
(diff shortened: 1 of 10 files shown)


pkg/image/definition.go
```diff
@@ -120,10 +120,11 @@ type HelmChart struct {
...
 type Kubernetes struct {
-	Version   string    `yaml:"version"`
-	Network   Network   `yaml:"network"`
-	Nodes     []Node    `yaml:"nodes"`
-	Manifests Manifests `yaml:"manifests"`
+	Version    string      `yaml:"version"`
+	Network    Network     `yaml:"network"`
+	Nodes      []Node      `yaml:"nodes"`
+	Manifests  Manifests   `yaml:"manifests"`
+	HelmCharts []HelmChart `yaml:"charts"`
```
atanasdinov, line 127: This is not yet dropped from the `EmbeddedArtifactRegistry`.

pkg/combustion/templates/13-embedded-registry.sh.tpl
```diff
@@ -4,7 +4,7 @@ set -euo pipefail
 mkdir /opt/hauler
 
 mount /usr/local
-mv {{ .EmbeddedRegistryTar }} /opt/hauler/{{ .EmbeddedRegistryTar }}
+mv ./registry/* /opt/hauler/
```
atanasdinov, line 7: I'd suggest to template the directory value. As mentioned in one of the previous PRs, it's very easy for the Go implementation to start using something different than a `registry` directory which will completely invalidate the script but the tests would continue passing.

pkg/combustion/registry.go
```diff
@@ -173,6 +169,28 @@ func createRegistryCommand(ctx *image.Context, commandName string, args []string
 	return cmd, logFile, nil
 }
 
-func IsEmbeddedArtifactRegistryEmpty(registry image.EmbeddedArtifactRegistry) bool {
-	return len(registry.HelmCharts) == 0 && len(registry.ContainerImages) == 0
+func IsEmbeddedArtifactRegistryAndKubernetesManifestsEmpty(ctx *image.Context) bool {
```
atanasdinov, line 172: This is now too complex. I'd suggest to invert the naming and usage.
```suggestion
func IsEmbeddedArtifactRegistryConfigured(ctx *image.Context) bool {
```

pkg/combustion/registry.go
```diff
@@ -30,50 +28,48 @@ var haulerManifest string
...
+	haulerBinaryPath := fmt.Sprintf("hauler-%s", string(ctx.ImageDefinition.Image.Arch))
+	err := copyHaulerBinary(ctx, haulerBinaryPath)
 	if err != nil {
 		log.AuditComponentFailed(registryComponentName)
-		return nil, fmt.Errorf("populating hauler store: %w", err)
+		return nil, fmt.Errorf("copying hauler binary: %w", err)
 	}
 
-	err = generateRegistryTar(ctx)
+	registryScriptNameResult, err := writeRegistryScript(ctx)
```
atanasdinov, line 43: This is the operation which should be the last in this component. All other steps are the prerequisites of having the final Combustion script written to the filesystem. This would also make templating the "registry" directory easier.

pkg/combustion/registry.go
```diff
@@ -30,50 +28,48 @@ var haulerManifest string
...
 	if err != nil {
 		log.AuditComponentFailed(registryComponentName)
-		return nil, fmt.Errorf("generating hauler store tar: %w", err)
+		return nil, fmt.Errorf("writing registry script: %w", err)
 	}
 
-	haulerBinaryPath := fmt.Sprintf("hauler-%s", string(ctx.ImageDefinition.Image.Arch))
-	err = copyHaulerBinary(ctx, haulerBinaryPath)
+	registriesDir := filepath.Join(ctx.CombustionDir, "registry")
```
atanasdinov, line 49: The dir name should be a constant in order for us to be able to easily reuse it.

pkg/combustion/registry.go
```diff
@@ -30,50 +28,48 @@ var haulerManifest string
...
 		log.AuditComponentFailed(registryComponentName)
-		return nil, fmt.Errorf("copying hauler binary: %w", err)
+		return nil, fmt.Errorf("creating registry dir: %w", err)
 	}
 
-	registryScriptNameResult, err := writeRegistryScript(ctx)
-	if err != nil {
-		log.AuditComponentFailed(registryComponentName)
-		return nil, fmt.Errorf("writing registry script: %w", err)
+	if len(ctx.ImageDefinition.EmbeddedArtifactRegistry.ContainerImages) != 0 {
```
atanasdinov, line 56: We're missing the Helm charts implementation.

pkg/combustion/registry.go
```diff
@@ -30,50 +28,48 @@ var haulerManifest string
...
 	return []string{registryScriptNameResult}, nil
 }
 
-func writeHaulerManifest(ctx *image.Context) error {
+func writeHaulerManifest(ctx *image.Context, images []image.ContainerImage, charts []image.HelmChart, haulerManifestYamlName string) error {
 	haulerManifestYamlFile := filepath.Join(ctx.BuildDir, haulerManifestYamlName)
+	registryDef := image.EmbeddedArtifactRegistry{ContainerImages: images, HelmCharts: charts}
 
-	data, err := template.Parse(haulerManifestYamlName, haulerManifest, ctx.ImageDefinition.EmbeddedArtifactRegistry)
+	data, err := template.Parse(haulerManifestYamlName, haulerManifest, registryDef)
```
atanasdinov, line 72: A unified manifest is not what we want if we're creating separate tarballs. This makes the logic too hard to follow as well as having empty manifest sections.

pkg/combustion/registry.go
```diff
@@ -173,6 +169,28 @@ func createRegistryCommand(ctx *image.Context, commandName string, args []string
 	return cmd, logFile, nil
 }
 
-func IsEmbeddedArtifactRegistryEmpty(registry image.EmbeddedArtifactRegistry) bool {
-	return len(registry.HelmCharts) == 0 && len(registry.ContainerImages) == 0
+func IsEmbeddedArtifactRegistryAndKubernetesManifestsEmpty(ctx *image.Context) bool {
+	return len(ctx.ImageDefinition.EmbeddedArtifactRegistry.ContainerImages) == 0 && len(ctx.ImageDefinition.Kubernetes.HelmCharts) == 0
+}
+
+func writeHaulerManifestAndGenerateTar(ctx *image.Context, images []image.ContainerImage, charts []image.HelmChart, registryOrigin string) error {
```
atanasdinov, line 176: On my previous note, once you start having to call functions `doThisAndThis`, they're no longer easy to understand, maintain and test. What we can do here is to split the flows and have `configureContainerImages` and `configureHelmCharts` functions or something similar. Each one would be working with its separate template and would be called based on its own condition e.g. 
```
if len(ctx.ImageDefinition.EmbeddedArtifactRegistry.ContainerImages) != 0 {
   configureContainerImages()
}
```

This will allow better testing and easier upgrades once we get to adding manifests in as well.

pkg/combustion/templates/13-embedded-registry.sh.tpl
dbw7: Sounds good 👍

pkg/combustion/registry.go
dbw7: Will do👍

dbw7:
Closing this PR based on the implementation changing and being present in #123 .

PR #130, opened 2024-01-23
Cli UX Improvements

jdob:
Prior to this patch, the output when running EIB with a definition file that doesn't exist is:
```
2024/01/23 15:42:18 CLI arguments could not be parsed: parsing image definition file raw-image.yaml: image definition file "raw-image.yaml" cannot be read: open /eib/raw-image.yaml: no such file or directory
```

This patch uses a similar approach as with image validation errors to display more user-friendly messages in the case of most user errors. The following are examples of various ways that EIB can fail to even start building an image:

### Missing Image Configuration Directory
_Note: This is the only step that occurs prior to initializing logging_
```
$ podman run --rm -it -v ~/Code/eib/:/not-there eib:dev /bin/eib -config-file ./raw-simple.yaml -config-dir /eib -build-dir /eib/_build
The specified image configuration directory '/eib' could not be found.
```

### Image Definition File does not exist
```
$ eib-run not-there
The specified definition file '/eib/not-there' could not be found.
```

### Unparsable (not-YAML) Image Definition File
```
$ eib-run not-yaml.yaml
The image definition file '/eib/not-yaml.yaml' could not be parsed. Please check the eib-build.log file under the build directory for more information.

$ cat _build/build-Jan23_20-22-06/eib-build.log
2024-01-24T19:32:17.968Z	INFO	log/audit.go:58	The image definition file '/eib/not-yaml.yaml' could not be parsed.
2024-01-24T19:32:17.968Z	ERROR	eib/main.go:176	could not parse the image definition: yaml: did not find expected whitespace or line break
main.parseImageDefinition
	/src/cmd/eib/main.go:176
main.main
	/src/cmd/eib/main.go:59
runtime.main
```

### Invalid Definition File (fails our semantic checks)
_Note: This has changed from the previous patch to point users towards the log file._
```
$ eib-run raw-invalid.yaml
  Image
    The 'arch' field is required in the 'image' section.
Please check the eib-build.log file under the build directory for more information.

$ cat _build/build-Jan23_20-23-03/eib-build.log
2024-01-24T19:33:03.935Z	ERROR	eib/main.go:226	image definition validation failures:
The 'arch' field is required in the 'image' section.

main.isImageDefinitionValid
	/src/cmd/eib/main.go:226
main.main
	/src/cmd/eib/main.go:66
runtime.main
```

### Validate Flag is Specified
```
$ eib-run raw-simple.yaml -validate
The specified image definition is valid.

$ cat _build/build-Jan23_20-23-41/eib-build.log
2024-01-23T20:23:41.008Z	INFO	log/audit.go:66	The specified image definition is valid.
```

### With RPM Dependency Resolution
_Note: This is currently failing for me on HEAD, unrelated to these changes._
```
$ eib-run raw-rpms.yaml
The services for RPM dependency resolution failed to start. Please check the eib-build.log file under the build directory for more information.

$ cat _build/build-Jan23_20-28-30/eib-build.log
2024-01-24T19:34:46.417Z	INFO	log/audit.go:58	The services for RPM dependency resolution failed to start.
2024-01-24T19:34:46.417Z	ERROR	eib/main.go:243	creating new podman connection: unable to connect to Podman socket: Get "http://d/v4.8.3/libpod/_ping": dial unix /run/podman/podman.sock: connect: no such file or directory
main.bootstrapRpmDependencyServices
	/src/cmd/eib/main.go:243
main.main
	/src/cmd/eib/main.go:77
runtime.main
	/usr/lib64/go/1.21/src/runtime/proc.go:267```

### DIFF
--- a/cmd/eib/main.go
+++ b/cmd/eib/main.go
@@ -1,9 +1,10 @@
 package main
 
 import (
+	"errors"
 	"flag"
 	"fmt"
-	"log"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"slices"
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: cmd/eib/main.go [hunk 1]
```diff
@@ -24,43 +25,176 @@ import (
...
-	flag.StringVar(&configFile, argConfigFile, "", "name of the image configuration file")
-	flag.StringVar(&configDir, argConfigDir, "", "full path to the image configuration directory")
-	flag.StringVar(&rootBuildDir, argBuildDir, "", "full path to the directory to store build artifacts")
-	flag.BoolVar(&validate, argValidate, false, "if specified, the image definition will be validated but not built")
+const (
+	exitCodeDefinitionParse   = 10
+	exitCodeDefinitionInvalid = 11
+	exitCodeConfigDir         = 12
+	exitCodeBuildDir          = 13
+	exitCodeRpmDependencies   = 14
```
atanasdinov, line 39: I'm not particularly against these but I want to raise the question of how do we track / document custom exit status codes. AFAIK there is some kind of standartisation in place but I'm not familiar enough with it. Please ignore if our lovely PM (@rdoxenham) is fine with using these.

cmd/eib/main.go
```diff
@@ -24,43 +25,176 @@ import (
...
+}
+
+func main() {
+	cliArguments := parseCliArguments()
+
+	buildDir, combustionDir, err := build.SetupBuildDirectory(cliArguments.rootBuildDir)
+	if err != nil {
+		audit.Auditf("The build directory could not be setup under the configuration directory '%s'.", cliArguments.configDir)
+		audit.Audit("Reason:")
+		audit.AuditError(err)
```
atanasdinov, line 56: The main issue I find with this sort of logging is that if (or rather when) the writer behind these changes, then it'd be a hassle to fix the messages e.g. if instead of just writing plain text to stdout via `fmt.Println` we switch to `log.Println` which includes timestamps, the "Reason" and the error logs will be separate log messages with different timestamps.

As a bonus point string formatting which is not currently detectable because it's abstracted under a method call, will instead be in place and very easy to maintain e.g. it's very easy to call `audit.AuditfAndLog` without specifying a `%s` (or any other) formatter.

I'd suggest to use a `strings.Builder` similarly to its usage for the validation and build and a log a single message instead of multiple ones.

cmd/eib/main.go
jdob: Admittedly, I don't have a ton of experience with this in the past. I found myself exiting with 1 a lot and it felt like I should at least take a shot doing it less wrong.

rdoxenham: There are quite a few commonly-used exit codes in Linux and I don't suggest that we re-use these. See here: https://tldp.org/LDP/abs/html/exitcodes.html

jdob: I intentionally didn't use any in that table. Thankfully, the range is available codes is pretty large.

rdoxenham: Yeah, they are, but if you look in other sources it seems that exit codes under 133 are also widely used, so that link I shared isn't exhaustive. Maybe we can go with 166 and above just to be on the safe side?

jdob: I'm also completely fine with just using 1 for all error exists and then revisiting on the off chance anyone ever asks us for this. That's probably the safest approach to start with, and I'd rather add complexity when it's actually needed/going to be used.

cmd/eib/main.go
```diff
@@ -24,43 +25,170 @@ import (
...
+			zap.S().Error(err)
+		}
+		return nil
 	}
 
-	buildDir, combustionDir, err := build.SetupBuildDirectory(rootBuildDir)
+	configData, err := os.ReadFile(definitionFilePath)
 	if err != nil {
-		return nil, fmt.Errorf("setting up build directory: %w", err)
+		audit.AuditInfof("The specified definition file '%s' could be read. %s", definitionFilePath, checkLogMessage)
```
atanasdinov, line 175:
```suggestion
		audit.AuditInfof("The specified definition file '%s' could not be read. %s", definitionFilePath, checkLogMessage)
```

cmd/eib/main.go
```diff
@@ -24,43 +25,170 @@ import (
...
+		} else {
+			audit.AuditInfof("Unable to check the filesystem for the image definition file '%s'. %s",
+				definitionFilePath, checkLogMessage)
+			zap.S().Error(err)
+		}
+		return nil
 	}
 
-	buildDir, combustionDir, err := build.SetupBuildDirectory(rootBuildDir)
+	configData, err := os.ReadFile(definitionFilePath)
```
atanasdinov, line 173: This will already return `fs.ErrNotExist` if the file doesn't exist so we can skip the `os.Stat` call if you'd like.

atanasdinov (approved):

PR #134, opened 2024-01-25
Manifest handling fixes

dbw7:
The previous version of manifest handling worked well for single-document yaml manifests, however, would fail to completely parse multi-document yaml manifests such as:
```
apiVersion: v1
kind: Service
metadata:
  name: my-nginx-svc
  labels:
    app: nginx
spec:
  type: LoadBalancer
  ports:
  - port: 80
  selector:
    app: nginx
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: my-nginx
  labels:
    app: nginx
spec:
  replicas: 3
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx:1.14.2
        ports:
        - containerPort: 80


```
So this update to manifest handling fixes that in addition to updating other small bugs.


pkg/registry/manifests.go
```diff
@@ -64,23 +68,31 @@ func GetAllImages(embeddedContainerImages []image.ContainerImage, manifestURLs [
...
-		return nil, fmt.Errorf("error reading manifest: %w", err)
+		return nil, fmt.Errorf("error opening manifest: %w", err)
 	}
 
-	if len(manifestData) == 0 {
-		return nil, fmt.Errorf("invalid manifest")
+	var manifests []any
+	decoder := yaml.NewDecoder(manifestFile)
+	for {
+		var manifest any
```
atanasdinov, line 80: I love that we're able to parse each YAML object one by one. Please consider using this structure in order to make the parsing step a little less "generic":
```suggestion
	var manifests []map[string]any
	decoder := yaml.NewDecoder(manifestFile)
	for {
		var manifest map[string]any
```

The output of the `readManifest` function will then be of type `[]map[string]any` and we should only need to make sure that `storeManifestImageNames` accepts it directly and works over it:
```
func storeManifestImageNames(data []map[string]any...
```

```
for _, d := range data {
	findImages(d)
}
```

pkg/registry/manifests_test.go
```diff
@@ -70,7 +78,7 @@ func TestFindImagesInManifest(t *testing.T) {
 	manifestData, err := readManifest(manifestPath)
```
atanasdinov, line 78: Additional note to the suggestion above. This only works because we're dealing with `any` type. The `GetAllImages` function iterates over the manifests when calling `storeManifestImageNames` whereas this one does not:

```
// GetAllImages(...)
manifests, err := readManifest(manifestPath)
if err != nil {
	return nil, fmt.Errorf("error reading manifest %w", err)
}

for _, manifestData := range manifests {
	storeManifestImageNames(manifestData, extractedImagesSet)
}
```

pkg/combustion/registry.go
```diff
@@ -45,9 +45,9 @@ func configureRegistry(ctx *image.Context) ([]string, error) {
...
 
-	var localManifestSrcDir = filepath.Join(ctx.ImageConfigDir, "kubernetes", "manifests")
-
-	configured := isComponentConfigured(ctx, localManifestSrcDir)
+	localManifestSrcDir := filepath.Join(ctx.ImageConfigDir, "kubernetes", "manifests")
+	localManifestComponentDir := filepath.Join("kubernetes", "manifests")
+	configured := isComponentConfigured(ctx, localManifestComponentDir)
 	if !configured {
 		localManifestSrcDir = ""
 	}
```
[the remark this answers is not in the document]
dbw7: I like that 👍

pkg/registry/manifests_test.go
```diff
@@ -43,7 +51,7 @@ func TestReadManifestNoManifest(t *testing.T) {
 
 func TestReadManifestInvalidManifest(t *testing.T) {
 	// Setup
-	manifestPath := filepath.Join("testdata", "invalid-crd.yaml")
+	manifestPath := filepath.Join("testdata", "invalid-crd.yml")
```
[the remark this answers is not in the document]
dbw7: Yes that's a side effect of showing that we verify for ignore anything that isn't `.yaml` or `.yml`

pkg/registry/manifests.go
dbw7: Done, thanks 👍

atanasdinov (approved):

PR #135, opened 2024-01-26
Manifest deployment

dbw7:
This covers the complete manifest deployment with the exception of configuring k8s to use the hauler registry.

Putting this up as a draft because, while this is almost complete, I want opinions on whether or not the tests are comprehensive enough. Additionally, there’s some inconsistency around a lot of the “path” usage that I can’t decide on.

To be merged after #134  because of duplicate fixes in #134 that were needed for this.


pkg/registry/manifests.go
```diff
@@ -147,3 +160,35 @@ func downloadManifests(manifestURLs []string, destPath string) ([]string, error)
 
 	return manifestPaths, nil
 }
+
+func CopyManifests(src string, dest string) ([]string, error) {
```
atanasdinov, line 164: Is all of this replaceable by `fileio.CopyFiles()`?

pkg/combustion/kubernetes.go
```diff
@@ -57,6 +58,25 @@ func configureKubernetes(ctx *image.Context) ([]string, error) {
 		return nil, fmt.Errorf("configuring kubernetes components: %w", err)
 	}
 
+	manifestURLs := ctx.ImageDefinition.Kubernetes.Manifests.URLs
```
atanasdinov, line 61: All of this belongs to `configureManifest` instead. Here we should only be doing:

```
if err = configureManifests(...); err != nil {
	log.AuditComponentFailed(k8sComponentName)
	return nil, fmt.Errorf("configuring kubernetes manifests: %w", err)
}
```

pkg/combustion/kubernetes.go
```diff
@@ -57,6 +58,25 @@ func configureKubernetes(ctx *image.Context) ([]string, error) {
...
+	localManifestsComponentDir := filepath.Join("kubernetes", "manifests")
```
atanasdinov, line 62: There's already a `k8sDir` constant so let's reuse it and add one for `manifests`.

pkg/combustion/kubernetes.go
```diff
@@ -57,6 +58,25 @@ func configureKubernetes(ctx *image.Context) ([]string, error) {
...
+	localManifestsConfigured := isComponentConfigured(ctx, localManifestsComponentDir)
+	if localManifestsConfigured || len(manifestURLs) != 0 {
+		localManifestsSrcDir := filepath.Join(ctx.ImageConfigDir, "kubernetes", "manifests")
+		k8sCombustionDir := filepath.Join(ctx.CombustionDir, k8sDir)
+		err = os.MkdirAll(k8sCombustionDir, os.ModePerm)
```
atanasdinov, line 67: This is something the artefact downloader is already doing. I'd suggest to just use another directory on the root combustion level until I sort out the kubernetes/ structure.

pkg/registry/manifests.go
dbw7: The benefit of this function is that it checks for `.yaml` and `.yml` + appends a tag to local files to prevent duplicate names. The previously mentioned `local-manifest` and `downloaded-manifest` solution ends up being moot cause all of the files end up in `...rke2/servers/manifests`.

pkg/combustion/kubernetes.go
dbw7: I was thinking about that, will refactor

pkg/combustion/kubernetes.go
dbw7: Will update 👍

pkg/combustion/kubernetes.go
```diff
@@ -114,16 +121,23 @@ func configureRKE2(ctx *image.Context) (string, error) {
 		return "", fmt.Errorf("downloading RKE2 artefacts: %w", err)
 	}
 
+	manifestsPath := ""
+	if isComponentConfigured(ctx, filepath.Join(k8sDir, "manifests")) {
```
atanasdinov, line 125: This is not really correct since it doesn't include manifests from the image definition.

pkg/combustion/kubernetes.go
```diff
@@ -114,16 +121,23 @@ func configureRKE2(ctx *image.Context) (string, error) {
...
 	}{
-		Kubernetes:  ctx.ImageDefinition.Kubernetes,
-		ConfigFile:  configFile,
-		InstallPath: installPath,
-		ImagesPath:  imagesPath,
+		Kubernetes:    ctx.ImageDefinition.Kubernetes,
+		ConfigFile:    configFile,
+		InstallPath:   installPath,
+		ImagesPath:    imagesPath,
+		ManifestsPath: manifestsPath,
```
atanasdinov, line 140: It is very weird to render the template with the directory and only after that actually attempt to `configureManifests()`. The more I think about it, the more I believe that `configureManifests()` should also be called as part of this function instead of using `configureKubernetes()` for them.

If the concern here is that the same would apply for k3s installations, I'd rather leave that for me once I get to introducing those changes. In the meantime, it's best not to mix up the internal state - `manifestsPath` and parsing the template should only be done if `configureManifests()` is successful.

pkg/combustion/kubernetes_test.go
```diff
@@ -186,6 +191,102 @@ func TestConfigureKubernetes_SuccessfulRKE2Server(t *testing.T) {
 	assert.Equal(t, "cilium", configContents["cni"], "default CNI is not set")
 }
 
+func TestConfigureKubernetes_SuccessfulRKE2ServerWithLocalManifests(t *testing.T) {
```
atanasdinov, line 194: I'd suggest to reuse the full successful example instead of repeating 85% of it for the manifests scenario.

pkg/combustion/kubernetes_test.go
```diff
@@ -302,3 +403,57 @@ func TestExtractCNI(t *testing.T) {
...
+	}
+
+	// Test
+	err := configureManifests(ctx)
+
+	// Verify
+	require.ErrorContains(t, err, "downloading manifests to combustion dir: downloading manifest 'k8s.io/examples/application/nginx-app.yaml': executing request: Get \"k8s.io/examples/application/nginx-app.yaml\": unsupported protocol scheme \"\"")
+}
+
+func TestConfigureKubernetes_InvalidManifestURL(t *testing.T) {
```
atanasdinov, line 423: Feel free to drop this case, I don't see it bringing anything to the table since `configureManifests()` does not exist outside of `configureKubernetes()`.

pkg/combustion/kubernetes_integration_test.go
```diff
@@ -0,0 +1,122 @@
...
+		},
+	}
+	ctx.KubernetesArtefactDownloader = mockKubernetesArtefactDownloader{
+		downloadArtefacts: func(arch image.Arch, version, cni string, multusEnabled bool, destPath string) (string, string, error) {
+			return serverInstaller, serverImages, nil
+		},
+	}
+
+	ctx.ImageDefinition.Kubernetes.Manifests.URLs = []string{
+		"https://k8s.io/examples/application/nginx-app.yaml",
```
atanasdinov, line 65: It doesn't seem right with me at all that we have to perform an actual download here instead of mocking the call. I guess this can land as is but we desperately need an interface behind the registry calls.

pkg/combustion/kubernetes_test.go
dbw7: Yeah was think that as well, will update 👍

pkg/combustion/kubernetes_test.go
dbw7: 👍

pkg/combustion/templates/15-rke2-single-node-installer.sh.tpl
```diff
@@ -2,13 +2,18 @@
...
 cp {{ .imagesPath }}/* /var/lib/rancher/rke2/agent/images/
 
 {{- if .vipManifest }}
 mkdir -p /var/lib/rancher/rke2/server/manifests/
 cp {{ .vipManifest }} /var/lib/rancher/rke2/server/manifests/{{ .vipManifest }}
 {{- end }}
+{{ if .manifestsPath }}
+mkdir -p /var/lib/rancher/rke2/server/manifests/
+cp {{ .manifestsPath }}/* /var/lib/rancher/rke2/server/manifests/
+{{ end }}
```
atanasdinov, line 16: Nit:
```suggestion
{{- if .manifestsPath }}
mkdir -p /var/lib/rancher/rke2/server/manifests/
cp {{ .manifestsPath }}/* /var/lib/rancher/rke2/server/manifests/
{{- end }}
```

I'd also put empty lines before and after this conditional block. This way the template becomes more readable.

pkg/combustion/kubernetes_integration_test.go
```diff
@@ -0,0 +1,173 @@
...
+	require.NoError(t, os.Mkdir(k8sCombDir, os.ModePerm))
+
+	localManifestsSrcDir := filepath.Join(ctx.ImageConfigDir, k8sDir, manifestsDir)
+	require.NoError(t, os.MkdirAll(localManifestsSrcDir, 0o755))
+
+	localSampleManifestPath1 := filepath.Join("..", "registry", "testdata", "sample-crd.yaml")
+	err := fileio.CopyFile(localSampleManifestPath1, filepath.Join(localManifestsSrcDir, "sample-crd.yaml"), fileio.NonExecutablePerms)
+	require.NoError(t, err)
+
+	localSampleManifestPath2 := filepath.Join("..", "registry", "testdata", "invalid-crd.yml")
```
atanasdinov, line 87: Something we discussed last week is that we don't particularly want to do this since it suggests that we're able to ship invalid manifests. Can we drop it?

pkg/combustion/kubernetes_integration_test.go
```diff
@@ -0,0 +1,173 @@
...
+
+	info, err := os.Stat(scriptPath)
+	require.NoError(t, err)
+
+	assert.Equal(t, fileio.ExecutablePerms, info.Mode())
+
+	b, err := os.ReadFile(scriptPath)
+	require.NoError(t, err)
+
+	contents := string(b)
```
atanasdinov, line 106: Can we include an assertion for the manifests path templating?

pkg/combustion/templates/15-rke2-multi-node-installer.sh.tpl
```diff
@@ -23,6 +23,11 @@ mount /var
 mkdir -p /var/lib/rancher/rke2/agent/images/
 cp {{ .imagesPath }}/* /var/lib/rancher/rke2/agent/images/
 
+{{ if .manifestsPath }}
+mkdir -p /var/lib/rancher/rke2/server/manifests/
```
atanasdinov, line 27: Did we confirm if this is something that we want to perform on all nodes as opposed to just agents for example?

atanasdinov (approved):

PR #140, opened 2024-01-29
Enable GPG validation by default in the EIB resolver image

ipetrov117:
This PR enables GPG validation in the EIB resolver by default.

Validation is done in the following way:

1. All additional repositories added to the [packages configuration](https://github.com/suse-edge/edge-image-builder/blob/main/pkg/image/definition.go#L73) must have a valid GPG key in their `repodata` directory. The EIB resolver will then automatically import the keys of each third-party repo using zypper's `--gpg-auto-import-keys` global option (doc [here](https://en.opensuse.org/SDB:Zypper_manual_(plain))). The user is responsible for making sure that the third-party repository key is valid. If the third-party repository does not contain a valid GPG key, then EIB will fail.
2. All side-loaded RPMs must be signed with valid GPG keys. All GPG keys used in side-loaded RPM signatures must be provided in the `<config-dir>/rpms/gpg-keys` directory. These keys will then be imported in the EIB resolver image. After that the EIB resolver will validate the signatures of each RPM in the `<config-dir>/rpms` directory. If an RPM contains an empty or unrecognised signature, then EIB will fail.

For users that would want to add either an unsigned third-party repository or an unsigned side-loaded RPM, then they would need to pass the following configuration to the packages setup:
```yaml
...
operatingSystem:
  packages:
    noGPGCheck: true
...
```
The `noGPGCheck: true` configuration effectively disables GPG validation for all additional repositories and side-loaded RPMs providing the user with the chance to run non-signed repos/rpms.

**Note:** We continue to use the [RPM repository](https://github.com/suse-edge/edge-image-builder/blob/main/pkg/combustion/templates/10-rpm-install.sh.tpl#L9) that the EIB resolver creates with the `--no-gpg-checks` option. This is done, because all validation of the packages that make the contents of this repository has been done in the EIB resolver image, thus giving us the assurance that the contents of this repo are not corrupted.


pkg/combustion/rpm.go
```diff
@@ -32,13 +33,25 @@ func configureRPMs(ctx *image.Context) ([]string, error) {
...
+	var localRPMConfig *image.LocalRPMConfig
 	if isComponentConfigured(ctx, userRPMsDir) {
-		rpmDir = generateComponentPath(ctx, userRPMsDir)
+		rpmDir := generateComponentPath(ctx, userRPMsDir)
+		localRPMConfig = &image.LocalRPMConfig{
+			RPMPath: rpmDir,
+		}
+
+		gpgPath := filepath.Join(rpmDir, userGPGsDir)
+		_, err := os.Stat(gpgPath)
```
atanasdinov, line 44: Do we want to check the `NoGPGCheck` flag here? Audit and log the warning if true and fail right away if true but the path exists?

pkg/image/definition.go
```diff
@@ -69,6 +69,7 @@ type OperatingSystem struct {
 }
 
 type Packages struct {
+	NoGPGCheck      bool      `yaml:"noGPGCheck"`
```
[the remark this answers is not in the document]
ipetrov117: We decided to move away from the more granular `gpg-check` approach, mainly for two reasons:

1. It will offer significant complexity to the code for something that we do not want to encourage
2. If we do it this way we essentially encourage the usage of non-signed repos/rpms along with signed one's. Which IMO is a disaster waiting to happen

Disabling GPG checks should be done for dev/test purposes only and by having only a global switch IMO this will enforce users to GPG sign their repos/rpms more than having a granular approach to GPG validation.

This will be thoroughly documented in the EIB RPM resolver documentation PR which I am currently working on.

pkg/combustion/rpm_test.go
```diff
@@ -245,14 +245,20 @@ func TestConfigureRPMSSuccessfulConfig(t *testing.T) {
...
 			return nil
 		},
 	}
 
 	ctx.RPMResolver = mockRPMResolver{
-		resolveFunc: func(packages *image.Packages, localPackagesPath, outputDir string) (rpmDir string, pkgList []string, err error) {
+		resolveFunc: func(packages *image.Packages, localRPMConfig *image.LocalRPMConfig, outputDir string) (string, []string, error) {
+			require.NotNil(t, localRPMConfig)
+			assert.Equal(t, rpmDir, localRPMConfig.RPMPath)
+			assert.Equal(t, gpgDir, localRPMConfig.GPGKeysPath)
```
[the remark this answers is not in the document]
ipetrov117: Agreed, will fix that asap.

pkg/rpm/resolver/templates/Dockerfile.tpl
```diff
@@ -1,46 +1,69 @@
...
+#  LocalRPMList - list of local RPMs for which dependency resolution has to be done
+#  LocalGPGList - list of local GPG keys that will be imported in the resolver image
+#  NoGPGCheck   - when set to true skips the GPG validation for all third-party repositories and local RPMs
+FROM {{ .BaseImage }}
 
 {{ if and (ne .FromRPMPath "") (ne .ToRPMPath "") -}}
-COPY {{ .FromRPMPath }} {{ .ToRPMPath -}}
-{{ end }}
+COPY {{ .FromRPMPath }} {{ .ToRPMPath }}
+{{ if and (ne .FromGPGPath "") (ne .ToGPGPath "") -}}
```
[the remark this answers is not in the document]
ipetrov117: I will test this out and update all occurrences that rely on such logic.

pkg/image/definition.go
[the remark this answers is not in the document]
atanasdinov: Thanks for the clarification! Do we want to then fail the build by default if `gpg-keys` dir is not present in order to "enforce" this default behaviour?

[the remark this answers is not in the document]
ipetrov117: So currently:

1.  If the RPM dir is present and the `noGPGCheck: true`, then we do not fail and we do all installation options with `--no-gpg-check`
2. If the RPM dir is present, but no GPG dir is present and `noGPGCheck: false`, then the build will fail because the validation of the RPMs will not succeed - RPMs will either be missing signatures, or have an unrecognised (non-imported) key signature.

There is also a corner case in that you may side-load an RPM whose GPG key comes from a third-party repository that you have added to the EIB configuration, then the build will not fail, because the GPG key will be already present on the machine since it will be added from the repo itself (using the `--gpg-auto-import-keys`).

So that said I think that failing the build by default will brake this corner case. I think that the current validation is enough, WDYT?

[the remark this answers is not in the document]
ipetrov117: That said I can definitely validate and fail if all of the below is true:

1. `additionalRepositories` is missing
2. `noGPGCheck: false`
3. `gpg-keys` directory is missing

Adding this to my TODO.

pkg/combustion/rpm.go
```diff
@@ -32,13 +35,38 @@ func configureRPMs(ctx *image.Context) ([]string, error) {
...
+
+		packages := ctx.ImageDefinition.OperatingSystem.Packages
+		switch {
+		case err == nil:
+			if !packages.NoGPGCheck {
+				localRPMConfig.GPGKeysPath = gpgPath
+			} else {
+				log.Auditf("WARNING: Found %s directory, when GPG validation is disabled", gpgPath)
```
atanasdinov, line 54: Isn't this where we want to fail due to a conflict? The audit and warning logs I mentioned earlier referred to notifying the user that setting `NoGPGCheck: true` is in general not advised.

pkg/combustion/rpm.go
ipetrov117: Added the following validation:

1. If the `gpg-keys` dir exists, but  `noGPGChecks: true`, we audit and log a warning
2. If the `gpg-keys` dir does not exist, but `noGPGChecks: false`, we fail right away and do not go into any resolution logic

atanasdinov (approved):

pkg/combustion/rpm.go
ipetrov117: I didn't keep EIB's failure convention in mind. This has been fixed, if we have a GPG dir, but GPG validation is disabled we fail instead of giving a warning. Furthermore, I added a warning log when using the `noGPGChecks: true` parameter.

pkg/combustion/rpm.go
```diff
@@ -32,13 +35,42 @@ func configureRPMs(ctx *image.Context) ([]string, error) {
...
+
+		gpgPath := filepath.Join(rpmDir, userGPGsDir)
+		_, err := os.Stat(gpgPath)
+		switch {
+		case err == nil:
+			if !packages.NoGPGCheck {
+				localRPMConfig.GPGKeysPath = gpgPath
+			} else {
+				log.AuditComponentFailed(rpmComponentName)
+				return nil, fmt.Errorf("found existing '%s' directory, but GPG validaiton is disabled", userGPGsDir)
```
rdoxenham, line 59:
```suggestion
				return nil, fmt.Errorf("found existing '%s' directory, but GPG validation is disabled", userGPGsDir)
```

pkg/combustion/rpm.go
```diff
@@ -32,13 +35,42 @@ func configureRPMs(ctx *image.Context) ([]string, error) {
...
+			if !packages.NoGPGCheck {
+				localRPMConfig.GPGKeysPath = gpgPath
+			} else {
+				log.AuditComponentFailed(rpmComponentName)
+				return nil, fmt.Errorf("found existing '%s' directory, but GPG validation is disabled", userGPGsDir)
+			}
+		case errors.Is(err, fs.ErrNotExist):
+			if !packages.NoGPGCheck {
+				log.AuditComponentFailed(rpmComponentName)
+				return nil, fmt.Errorf("GPG validation is enabled, but '%s' directory is missing", userGPGsDir)
```
rdoxenham, line 64: We should likely add an additional clarifying warning here that because they're side-loading RPM's, this check is triggered. It's not there if the user is simply adding repo's or trying to install packages via the list.

```suggestion
				return nil, fmt.Errorf("GPG validation is enabled, but '%s' directory is missing for side-loaded RPMs", userGPGsDir)
```

pkg/combustion/rpm_test.go
```diff
@@ -222,6 +222,55 @@ func TestConfigureRPMSError(t *testing.T) {
...
+			createGPGDir: true,
+			pkgs: image.Packages{
+				NoGPGCheck: true,
+			},
+			expectedErr: fmt.Sprintf("found existing '%s' directory, but GPG validation is disabled", userGPGsDir),
+		},
+		{
+			name:        "Enabled GPG validation, but missing GPG dir",
+			pkgs:        image.Packages{},
+			expectedErr: "GPG validation is enabled, but 'gpg-keys' directory is missing",
```
rdoxenham, line 252:
```suggestion
			expectedErr: "GPG validation is enabled, but 'gpg-keys' directory is missing for side-loaded RPMs",
```

rdoxenham (approved):
Fully tested successfully... both with/without gpg-keys. Thank you! LGTM 🚀

jdob (approved):

PR #141, opened 2024-01-29
Download k3s artefacts

atanasdinov:


jdob (approved):
This PR looks good to land. I would like to continue the discussion of whether or not we embed both installers (RKE2 and K3s) in the container image v. downloading, but that decision will apply to both k8s flavors and if we decide on changes will be a future PR.

PR #143, opened 2024-01-29
Added ability to include CA certificates in the built image

jdob:
Misc other fixes:
* Reordered directory listing in building-images to be alphabetical
* Removed blank line and corrected log message capitalization in elemental.go
* Fixed typo in files_io.go


pkg/combustion/elemental.go
```diff
@@ -24,10 +24,9 @@ const (
 var elementalScript string
 
 func configureElemental(ctx *image.Context) ([]string, error) {
-
 	if !isComponentConfigured(ctx, elementalConfigDir) {
 		log.AuditComponentSkipped(elementalComponentName)
-		zap.S().Info("Skipping elemental registration component, configuration is not provided")
+		zap.S().Info("skipping elemental registration component, configuration is not provided")
```
atanasdinov, line 29: I'm pretty sure 95%+ of the log messages are capitalised. Do we want to use lower cased ones instead?

pkg/combustion/certificates.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/suse-edge/edge-image-builder/pkg/fileio"
+	"github.com/suse-edge/edge-image-builder/pkg/image"
+	"github.com/suse-edge/edge-image-builder/pkg/log"
+	"github.com/suse-edge/edge-image-builder/pkg/template"
+	"go.uber.org/zap"
+)
+
+const (
+	certsComponentName = "certificates"
+	certsScriptName    = "30-certificates.sh"
```
atanasdinov, line 18: This seems rather low on the priority list. The task itself mentions:
> but likely needs to do it before Elemental/SUMA/Proxy is used.

I'm also wondering if this component should also differentiate the usages e.g. we'd eventually want to be able to provide custom certificates for Kubernetes clusters ([RKE2 docs for reference](https://docs.rke2.io/security/certificates#using-custom-ca-certificates)). Is this something we'd track separately? @rdoxenham

rdoxenham: I agree, this should be moved to a much higher priority in my opinion. If the proxy has a self-signed certificate (or they want to make sure it's actually the right proxy server being used), or the same situation for SUMA/Elemental/etc. then we need to make sure that the certificates are imported much earlier in the process. Suggest we go with 07 so it runs before the proxy is setup, especially as the proxy may be required for *other* things.

docs/building-images.md
```diff
@@ -144,14 +144,16 @@ the structure of this directory will be better fleshed out. For now, the require
 
 There are a number of optional directories that may be included in the image configuration directory:
 
-* `network` - If present, network configurations will be generated from all desired states in this directory
-  and will be included in the built image. The configurations relevant for the particular host will be identified
-  and applied during the combustion phase. Check [nm-configurator](https://github.com/suse-edge/nm-configurator/)
-  for more information.
+* `certificates` - If present, all files with the extension ".pem" will be installed as CA certificates
+  in the built image.
```
rdoxenham, line 148:
```suggestion
* `certificates` - If present, all files with the extension ".pem" will be installed as CA certificates
  in the built image. Note that ".crt" files are permitted, but please ensure they have a ".pem" extension.
```

pkg/combustion/templates/30-certificates.sh.tpl
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+set -euo pipefail
+
+cp ./{{ .CertificatesDir }}/* /etc/pki/trust/anchors/.
+update-ca-certificates
```
rdoxenham, line 5: We may want to think about making this output verbose so users can see this happening during boot up logs.

```suggestion
update-ca-certificates -v
```

rdoxenham (changes requested):
We may want to think about ".crt" files as well... these should work as `update-ca-certificates` should process these accordingly. Right now the code explicitly looks for ".pem" files so I've added a suggestion for the documentation, but perhaps we also allow the user to specify ".crt" too? Also a note that we should make the certificates happen much earlier on in the boot process. Thanks!

pkg/combustion/certificates.go
rdoxenham: @atanasdinov yes, we would definitely want to provide a way of specifying Kubernetes CA certificates. Maybe a subdirectory within the `cerfificates/` directory (or perhaps one inside of `kubernetes/`?) would be great to have at a later date. We should track this separately.

atanasdinov: Probably it makes more sense for those to be under `kubernetes/` then. Thanks for the clarification.

pkg/combustion/elemental.go
jdob: I may have been confusing myself with error messages. I'm happy to revert.

docs/building-images.md
```diff
@@ -144,14 +144,16 @@ the structure of this directory will be better fleshed out. For now, the require
 
 There are a number of optional directories that may be included in the image configuration directory:
 
-* `network` - If present, network configurations will be generated from all desired states in this directory
-  and will be included in the built image. The configurations relevant for the particular host will be identified
-  and applied during the combustion phase. Check [nm-configurator](https://github.com/suse-edge/nm-configurator/)
-  for more information.
+* `certificates` - If present, all files with the extension ".pem" will be installed as CA certificates
+  in the built image. Note that ".crt" files are permitted, but please ensure they have a ".pem" extension.
```
atanasdinov, line 148: The second sentence is no longer applicable.

atanasdinov (approved):

rdoxenham (approved):
LGTM 🚀

PR #145, opened 2024-01-31
Complete air-gapped manifest deployment

dbw7:
This is the finale in the manifest deployment saga, excluding helm charts. This includes the final piece of the puzzle for actually using all of the previous merged implementations for completely air-gapped manifest deployment.

### DIFF
--- a/pkg/combustion/registry.go
+++ b/pkg/combustion/registry.go
@@ -35,7 +35,7 @@ var haulerManifest string
 var registryScript string
 
 //go:embed templates/registries.yaml.tpl
-var k8sRegistriesManifest string
+var k8sRegistryMirrors string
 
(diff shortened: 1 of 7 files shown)


pkg/combustion/registry.go
```diff
@@ -68,6 +73,14 @@ func configureRegistry(ctx *image.Context) ([]string, error) {
 		return nil, fmt.Errorf("getting all container images: %w", err)
 	}
 
+	hostnames := getImageHostnames(containerImages)
+
+	err = writeRegistriesManifest(ctx, hostnames)
```
atanasdinov, line 78: Can we only do this if Kubernetes is configured (i.e. `Version != ""`)?

PR #146, opened 2024-01-31
Configure multi node k3s

atanasdinov:


pkg/combustion/kubernetes.go
```diff
@@ -147,9 +159,24 @@ func configureK3S(ctx *image.Context) (string, error) {
 
 		templateValues["configFile"] = k8sServerConfigFile
 		templateValues["vipManifest"] = vipManifest
+
+		return storeKubernetesInstaller(ctx, "single-node-k3s", k3sSingleNodeInstaller, templateValues)
+	}
+
+	log.Audit("WARNING: An external IP address for the Ingress controller (Traefik) must be manually configured in multi node clusters")
```
jdob, line 166: Nit: Audit messages (like the ones for validation) end with punctuation.

jdob (approved):
Untested, but from a codestand point LGTM.

pkg/combustion/kubernetes.go
rdoxenham, line 166: This is only the case when `apiVip` is specified, right? Because if it's not, we should be defaulting to out of the box functionality, where klipper is still enabled, and it takes the external IP of all of the hosts. Can you confirm @atanasdinov ?

```suggestion
	log.Audit("WARNING: An external IP address for the Ingress Controller (Traefik) must be manually configured in multi-node clusters where you've specified an API VIP.")
```

atanasdinov: Not really. There's no multi node cluster without an apiVIP.

rdoxenham (approved):
LGTM 🚀

PR #149, opened 2024-02-02
Air-gapped helm chart deployment

dbw7:
Although this code is extremely rough, it is extremely close to completely air-gapped helm chart deployment. All that's left is actually writing the helm manifests to combustion after editing them, and fixing a small bug with using `/static/charts` on RKE2.

After that and some refactoring, everything should be good.


pkg/combustion/templates/14-embedded-registry.sh.tpl
```diff
@@ -18,11 +20,67 @@ cat <<- EOF > /etc/systemd/system/eib-embedded-registry.service
...
+  [Unit]
+  Description=Load and Serve Embedded File Server
+  After=network.target
+
+  [Service]
+  Type=simple
+  User=root
+  WorkingDirectory=/opt/hauler
+  ExecStartPre=/bin/sleep 30
+  ExecStart=/usr/local/bin/hauler store serve fileserver -p {{ .FileServerPort }}
```
atanasdinov, line 43: Another implementation idea, what about running this inside Kubernetes? This way you wouldn't have to look for the host IP at all.

PR #150, opened 2024-02-02
Allow SELinux configuration for k3s clusters

atanasdinov:
- The existing [k3s-selinux](https://github.com/k3s-io/k3s-selinux/blob/master/policy/slemicro/k3s.fc) policy does not cover `/opt` installation as a supported context
```
localhost:~ # cat /etc/rancher/k3s/config.yaml
selinux: true
localhost:~ # semodule -l | grep k3s
k3s
localhost:~ # journalctl -u k3s | grep selinux
Feb 02 13:48:27 localhost.localdomain k3s[2408]: time="2024-02-02T13:48:27Z" level=warning msg="SELinux is enabled for k3s but process is not running in context 'container_runtime_t', k3s-selinux policy may need to be applied"
```


rdoxenham:
I think we can close this following #158

PR #151, opened 2024-02-02
Updates to the image definition for 1.0 release - Part 1

jdob:
Includes:
* Moving the two installation-related fields (`installDevice` and `unattended`) under their own section called `isoInstallation`
* Moved the two NTP-related fields (`chronyServers` and `chronyPools`) under their own section called `ntp`, removing the `chrony` prefix in the process

### DIFF
--- a/docs/building-images.md
+++ b/docs/building-images.md
@@ -37,15 +37,17 @@ The operating system configuration section is entirely optional.
 The following describes the possible options for the operating system section:
 ```yaml
 operatingSystem:
-  installDevice: /path/to/disk
-  unattended: false
+  isoInstallation:
+    installDevice: /path/to/disk
+    unattended: false
   time:
     timezone: Europe/London
-    chronyPools:
-      - 1.pool.server.com
-    chronyServers:
-      - 10.0.0.1
-      - 10.0.0.2
+    ntp:
+      pools:
+        - 1.pool.server.com
+      servers:
+        - 10.0.0.1
+        - 10.0.0.2
   proxy:
     httpProxy: http://10.0.0.1:3128
     httpsProxy: http://10.0.0.1:3128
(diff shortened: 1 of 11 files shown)


docs/building-images.md
```diff
@@ -37,15 +37,17 @@ The operating system configuration section is entirely optional.
...
+    installDevice: /path/to/disk
+    unattended: false
   time:
     timezone: Europe/London
-    chronyPools:
-      - 1.pool.server.com
-    chronyServers:
-      - 10.0.0.1
-      - 10.0.0.2
+    ntp:
```
atanasdinov, line 45: Wasn't this supposed to still be `chrony`?

pkg/combustion/time.go
```diff
@@ -39,7 +39,17 @@ func configureTime(ctx *image.Context) ([]string, error) {
 func writeTimeCombustionScript(ctx *image.Context) error {
 	timeScriptFilename := filepath.Join(ctx.CombustionDir, timeScriptName)
 
-	data, err := template.Parse(timeScriptName, timeScript, ctx.ImageDefinition.OperatingSystem.Time)
+	values := struct {
```
atanasdinov, line 42: I don't have anything against this but was just wondering why not pass the object as before?

atanasdinov (approved):

docs/building-images.md
jdob: We decided to go with the protocol instead of a specific implementation.

pkg/combustion/time.go
jdob: Largely just consistency and personal preference. In many other places (at least the ones I wrote), we do a custom struct. I hate the templating syntax and would rather the data shuffling in Go instead of having to access nested variables in the template.

PR #152, opened 2024-02-02
restructured noProxy into a list in the definition

jdob:


atanasdinov (approved):

pkg/combustion/proxy.go
```diff
@@ -39,7 +40,17 @@ func configureProxy(ctx *image.Context) ([]string, error) {
...
 
-	data, err := template.Parse(proxyScriptName, proxyScript, ctx.ImageDefinition.OperatingSystem.Proxy)
+	values := struct {
+		HTTPProxy  string
+		HTTPSProxy string
+		NoProxy    string
+	}{
+		HTTPProxy:  ctx.ImageDefinition.OperatingSystem.Proxy.HTTPProxy,
+		HTTPSProxy: ctx.ImageDefinition.OperatingSystem.Proxy.HTTPSProxy,
+		NoProxy:    strings.Join(ctx.ImageDefinition.OperatingSystem.Proxy.NoProxy, ", "),
```
jdob, line 50: I tried to do this in the template itself; this feels like the template details are bleeding into the code here by having us do formatting like this. But I couldn't find a way to not have a trailing comma and I don't have the time to see if that trailing comma is going to cause the proxy functionality to fail.

PR #153, opened 2024-02-02
removed SSL cert retrieval from suma

jdob:


atanasdinov (approved):
We're missing the docs update. Feel free to merge after including it.

PR #158, opened 2024-02-05
Moving the k3s bin directory to /opt/k3s/bin/

rdoxenham:
In the original version we were moving the k3s binaries to `/opt/k3s/` which doesn't match the behaviour of rke2 (`/opt/rke2/bin/`), so this brings it inline so we can push the latest requirements into k3s-selinux in a consistent way.


atanasdinov (commented):
Can we update the tests at https://github.com/suse-edge/edge-image-builder/blob/main/pkg/combustion/kubernetes_test.go#L309 and https://github.com/suse-edge/edge-image-builder/blob/main/pkg/combustion/kubernetes_test.go#L212?

atanasdinov (approved):

PR #161, opened 2024-02-06
Add documentation for EIB's RPM resolution logic

ipetrov117:
Documentation on how:
1. to configure packages for installation in the EIB image
2. E2E description of how EIB's RPM resolution logic works - from running the EIB container to booting the EIB image

Tested on:
1. SUSE Linux - SLES 15 and openSUSE Tumbleweed
2. Ubuntu 22.04
3. Fedora Linux

PR #162, opened 2024-02-06
Decouple main and base resolver image builds

ipetrov117:
This PR is not a hard-requirement for the 3.0 release, but it will significantly increase the testability of the `resolver` package in the future. 

---
**What is achieved?**
The logic which is responsible for building the image that the resolver will use as base has been decoupled from the `Resolver` struct. 

By doing this we improve the following points:
1. _overall testability of the main `resolver` logic_ - with this change, there are no more shell script executions inside the main `resolver` logic, which will enable us to test the full functionality of the code. Previously this was a problem since we cannot easily mock a shell script execution
2. _support for different base image build logic_ - In the future we might have the need to build an image with different build requirements (e.g. not using a tarball to import the image, but building the image in a different way). This change makes the base resolver image build logic interchangeable.

---

**Note:** I opened this PR as an improvement suggestion, if you feel like this is over-engineered or that it does not offer any value, feel free to decline the PR. If we decide to merge this, some changes would have to be done to the **RPM resolution** doc that comes from #161. But they can be handled in a different documentation PR.


pkg/rpm/resolver/base.go
```diff
@@ -14,109 +14,129 @@ import (
...
 
 const (
-	prepareBaseScriptName = "prepare-base.sh"
-	prepareBaseScriptLog  = "prepare-base.log"
-	baseImageArchiveName  = "sle-micro-base.tar.gz"
-	baseImageRef          = "slemicro"
+	prepareTarballScriptName = "prepare-tarball.sh"
+	prepareTarballScriptLog  = "prepare-tarball.log"
+	tarballName              = "tarball-image.tar.gz"
+	tarballImgRef            = "tarball"
```
atanasdinov, line 20: Can we include more information in the values themselves i.e. what are the tarballs about e.g. `resolver-tarball.log`, `base-resolver-image.log` or something similar? It is completely clear what these are in the context of this package but I'm afraid it'd be ambiguous when users are only looking at the build dir contents.

pkg/rpm/resolver/base.go
```diff
@@ -14,109 +14,129 @@ import (
...
+
+type TarballImageBuilder struct {
+	// dir from where the image builder will work
+	dir string
+	// path to the ISO/RAW file from which a tarball will be created
+	imgPath string
+	// type of the image that will be used as base (either ISO or RAW)
+	imgType string
+	// podman client which to use for image build
+	podman Podman
```
atanasdinov, line 34: Nit: This could be simplified a lot by splitting up the Podman interface. What this type needs is an "image importer" having an `Import()` / `ImportImage()` method. Since the rest of the Podman functionality is not necessary, implementing the whole interface in unit tests will require mocking the other three functions which are never used.

atanasdinov (approved):

PR #167, opened 2024-02-06
[WIP] Add support for nmc configuration via configdrive

hardys:
To enable network configuration in the Metal3 DHCP-less flow, we need to inject nmc and a script which reads the config-drive data, then runs nmc to generate/apply the config.

This draft adds a new os.configdrive definition option, which enables this behavior - open to suggestions on alternative interfaces though.

### DIFF
--- a/pkg/combustion/templates/03-configure-network-configdrive.sh
+++ b/pkg/combustion/templates/03-configure-network-configdrive.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+set -euo pipefail
+
+# Attempt to statically configure a nic in the case where we find a network_data.json
+# In a configuration drive
+
+CONFIG_DRIVE=$(blkid --label config-2)
+if [ -z "${CONFIG_DRIVE}" ]; then
+  echo "No config-2 device found, skipping network configuration"
+  exit 0
+fi
+
+mount -o ro $CONFIG_DRIVE /mnt
+
+NETWORK_DATA_FILE="/mnt/openstack/latest/network_data.json"
+
+if [ ! -f "${NETWORK_DATA_FILE}" ]; then
+  echo "No ${NETWORK_DATA_FILE} found, skipping network configuration"
+  umount /mnt
+  exit 0
+fi
+
+mkdir -p /tmp/nmc/{desired,generated}
+cp ${NETWORK_DATA_FILE} /tmp/nmc/desired/hostname.yaml
+umount /mnt
+
+nmc generate --config-dir /tmp/nmc/desired --output-dir /tmp/nmc/generated
+nmc apply --config-dir /tmp/nmc/generated
+
+systemctl restart NetworkManager
--- a/pkg/image/testdata/full-valid-example.yaml
+++ b/pkg/image/testdata/full-valid-example.yaml
@@ -34,6 +34,7 @@ operatingSystem:
     disable:
       - disable0
   keymap: us
+  configdrive: true
   users:
     - username: alpha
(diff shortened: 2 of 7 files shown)


pkg/image/testdata/full-valid-example.yaml
```diff
@@ -34,6 +34,7 @@ operatingSystem:
     disable:
       - disable0
   keymap: us
+  configdrive: true
```
hardys, line 37: Since this is specific to metal3 perhaps this should be in a new metal3 section?  

Or alternatively it could be made generic by templating the disklabel and path, but I'm not aware of any other use-cases for config-drive usage such as this atm?

jdob: I agree that it doesn't really feel correct here.

I'm not sure a section called `metal3` works either. For example, we just had this discussion of using chrony v. ntp. Part of me is of the mind that we should keep the sections conceptual instead of implementation specific, which would suggest something like `provisioning`. But without having a bigger idea of what we want to configure related to that, it's hard to tell now if the config will be super metal3 specific or if we want to leave room for others.

My gut reaction is that I'd prefer a new section called `network`, since that's what this really applies to (assuming I'm following this correctly, which I may not be). It's not amazing to have a `network` section with a single value, but I could imagine scenarios where we'd add more to a network section in the future.

One minor nit, I'd prefer to use `configDrive` to stay consistent with the camel case elsewhere in the definition.

jdob: Still another option is to not have this be part of EIB at all. The script isn't templated, so it's not taking advantage of any sort of user-provided parameters where EIB would provide value. The `03-configure-network could simply be a file under `custom/scripts` and the nmc binary under `custom/files`. That's not entirely clean in and of itself since we already include nmc in the EIB image (there may be a cleaner way of just telling EIB to include that in the image), but it does feel cleaner than modifying our definition API for a single use case that doesn't really need EIB to do anything complex to it.

hardys: Yeah my original plan was to do this outside of EIB (at least for now), however the only (supportable) method for delivering nmc is currently via EIB, so I went with this instead in an attempt to simplify the release dependencies - for product usage we'd have to provide a fully supported RPM version of EIB which is installable on SLEMicro, then use EIB to install it along with the script included here.

If that's your preferred direction I can look into it, another option would be some flag which just injects nmc, or @atanasdinov and I discussed it and it was mentioned that we could consider always injecting nmc (which may not be ideal for small footprint situations, but it's a small dependency)

pkg/combustion/templates/03-configure-network-configdrive.sh
```diff
@@ -0,0 +1,30 @@
...
+  echo "No ${NETWORK_DATA_FILE} found, skipping network configuration"
+  umount /mnt
+  exit 0
+fi
+
+mkdir -p /tmp/nmc/{desired,generated}
+cp ${NETWORK_DATA_FILE} /tmp/nmc/desired/hostname.yaml
+umount /mnt
+
+nmc generate --config-dir /tmp/nmc/desired --output-dir /tmp/nmc/generated
```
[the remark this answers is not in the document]
hardys: Yes it should, thanks - I've done it that way in the IPA ramdisk but forgot to do the same here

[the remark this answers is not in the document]
hardys: Actually isn't combustion configured to run only before firstboot.service, so we shouldn't have to wrap each individual script?

hardys:
Superseded by #189

PR #173, opened 2024-02-08
Improve Dockerfile layer caching

atanasdinov:
- Reorders builder image layers
- Only includes source code copies
- Utilises `RUN --mount=type=cache` option

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,16 +1,20 @@
 # ----- EIB Builder Image -----
 FROM registry.suse.com/bci/golang:1.21
 
-WORKDIR /src
-COPY . ./
-
 # Dependency uses by line
 # 1. Podman Go library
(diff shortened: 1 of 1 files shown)


rdoxenham (approved):
Code looks good, although I'm not totally sure on all of the benefits of doing this. Tested too. 🚀

jdob:
> Code looks good, although I'm not totally sure on all of the benefits of doing this. Tested too. 🚀

During the dev process, we were seeing long rebuild times that gets really old, really quick. The idea here is to minimize the image build time by moving the more volatile layers later in the process.

jdob:
It'll be hard to really judge if this worked until I'm more actively playing with the code, but this change works and we can revisit if the dev inner loop still needs some work.

jdob (approved):

PR #174, opened 2024-02-08
Download Kubernetes SELinux policies

atanasdinov:
- Automatically downloads the necessary SELinux policies if SELinux is enabled for k3s or RKE2


pkg/kubernetes/selinux.go
```diff
@@ -0,0 +1,82 @@
...
+		rke2PolicyChannel       = "stable"
+		rke2PolicyReleaseNumber = "1"
+
+		k3sPolicyVersion       = "1.4" // TODO: Bump to 1.5 once the release assets are fixed
+		k3sPolicyChannel       = "stable"
+		k3sPolicyReleaseNumber = "1"
+
+		rke2PolicyRPM = "rke2-selinux-%s-%s.slemicro.noarch.rpm"
+		k3sPolicyRPM  = "k3s-selinux-%s-%s.slemicro.noarch.rpm"
+	)
```
rdoxenham, line 63: I understand the approach, but we're going to have to think about version management here. We will have branches of EIB for the respective SUSE Edge version, but it's still something for us to think about.

atanasdinov: You're absolutely correct. We'll need to figure out a way to extract separate configuration values. At this point though, hardcoding these is our best bet to get some kind of automation in right now.

rdoxenham (approved):
I'm fine with landing as it is, I've also tested it successfully, and as you call out we need to fix this for k3s-selinux 1.5 when it's released. Can we add an information note to say that we've detected `selinux: true` and therefore we're downloading the packages? As otherwise it seems a little odd to just immediately start downloading the packages before any other output:

```
rdo@woden:~/work/eib-testing% podman run --rm --privileged -it -v /home/rdo/work/eib-testing/:/eib eib:174 /bin/eib build -config-file eib-config-iso.yaml -config-dir /eib -build-dir /eib/_build
Downloading file: k3s-selinux-1.4-1.slemicro.noarch.rpm 100% |█████████████████████████████████████| (21/21 kB, 3.0 MB/s)
Downloading file: rancher-public.key 100% |███████████████████████████████████████████████████████| (2.4/2.4 kB, 11 MB/s)
Identifier ................... [SUCCESS]
Custom Files ................. [SKIPPED]
Time ......................... [SUCCESS]
Network ...................... [SUCCESS]
Proxy ........................ [SKIPPED]
Resolving package dependencies...
```

pkg/cli/build/build.go
```diff
@@ -66,6 +66,11 @@ func Run(*cli.Context) error {
 		return nil
 	}
 
+	if err = downloadKubernetesSELinuxPolicies(ctx); err != nil {
+		audit.AuditInfof("Configuring Kubernetes failed. %s", checkLogMessage)
```
jdob, line 70: When I added `AuditInfof`, I had an iteration where there was an `AuditErrorf`. I removed it because I didn't have a use for it, but this feels like a use (and probably not the last).

How would you feel about adding in `AuditErrorf` and for now simply chaining it back to `AuditInfof` under the covers? That way if we do decide we want to output error information differently (*cough* in red *cough*), we can simply update that implementation and it'll be consistent across the entire CLI.

pkg/cli/build/build.go
```diff
@@ -199,6 +204,36 @@ func isImageDefinitionValid(ctx *image.Context) bool {
 	return true
 }
 
+func downloadKubernetesSELinuxPolicies(ctx *image.Context) error {
+	if ctx.ImageDefinition.Kubernetes.Version == "" {
+		return nil
+	}
+
+	configPath := filepath.Join(ctx.ImageConfigDir, "kubernetes", "config", "server.yaml")
```
jdob, line 212: This module already imports from `combustion`. Is there a reason you can't have a method in the kubernetes file to return this path? Feels cleaner than having the specific config directory structure leak out into this file.

pkg/cli/build/build.go
```diff
@@ -199,6 +204,36 @@ func isImageDefinitionValid(ctx *image.Context) bool {
...
+	}
+
+	selinuxEnabled, _ := config["selinux"].(bool)
+	if !selinuxEnabled {
+		return nil
+	}
+
+	audit.Audit("SELinux is enabled in the Kubernetes configuration. " +
+		"The necessary RPM packages will be downloaded.")
+	zap.S().Info("SELinux mode requested for Kubernetes")
```
jdob, line 225: If you use `AuditInfo`, you can skip the explicit zap log, it'll do it for you.

pkg/cli/build/build.go
```diff
@@ -199,6 +204,36 @@ func isImageDefinitionValid(ctx *image.Context) bool {
...
+
+	rpmDir := filepath.Join(ctx.ImageConfigDir, "rpms")
+	gpgKeysDir := filepath.Join(rpmDir, "gpg-keys")
```
jdob, line 228: Similar point here as above, I don't love the idea of the directory structure being hardcoded in multiple places. Can we make this a function in the rpm file to keep the details of how the config directory is laid out in one location?

pkg/kubernetes/selinux.go
```diff
@@ -0,0 +1,82 @@
...
+	"strings"
+
+	"github.com/suse-edge/edge-image-builder/pkg/http"
+	"github.com/suse-edge/edge-image-builder/pkg/image"
+	"go.uber.org/zap"
+)
+
+const (
+	rke2SELinuxPolicyReleaseURL = "https://github.com/rancher/rke2-selinux/releases/download/%s/%s"
+	k3sSELinuxPolicyReleaseURL  = "https://github.com/k3s-io/k3s-selinux/releases/download/%s/%s"
```
jdob, line 19: @rdoxenham @hardys Apologies if this is a question that's been answered already, but is pulling this from github instead of something like OBS going to be a problem for us from a release artifacts point of view?

pkg/kubernetes/selinux.go
```diff
@@ -0,0 +1,82 @@
...
+	signingKeyPath := filepath.Join(gpgKeysDir, "rancher-public.key")
+	if err := http.DownloadFile(context.Background(), rancherSigningKeyURL, signingKeyPath, nil); err != nil {
+		return fmt.Errorf("downloading rancher signing key: %w", err)
+	}
+
+	return nil
+}
+
+func newSELinuxPolicy(kubernetesVersion string) selinuxPolicy {
+	const (
```
jdob, line 52: Nit: I understand defining them in here from a scope perspective, but it feels weird to have the download locations and this version information in two different places instead of right next to each other for a simpler overall picture.

pkg/kubernetes/selinux_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+func TestNewSELinuxPolicy_RKE2(t *testing.T) {
+	policy := newSELinuxPolicy("v1.29.0+rke2r1")
+
+	expectedURL := "https://github.com/rancher/rke2-selinux/releases/download/v0.17.stable.1/rke2-selinux-0.17-1.slemicro.noarch.rpm"
+	expectedRPM := "rke2-selinux-0.17-1.slemicro.noarch.rpm"
+
+	assert.Equal(t, expectedURL, policy.downloadURL)
+	assert.Equal(t, expectedRPM, policy.rpmName)
+}
```
jdob, line 27: I could see an argument for a test case that throws in a completely arbitrary version string and shows the URL/RPM are still generated. The rationale, for the types of devs that buy into this sort of thing, is that a test case like this will show that this method does no parsing or validation of the version. It's entirely up to you; I've never particularly bought into the idea of using tests like this.

jdob (commented):
A few nits and a few comments about business logic leaking where it shouldn't (IMO). The only reason I didn't approve it yet is that I want someone to tell us that the download location is acceptable from a productization and maintenance standpoint.

pkg/cli/build/build.go
atanasdinov: I don't think we want to have the error itself in the audits. I'm pretty sure we set on having any Go errors be logged to the file and only redirect the user there since their formatting can vary wildly and may lead to a bad UX. This is why `AuditErrorf` seems like a bad idea to me but I'm open to suggestions.

pkg/kubernetes/selinux.go
atanasdinov: These RPMs don't exist in OBS and we don't really want to maintain a fork.

pkg/kubernetes/selinux.go
atanasdinov: I went back and forth with this one and liked that I can simplify the constant names if they were function scoped instead of being defined on package level i.e. `rke2PolicyVersion` instead of `rke2SELinuxRPMPolicyVersion`.

atanasdinov: I'll move the policy URLs to the function as well.

pkg/cli/build/build.go
atanasdinov: I know but at the same time I didn't want the verbose UX logs in the log file. This is why I've went for "duplication" instead of using `AuditInfo`.

It's not a hard opinion though and I can change if you feel it's better.

pkg/cli/build/build.go
atanasdinov: I definitely see this as an issue but there's just no clean way to do it right now.. The `combustion` package understands the `/kubernetes/config` structure but it's the `kubernetes` package which knows that there's a `server.yaml` in there. I can extract it as suggested to somewhat improve it but the implementation will still bleed into the different packages in a way I don't like at all.

pkg/cli/build/build.go
atanasdinov: This one is even worse. 😅 

It is duplicated in both the `combustion` and `rpm/resolver` packages. I'll think about it.

pkg/cli/build/build.go
jdob: It's not worth speculating too much right now. Let's see what the next few weeks of testing bring up in terms of UX (I suspect we'll need some significant changes across the board) and then this may be a moot point.

pkg/kubernetes/selinux_test.go
jdob: Works for me. Like I said, I've never bought into that school of thought.

pkg/kubernetes/selinux.go
hardys: I discussed this with @atanasdinov - we should consume RPM releases from rke2/k3s instead of the GH releases, these will be signed artefacts with a discoverable version, and there will presumably be a day-2 strategy to upgrade them alongside the other rke2/k3s components which we don't have to own directly.

pkg/cli/build/build.go
```diff
@@ -201,6 +206,49 @@ func isImageDefinitionValid(ctx *image.Context) bool {
 	return true
 }
 
+func appendKubernetesSELinuxRPMs(ctx *image.Context) error {
+	if ctx.ImageDefinition.Kubernetes.Version == "" {
+		return nil
+	}
+
+	configPath := combustion.KubernetesConfigPath(ctx)
+	config, err := kubernetes.ParseKubernetesConfig(configPath)
```
jdob, line 215: This definitely feels better IMO.

jdob (approved):

PR #175, opened 2024-02-08
Download Elemental RPMs

atanasdinov:
- Downloads `elemental-register` and `elemental-system-agent` when the Elemental component is configured

### DIFF
--- a/pkg/cli/build/build.go
+++ b/pkg/cli/build/build.go
@@ -66,6 +66,8 @@ func Run(*cli.Context) error {
 		return nil
 	}
 
+	appendElementalRPMs(ctx)
+
 	if !bootstrapDependencyServices(ctx, args.RootBuildDir) {
 		os.Exit(1)
 	}
@@ -197,6 +199,28 @@ func isImageDefinitionValid(ctx *image.Context) bool {
 	}
 
 	return true
+}
+
+func appendElementalRPMs(ctx *image.Context) {
+	elementalDir := filepath.Join(ctx.ImageConfigDir, "elemental")
+	if _, err := os.Stat(elementalDir); err != nil {
+		if !errors.Is(err, fs.ErrNotExist) {
+			zap.S().Warnf("Looking for 'elemental' dir failed unexpectedly: %s", err)
+		}
+
+		return
+	}
+
+	packageList := ctx.ImageDefinition.OperatingSystem.Packages.PKGList
+	packageList = append(packageList, "elemental-register", "elemental-system-agent")
+
+	// TODO: Use an official repository URL once it's out
+	repositories := ctx.ImageDefinition.OperatingSystem.Packages.AdditionalRepos
+	repositories = append(repositories,
+		image.AddRepo{URL: "https://download.opensuse.org/repositories/isv:/Rancher:/Elemental:/Staging/standard/"})
+
+	ctx.ImageDefinition.OperatingSystem.Packages.PKGList = packageList
+	ctx.ImageDefinition.OperatingSystem.Packages.AdditionalRepos = repositories
 }
 
 // If the image definition requires it, starts the necessary services, displaying appropriate messages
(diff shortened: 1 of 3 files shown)


rdoxenham (changes requested):
I've tested this successfully, but I think we need to do two things-

1. It would be great if we could give some feedback to the user that we're automatically downloading (and resolving for) the elemental RPM's. 
2. We should also make sure in our validation that the user has specified their SCC login, as this seems to execute the package resolver code without the user supplying credentials.

rdoxenham (approved):

jdob:
> It would be great if we could give some feedback to the user that we're automatically downloading (and resolving for) the elemental RPM's.

@rdoxenham I have a feeling that during QA we're going to identify a few of these sorts of UX issues. It's arguable if we're skirting code freeze by calling it a "bug" later, but realistically I don't think we've had enough wide-scale usage to know how well the UX holds up. Are you cool with us landing this and then in early March reevaluating all UX bugs that come in and addressing them then?

pkg/cli/build/build.go
```diff
@@ -199,6 +201,31 @@ func isImageDefinitionValid(ctx *image.Context) bool {
 	return true
 }
 
+func appendElementalRPMs(ctx *image.Context) {
+	elementalDir := filepath.Join(ctx.ImageConfigDir, "elemental")
```
jdob, line 205: Similar to my comments in your other PR, it feels wrong to me that the directory structure information would be hardcoded in multiple locations. Can we make this a call into `elemental.go` that retrieves the directory instead of hardocoding it here?

pkg/cli/build/build.go
```diff
@@ -199,6 +201,31 @@ func isImageDefinitionValid(ctx *image.Context) bool {
...
+	if _, err := os.Stat(elementalDir); err != nil {
+		if !errors.Is(err, fs.ErrNotExist) {
+			zap.S().Warnf("Looking for 'elemental' dir failed unexpectedly: %s", err)
+		}
+
+		return
+	}
+
+	audit.Audit("Elemental registration is configured. The necessary RPM packages will be downloaded.")
+	zap.S().Info("Elemental registration requested")
```
jdob, line 215: `AuditInfo` will do both write to the console and log the message; is there a reason the messages need to be different?

pkg/cli/build/build.go
```diff
@@ -199,6 +201,31 @@ func isImageDefinitionValid(ctx *image.Context) bool {
...
+
+	packageList := ctx.ImageDefinition.OperatingSystem.Packages.PKGList
+	packageList = append(packageList, "elemental-register", "elemental-system-agent")
+
+	// TODO: Use an official repository URL once it's out
+	repositories := ctx.ImageDefinition.OperatingSystem.Packages.AdditionalRepos
+	repositories = append(repositories,
+		image.AddRepo{URL: "https://download.opensuse.org/repositories/isv:/Rancher:/Elemental:/Staging/standard/"})
```
jdob, line 223: Not to keep beating the same drum, but in my head, all of the logic/details around a specific component should be located in a singular location. How would you feel about having this be a constant in elemental.go so that it's the sole owner of all of this sort of elemental-specific data? 

We'll have to readdress when we get a less in-code mechanism for storing all of these values, but I think that further enforces that point. We'll be able to look in each component module and know all of the values we may want to pull out and make overridable.

jdob (approved):
Feel free to merge it if there are reasons against moving the values I suggested.

pkg/cli/build/build.go
atanasdinov: As mentioned on the other PR, not a hard "requirement" but I definitely don't see the point of being verbose / user friendly in the log files. I tend to prefer splitting up what the user sees from what the log files contain even at the cost of duplicating code.

I can change if you favour the `AuditInfo` approach, though.

PR #178, opened 2024-02-11
Helm Chart Deployment First Half

dbw7:
#166 was getting a little too out of control, so this is the first of 2 PRs so that things are split up.


pkg/registry/helm.go
```diff
@@ -0,0 +1,297 @@
...
+	if localHelmSrcDir == "" {
+		return nil, nil
+	}
+
+	helmManifestPaths, err := getLocalManifestPaths(localHelmSrcDir)
+	if err != nil {
+		return nil, fmt.Errorf("getting helm manifest paths: %w", err)
+	}
+
+	var allManifests [][]map[string]any
```
atanasdinov, line 286: Can we reduce this to slice of maps?

pkg/registry/helm.go
```diff
@@ -0,0 +1,297 @@
...
+			}
+		}
+
+		manifests = append(manifests, manifest)
+	}
+
+	return manifests, nil
+}
+
+func UpdateAllManifests(localHelmSrcDir string, chartTarsPath []string) ([][]map[string]any, error) {
```
atanasdinov, line 276: What do we refer to with "All" here? Can we rename this to `UpdateHelmManifests`?

pkg/registry/helm.go
```diff
@@ -0,0 +1,297 @@
...
+			manifests = append(manifests, manifest)
+			continue
+		}
+
+		if spec, ok := manifest["spec"].(map[string]any); ok {
+			if _, ok := spec["chartContent"].(string); ok {
+				manifests = append(manifests, manifest)
+				continue
+			}
+			oldChart := spec["chart"]
```
atanasdinov, line 255: This should instead be `chartName`.

pkg/registry/helm.go
```diff
@@ -0,0 +1,297 @@
...
+			delete(spec, "repo")
+			delete(spec, "chart")
+			for _, chartTar := range chartTarsPaths {
+				if strings.Contains(chartTar, oldChart.(string)) {
```
atanasdinov, line 259: What happens if none of the tarball paths contain the chart name? We should exit with an error.

pkg/registry/helm.go
```diff
@@ -0,0 +1,297 @@
...
+}
+
+func helmPullCommand(repository, chart, version string, destDir string) string {
+	repository = helmRepositoryName(repository, fmt.Sprintf("repo-%s", chart), chart)
+
+	pullCommand := fmt.Sprintf("helm pull %s", repository)
+	if version != "" {
+		pullCommand = fmt.Sprintf("helm pull %s --version %s", repository, version)
+	}
+	if destDir != "" {
```
atanasdinov, line 199: Correct me if I'm wrong but we should never end up in a case where `destDir` is empty. Can we enforce it in the `Generate...` function above?

pkg/registry/helm_test.go
```diff
@@ -0,0 +1,703 @@
...
+				require.NoError(t, err)
+				for _, command := range helmCommands {
+					if !strings.Contains(command, "--set") {
+						assert.Contains(t, test.expectedCommands, command)
+					} else {
+						actualFields := strings.Fields(command)
+						expectedFields := strings.Fields(test.expectedCommands[2])
+
+						if !slices.Contains(actualFields, "--set") {
+							assert.ElementsMatch(t, actualFields, expectedFields)
```
atanasdinov, line 684: This can never happen.

pkg/registry/helm_test.go
```diff
@@ -0,0 +1,703 @@
...
+						} else {
+							setIndex := slices.Index(actualFields, "--set")
+							setValuesIndex := setIndex + 1
+
+							actualSetFields := strings.Split(actualFields[setValuesIndex], ",")
+							expectedSetFields := strings.Split(expectedFields[setValuesIndex], ",")
+
+							assert.ElementsMatch(t, actualSetFields, expectedSetFields)
```
atanasdinov, line 692: This doesn't seem to be sufficient since we only compare the `--set` args and not the whole `template ` command.

pkg/registry/helm_test.go
```diff
@@ -0,0 +1,703 @@
...
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			manifest, err := updateHelmManifest(test.manifestPath, test.chartTarPaths)
+			if test.expectedError == "" && !test.tarTest {
+				require.NoError(t, err)
+				for _, doc := range manifest {
+					kind, kindOk := doc["kind"].(string)
```
atanasdinov, line 280: Let's add a `require.True(t, kindOk)` and simplify the assertions below.

pkg/registry/helm_test.go
```diff
@@ -0,0 +1,703 @@
...
+					spec, specOk := doc["spec"].(map[string]any)
```
atanasdinov, line 281: Should be move inside the check below.

pkg/combustion/registry.go
```diff
@@ -191,14 +194,32 @@ func copyHaulerBinary(ctx *image.Context, haulerBinaryPath string) error {
...
+		k8sType = image.KubernetesDistroK3S
+	}
+
 	values := struct {
-		Port                string
+		RegistryPort        string
 		RegistryDir         string
 		EmbeddedRegistryTar string
+		ChartsDir           string
+		K8sType             string
```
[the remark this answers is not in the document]
dbw7: Thanks for catching!

pkg/registry/helm.go
dbw7: You're right, having the option for `""` was useful for testing but no reason to keep it, will update.

pkg/registry/helm.go
```diff
@@ -252,11 +262,13 @@ func updateHelmManifest(manifestPath string, chartTarsPaths []string) ([]map[str
...
 			}
-			oldChart := spec["chart"]
+			chartName := spec["chart"]
 			delete(spec, "repo")
 			delete(spec, "chart")
+			var noMatchingCharts = true
 			for _, chartTar := range chartTarsPaths {
-				if strings.Contains(chartTar, oldChart.(string)) {
+				if strings.Contains(chartTar, chartName.(string)) {
+					noMatchingCharts = false
```
atanasdinov, line 271: We should also `break` from the loop.

atanasdinov (commented):
I'm fine with this landing but let's not merge it to the main branch until we branch off for the RC candidate we discussed we want to create shortly.

atanasdinov (approved):

PR #181, opened 2024-02-13
Adding support for specifying the desired raw disk size

rdoxenham:
This PR enables the user to specify a desired raw disk image size, this has two key benefits, firstly ensuring that the initial disk image is resized to accommodate the combustion directory accordingly, and secondly to set the size of the partitions for its first boot, noting that if there's any additional space the system will resize to the maximum block size automatically. This change also helps with automated testing and virtual setups, where there's no physical block device that's being used.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,11 +30,10 @@ RUN zypper addrepo https://download.opensuse.org/repositories/isv:SUSE:Edge:Edge
 # 5. Network configurator
 RUN zypper install -y \
     xorriso squashfs  \
-    libguestfs kernel-default e2fsprogs parted gptfdisk btrfsprogs \
+    libguestfs kernel-default e2fsprogs parted gptfdisk btrfsprogs guestfs-tools lvm2 \
     podman \
     createrepo_c \
-    nm-configurator \
-    guestfs-tools lvm2
+    nm-configurator
 
 RUN curl -o hauler-amd64.tar -L https://github.com/rancherfederal/hauler/releases/download/v0.4.2/hauler_0.4.2_linux_amd64.tar.gz && \
     tar -xf hauler-amd64.tar && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,7 +33,8 @@ RUN zypper install -y \
     libguestfs kernel-default e2fsprogs parted gptfdisk btrfsprogs \
     podman \
     createrepo_c \
-    nm-configurator
+    nm-configurator \
+    guestfs-tools lvm2
 
 RUN curl -o hauler-amd64.tar -L https://github.com/rancherfederal/hauler/releases/download/v0.4.2/hauler_0.4.2_linux_amd64.tar.gz && \
     tar -xf hauler-amd64.tar && \
--- a/docs/building-images.md
+++ b/docs/building-images.md
@@ -83,6 +83,12 @@ operatingSystem:
   system rather than prompting user to begin the installation. In combination with `installDevice` can create
   a fully unattended and automated install. Beware of creating boot loops and data loss with these options.
   If left omitted (or set to `false`) the user will still have to choose to install via the GRUB menu.
+* `rawConfiguration` - Optional; configuration in this section only applies to RAW images only.
+  * `diskSize` - Optional; sets the desired raw disk image size. This is important to ensure that your disk
+  image is large enough to accommodate any artifacts that you're embedding. It's advised to set this to slightly
+  smaller than your SDcard size (or block device if writing directly to a disk) and the system will automatically
+  expand at boot time to fill the size of the block device. This is optional, but highly recommended. Specify in
+  integer format with either "M" (Megabyte) or "G" (Gigabyte) as a suffix, e.g. "32G".
 * `time` - Optional; section where the user can provide timezone information and Chronyd configuration.
   * `timezone` - Optional; the timezone in the format of "Region/Locality", e.g. "Europe/London". Full list via `timedatectl list-timezones`.
   * `ntp` - Optional; contains attributes related to configuring NTP
--- a/docs/building-images.md
+++ b/docs/building-images.md
@@ -88,7 +88,7 @@ operatingSystem:
   image is large enough to accommodate any artifacts that you're embedding. It's advised to set this to slightly
   smaller than your SDcard size (or block device if writing directly to a disk) and the system will automatically
   expand at boot time to fill the size of the block device. This is optional, but highly recommended. Specify in
-  integer format with either "M" (Megabyte) or "G" (Gigabyte) as a suffix, e.g. "32G".
+  integer format with either "M" (Megabyte), "G" (Gigabyte), or "T" (Terabyte) as a suffix, e.g. "32G".
 * `time` - Optional; section where the user can provide timezone information and Chronyd configuration.
   * `timezone` - Optional; the timezone in the format of "Region/Locality", e.g. "Europe/London". Full list via `timedatectl list-timezones`.
   * `ntp` - Optional; contains attributes related to configuring NTP
--- a/pkg/build/raw.go
+++ b/pkg/build/raw.go
@@ -86,12 +86,14 @@ func (b *Builder) writeModifyScript(imageFilename string, includeCombustion, ren
 		ConfigureGRUB       string
 		ConfigureCombustion bool
 		RenameFilesystem    bool
+		DiskSize            string
 	}{
 		ImagePath:           imageFilename,
 		CombustionDir:       b.context.CombustionDir,
 		ConfigureGRUB:       grubConfiguration,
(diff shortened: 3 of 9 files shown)


Dockerfile
```diff
@@ -33,7 +33,8 @@ RUN zypper install -y \
     libguestfs kernel-default e2fsprogs parted gptfdisk btrfsprogs \
     podman \
     createrepo_c \
-    nm-configurator
+    nm-configurator \
+    guestfs-tools lvm2
```
atanasdinov, line 37: Nit: We can move this to the second line (`libguestfs...`) since those are the dependencies we've specified we need for RAW image modification.

pkg/build/templates/modify-raw-image.sh.tpl
```diff
@@ -12,6 +12,17 @@ set -euo pipefail
 #
 # Guestfish Command Documentation: https://libguestfs.org/guestfish.1.html
 
+# Resize the raw disk image to accommodate the users desired raw disk image size
+# This is also required if embedding content into /combustion, especially for airgap.
+# Should *only* execute if the user is building a raw disk image.
+{{ if ne .DiskSize "" -}}
+truncate -r {{.ImagePath}} {{.ImagePath}}.expanded
+truncate -s {{.DiskSize}} {{.ImagePath}}.expanded
+virt-resize --expand /dev/sda3 {{.ImagePath}} {{.ImagePath}}.expanded
```
atanasdinov, line 21: Can we include these in the unit test as well?

atanasdinov (approved):

pkg/build/templates/modify-raw-image.sh.tpl
rdoxenham: First is in, second I just added.

Dockerfile
rdoxenham: Done!

atanasdinov (approved):

jdob (changes requested):
This is awesome, thanks for doing this. Only missing piece is to the validation functions to make sure they specify a valid size (greater than zero, ends with G or M with potential case sensitivity if applicable, etc.)

docs/building-images.md
```diff
@@ -83,6 +83,12 @@ operatingSystem:
   system rather than prompting user to begin the installation. In combination with `installDevice` can create
   a fully unattended and automated install. Beware of creating boot loops and data loss with these options.
   If left omitted (or set to `false`) the user will still have to choose to install via the GRUB menu.
+* `rawConfiguration` - Optional; configuration in this section only applies to RAW images only.
```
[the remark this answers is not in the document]
jdob: I see both sides. For me it comes down to how we'd use the terms in a sentence:

"You can set the install device for an ISO."
"You can resize the RAW [image]."

Lemme further make this a headache. What about if we change it to:
`isoConfiguration`
`rawConfiguration`

That ignores my sentence argument, but if we're looking in terms of making definition writing as easy as possible, the consistency might go a long way.

pkg/build/templates/modify-raw-image.sh.tpl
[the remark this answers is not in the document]
rdoxenham: I'll address this by making sure that we only match for >0 and that we have either "M", "G", or "T" following an integer for matching the size.

docs/building-images.md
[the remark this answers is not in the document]
rdoxenham: I agree. Let's go with `isoConfiguration` as then we have consistency. We had talked originally about using `isoInstallation` as a method of describing this section as it's actually installing as part of the boot process as opposed to just booting an image, but I think consistency is important here. I'll follow-up with another PR to address this.

pkg/build/raw.go
```diff
@@ -86,12 +86,14 @@ func (b *Builder) writeModifyScript(imageFilename string, includeCombustion, ren
...
 		ConfigureCombustion bool
 		RenameFilesystem    bool
+		DiskSize            string
 	}{
 		ImagePath:           imageFilename,
 		CombustionDir:       b.context.CombustionDir,
 		ConfigureGRUB:       grubConfiguration,
 		ConfigureCombustion: includeCombustion,
 		RenameFilesystem:    renameFilesystem,
+		DiskSize:            b.context.ImageDefinition.OperatingSystem.RawConfiguration.DiskSize,
```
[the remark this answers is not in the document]
rdoxenham: It may work, but the validation I've added makes sure the user specifies a capitalised M/G/T, that way we know that it will input into the command properly.

pkg/image/validation/os.go
```diff
@@ -217,3 +219,31 @@ func validateUnattended(def *image.Definition) []FailedValidation {
 
 	return failures
 }
+
+func validateRawConfig(def *image.Definition) []FailedValidation {
+	var failures []FailedValidation
+	isValidSize := regexp.MustCompile(`^([1-9]\d+|[1-9])+[MGT]`).MatchString
+
+	if def.Image.ImageType != image.TypeRAW && def.OperatingSystem.RawConfiguration.DiskSize != "" {
```
atanasdinov, line 227: Nit: I'd include an "exit early" block instead of repeating the check:

```
if def.OperatingSystem.RawConfiguration.DiskSize == "" {
    return
}
```

atanasdinov (approved):

docs/building-images.md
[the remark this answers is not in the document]
jdob: I dig it, makes sense to do in a follow up.

docs/building-images.md
```diff
@@ -83,6 +83,12 @@ operatingSystem:
...
+  * `diskSize` - Optional; sets the desired raw disk image size. This is important to ensure that your disk
+  image is large enough to accommodate any artifacts that you're embedding. It's advised to set this to slightly
+  smaller than your SDcard size (or block device if writing directly to a disk) and the system will automatically
+  expand at boot time to fill the size of the block device. This is optional, but highly recommended. Specify in
+  integer format with either "M" (Megabyte) or "G" (Gigabyte) as a suffix, e.g. "32G".
```
jdob, line 91: Nit: The validation allows for "T" as well.

pkg/image/validation/os.go
```diff
@@ -217,3 +219,35 @@ func validateUnattended(def *image.Definition) []FailedValidation {
...
+func validateRawConfig(def *image.Definition) []FailedValidation {
+	var failures []FailedValidation
+	isValidSize := regexp.MustCompile(`^([1-9]\d+|[1-9])+[MGT]`).MatchString
+
+	if def.OperatingSystem.RawConfiguration.DiskSize == "" {
+		return nil
+	}
+
+	if def.Image.ImageType != image.TypeRAW {
+		msg := fmt.Sprintf("The 'rawConfiguration/diskSize' field can only be used when 'image type' is '%s'.", image.TypeRAW)
```
jdob, line 232: These messages have been using the actual name of the attribute to make things easier on the user, so 'imageType' instead of 'image type'.

pkg/image/validation/os.go
```diff
@@ -217,3 +219,35 @@ func validateUnattended(def *image.Definition) []FailedValidation {
...
+
+	if def.OperatingSystem.IsoInstallation.InstallDevice != "" {
+		msg := "You cannot simultaneously configure rawConfiguration and isoInstallation, regardless of image type."
+		failures = append(failures, FailedValidation{
+			UserMessage: msg,
+		})
+	}
+
+	if !isValidSize(def.OperatingSystem.RawConfiguration.DiskSize) {
+		msg := fmt.Sprintf("the 'rawConfiguration/diskSize' field must require an integer followed by an 'M/G/T' suffix to indicate size when 'imageType' is '%s'.", image.TypeRAW)
```
jdob, line 246: "must require" is redundant. Consider simplifying to "... field must be an integer..."

Also, it should be "a M/G/T suffix", not "an". Or even better to avoid this entirely "... followed by a suffix of either 'M', 'G', or 'T'".

(yes, this feels excessive to point out, but it's user-facing messages so I tend to be on the more compulsive side of things)

pkg/image/validation/os_test.go
```diff
@@ -566,3 +571,73 @@ func TestValidateUnattended(t *testing.T) {
...
+			Definition: image.Definition{
+				Image: image.Image{
+					ImageType: image.TypeRAW,
+				},
+				OperatingSystem: image.OperatingSystem{
+					RawConfiguration: image.RawConfiguration{
+						DiskSize: "0G",
+					},
+				},
+			},
```
jdob, line 620: Can you add tests for:
* Negative number
* No number at all but does end with the letter M, G, or T
* Lower case for M, G, or T (no need to do all three, just one)

jdob (changes requested):

jdob (approved):

PR #184, opened 2024-02-14
Remove package list validation

ipetrov117:
Remove the validation for the `packageList` RPM resolution property.

Use-case for this is that we want to be able to install packages on openSUSE images without having to provide a third-party repository or a registration code.


atanasdinov (approved):

rdoxenham (approved):
🚀

PR #185, opened 2024-02-14
added basic schema version validation for 1.0

jdob:
I'm writing up a design doc detailing the rules we're going to follow when it comes to changing the schema version, but in the interest of getting the first RC built, I'm going to do that in a separate PR in case we go back and forth on the phrasing and handling.

### DIFF
--- a/pkg/image/validation/validation.go
+++ b/pkg/image/validation/validation.go
@@ -14,6 +14,14 @@ type validateComponent func(ctx *image.Context) []FailedValidation
 func ValidateDefinition(ctx *image.Context) map[string][]FailedValidation {
 	failures := map[string][]FailedValidation{}
 
+	schemaValidationFailure := validateAPIVersion(ctx)
+	if schemaValidationFailure != nil {
+		// If the schema is invalid, there's no reason to even attempt the rest of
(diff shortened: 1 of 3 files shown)


pkg/image/validation/validation.go
```diff
@@ -14,6 +14,14 @@ type validateComponent func(ctx *image.Context) []FailedValidation
 func ValidateDefinition(ctx *image.Context) map[string][]FailedValidation {
 	failures := map[string][]FailedValidation{}
 
+	schemaValidationFailure := validateAPIVersion(ctx)
+	if schemaValidationFailure != nil {
+		// If the schema is invalid, there's no reason to even attempt the rest of
+		// the validation
+		failures["Definition Schema"] = []FailedValidation{*schemaValidationFailure}
```
atanasdinov, line 21: Can we get a constant for this "component" as well?

atanasdinov (approved):

pkg/image/validation/validation.go
jdob: I went back and forth far too long on this, I should have just asked for a second opinion. Will do.

Issue #72, opened 2024-01-02
EIB CLI to replace `podman run`


rdoxenham:
Rather than using `podman run` to execute EIB, we should create a small CLI tool that wraps this and checks system compatibility. For example:

```
$ image-builder --config-directory /path/to/config-directory
[INFO] Checking system compatibility...
[INFO] Creating temporary directories...
[INFO] Validating configuration file...
[INFO] Validating provided image...
[INFO] Validating networking configuration...
[INFO] Executing Podman build...

(tail logs)
```

jdob:
Would a shell script suffice?

rdoxenham:
@jdob it could... we just need to make it less error prone when running EIB via `podman run`

e-minguez:
What about an alias in the meantime?
```
$ alias eib='podman run --rm -it -v $IMAGE_DIR:/eib eib:dev /bin/eib build'
$ eib --config-file $CONFIG_FILE.yaml --config-dir /eib --build-dir /eib/_build
```

rdoxenham:
@e-minguez I think we need something that wraps the command and simplifies execution. We should also consider checking for pre-requisites like podman, the btrfs packages, and perhaps check cgroup version to run privileged automatically.

e-minguez:
> @e-minguez I think we need something that wraps the command and simplifies execution. We should also consider checking for pre-requisites like podman, the btrfs packages, and perhaps check cgroup version to run privileged automatically.

I agree. My proposal was just a _cheat_ in the meantime 😃

PR #191, opened 2024-02-16
Update README.md

wrkode:
document --privileged flag requirement for Elemental builds

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,14 +54,8 @@ podman run --rm -it \
 ```
 
 **NOTE:**
-When specifying packages for installation or doing elemental builds you need to also pass the [`--privileged`](https://docs.podman.io/en/latest/markdown/podman-run.1.html#privileged) flag as well. For more information on why this is needed, see the [Package resolution design](./docs/design/pkg-resolution.md#running-the-eib-container) documentation.
-```shell
-podman run --rm --privileged -it \
--v $IMAGE_DIR:/eib eib:dev /bin/eib build \
---config-file $CONFIG_FILE.yaml \
---config-dir /eib \
---build-dir /eib/_build
-```
+Build process which involves package resolution must include the [`--privileged`](https://docs.podman.io/en/latest/markdown/podman-run.1.html#privileged) flag.
+Package resolution will be automatically performed when requesting package installation or configuring components which require it (e.g. Elemental, Kubernetes SELinux, etc.)
 
 * `-v` - Used to mount a local directory (in this example, the value of $IMAGE_DIR) into the EIB container at `/eib`.
 * `--config-file` - Specifies which image configuration file to build. The path to this file will be relative to
--- a/README.md
+++ b/README.md
@@ -77,7 +77,6 @@ podman run --rm --privileged -it \
 * `--validate` - If specified, the specified image definition and configuration directory will be checked to ensure
   the build can proceed, however the image will not actually be built.
 
-> **_NOTE:_** When specifying packages for installation, you need to also pass the [`--privileged`](https://docs.podman.io/en/latest/markdown/podman-run.1.html#privileged) option as well. For more information on why this is needed, see the [Package resolution design](./docs/design/pkg-resolution.md#running-the-eib-container) documentation.
 
 ## Testing Images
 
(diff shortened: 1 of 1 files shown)


ipetrov117:
@wrkode, can you also remove the [package installation note](https://github.com/wrkode/edge-image-builder/blob/patch-1/README.md?plain=1#L80). It is obsolete now.

wrkode:
> @wrkode, can you also remove the [package installation note](https://github.com/wrkode/edge-image-builder/blob/patch-1/README.md?plain=1#L80). It is obsolete now.

removed and ready for review

ipetrov117 (approved):
Changes look good.

cc @atanasdinov  for a second set of eyes.

jdob (approved):
Good pickup, thanks :)

jdob (changes requested):
I think we can skip this in favor of what's in #192

wrkode:
> I think we can skip this in favor of what's in #192

We shouldn't skip this. these are two independent documents.
The information for the build must result in a successful execution. this change allows that to occur. I'm ok for a generic README, but must be correct.

wrkode:
> I looked at this before noticing #192. Feels redundant to have in both places, and this bleeds specific implementation stuff into an otherwise generic README.

the documents are independent.

README.md
```diff
@@ -67,8 +77,7 @@ podman run --rm -it \
 * `--validate` - If specified, the specified image definition and configuration directory will be checked to ensure
   the build can proceed, however the image will not actually be built.
 
-> **_NOTE:_** When specifying packages for installation, you need to also pass the [`--privileged`](https://docs.podman.io/en/latest/markdown/podman-run.1.html#privileged) option as well. For more information on why this is needed, see the [Package resolution design](./docs/design/pkg-resolution.md#running-the-eib-container) documentation.
```
atanasdinov, line 70: Please consider rewording this along the lines of:
```
Build process which involves package resolution must include the [`--privileged`](https://docs.podman.io/en/latest/markdown/podman-run.1.html#privileged) flag.
Package resolution will be automatically performed when requesting package installation or configuring components which require it (e.g. Elemental, Kubernetes SELinux, etc.)
```

README.md
```diff
@@ -53,6 +53,16 @@ podman run --rm -it \
 --build-dir /eib/_build
 ```
 
+**NOTE:**
+When specifying packages for installation or doing elemental builds you need to also pass the [`--privileged`](https://docs.podman.io/en/latest/markdown/podman-run.1.html#privileged) flag as well. For more information on why this is needed, see the [Package resolution design](./docs/design/pkg-resolution.md#running-the-eib-container) documentation.
+```shell
+podman run --rm --privileged -it \
```
atanasdinov, line 59: This example introduces redundancy.

atanasdinov (changes requested):
Hi, @wrkode. Thanks for the contribution!

I share @jdob's opinion in that this change would introduce redundancy and that component specific documentation belongs to the [Building images](https://github.com/suse-edge/edge-image-builder/blob/main/docs/building-images.md) guide.

We're in the process of actively building up our documentation as well as looking forward to introducing a simpler way of running EIB in the first place (with https://github.com/suse-edge/edge-image-builder/issues/72).

Until then keeping the main README as clean as possible is in my opinion the best way forward.

If you're keen on contributing this, please consider the suggestions above.

README.md
wrkode: agreed, this seems a sensible comment

wrkode (commented):
looks good

jdob (approved):

atanasdinov:
@wrkode do we want to proceed with this? I see that you've agreed with the suggestions but there's no commit addressing them yet.

wrkode:
> @wrkode do we want to proceed with this? I see that you've agreed with the suggestions but there's no commit addressing them yet.

missed the commit. yes LGTM and committing now

wrkode (commented):
LGTM

wrkode:
@atanasdinov , I've committed the suggest changes

atanasdinov (approved):
Thanks!

PR #195, opened 2024-02-16
Change location of Hauler bin

dbw7:
Instead of mounting `/usr/local/` we instead put in the hauler binary in `/opt/hauler/`


rdoxenham (approved):
This works just fine for me, and I like the fact that we're moving it to `/opt` instead of mounting `/usr/local` during boot. Only problem is that `/opt` is not in the default path, so you can no longer just execute `hauler` but I don't see that as an issue given that this is an embedded registry. Further, systemd is calling it from the correct location. Thus, I'm approving this. Thank you 🚀

atanasdinov (approved):

PR #196, opened 2024-02-17
Adding support for waiting for NTP synchronisation at boot

### LINKED ISSUE #170: K3s install fails with "failed to verify certificate, certificate has expired or is not yet valid"
When installing K3s, the certifcate time of K3s is 1 hour before the actual time. This means that I have to wait for an hour to get valid certificates. I tried setting the Timezone to `Etc/UTC` and to `Etc/Amsterdam` but both of the timezones give the error.

I checked the timezone with `datetimectl` and that is correctly set.

In a previous POC with SLE Micro we had a dependency on the timezone sync and after that install K3s. That worked fine. Not sure how I can solve this or our solution was the correct way of doing this.

image

rdoxenham:
In this PR we add support for waiting for NTP syncronisation before attempting to start Kubernetes (or achieving `multi-user.target`). We introduce a oneshot systemd unit file that has to complete before k3s/rke2-server/rke2-agent start, one that relies on `chrony-wait` completing. By default, `chrony-wait.service` has a timeout of 180s, so regardless, the system wont hang at boot time for longer than 180s. If the NTP source can be contacted and synchronised prior to the 180s, then we can immediately start the Kubernetes services.

Fixes #170


pkg/combustion/templates/11-time-setup.sh.tpl
```diff
@@ -16,3 +16,31 @@ echo "pool {{ . }} iburst" >> /etc/chrony.d/eib-sources.conf
...
+RemainAfterExit=true
+
+[Install]
+WantedBy=multi-user.target
+EOF
+
+systemctl enable chrony-wait
+systemctl enable firstboot-timesync.service
+
+echo "[WARN]: Waiting up to 180s to synchronise system clock with available NTP sources."
```
atanasdinov, line 45: This will log the message as part of the Combustion unit / phase. Isn't it better to include it as a `ExecStartPre` of the firstboot-timsync unit instead?

pkg/combustion/templates/11-time-setup.sh.tpl
```diff
@@ -16,3 +16,31 @@ echo "pool {{ . }} iburst" >> /etc/chrony.d/eib-sources.conf
 {{ range .Servers -}}
 echo "server {{ . }} iburst" >> /etc/chrony.d/eib-sources.conf
 {{ end -}}
+
+{{ if .ForceWait -}}
+cat <<EOF >/etc/systemd/system/firstboot-timesync.service
```
atanasdinov, line 21: Can we include a comment describing the behaviour of `chrony-wait`? New developers wouldn't know why `firstboot-timesync.service` only contains an "echo" call.

pkg/image/validation/os.go
```diff
@@ -28,6 +28,7 @@ func validateOperatingSystem(ctx *image.Context) []FailedValidation {
 	failures = append(failures, validatePackages(&def.OperatingSystem)...)
 	failures = append(failures, validateUnattended(def)...)
 	failures = append(failures, validateRawConfig(def)...)
+	failures = append(failures, validateTimesync(def)...)
```
atanasdinov, line 31: It's better to pass the `OperatingSystem` object instead of the full definition.

pkg/image/validation/os_test.go
```diff
@@ -677,3 +677,68 @@ func TestValidateRawConfiguration(t *testing.T) {
...
+
+func TestValidateTimeSync(t *testing.T) {
+	tests := map[string]struct {
+		Definition             image.Definition
+		ExpectedFailedMessages []string
+	}{
+		`not included`: {
+			Definition: image.Definition{},
+		},
+		`forceWait specified and NTP sources valid`: {
```
atanasdinov, line 689: I'd suggest to split the test case into two:
- one where only pools are defined
- one where only servers are defined

pkg/combustion/templates/11-time-setup.sh.tpl
rdoxenham: @atanasdinov the reason why we have this here is because the boot will pause and ideally we need something on the console to explain what's happening. Because the `firstboot-timesync` unit depends on `chrony-wait` I'm not sure that `ExecStartPre` will run until chrony has synchronised. I can check.

pkg/combustion/templates/11-time-setup.sh.tpl
rdoxenham: Where do you want this comment? The systemd unit is a workaround really, it enables us to force systemd to wait until its dependencies are satisfied before it lets k3s/rke2 start, and allows us to do it without modifying the standard k3s/rke2 systemd unit files.

pkg/image/validation/os.go
rdoxenham: Good catch, will fix.

pkg/image/validation/os_test.go
rdoxenham: OK will do.

pkg/combustion/templates/11-time-setup.sh.tpl
atanasdinov: Yes, I fully understand this (with the help of the PR description).

Ideally, a comment just above this `cat` command saying something along the lines of "A oneshot service used to synchronise NTP by leveraging the default timeout (180 seconds) of chrony-wait" would be more than enough to convey the idea in the code itself.

pkg/combustion/templates/11-time-setup.sh.tpl
rdoxenham: So, you can't use `ExecStartPre` with a `oneshot` service. Unfortunately I think this is as good as we're going to get.

atanasdinov (approved):

mpas:
Tested locally on a VM running on a Europe/Amsterdam Timezone. The pods now start nicely and certificates are valid from the start!

image

Great work!! Thanks!

jdob (approved):

PR #211, opened 2024-02-20
Handling for manifests and charts with no images

dbw7:
This PR adds an additional check to the registry combustion component that proceeds with creating an embedded artifact registry and configuring its script only when there are images present in user specified manifests and helm charts.

### DIFF
--- a/pkg/combustion/registry.go
+++ b/pkg/combustion/registry.go
@@ -69,7 +69,7 @@ func configureRegistry(ctx *image.Context) ([]string, error) {
 
 	if !configured {
 		log.AuditComponentSkipped(registryComponentName)
-		log.Audit("Embedded Artifact Registry skipped: Provided manifests/helm charts contain no images.")
+		zap.S().Info("Skipping embedded artifact registry since the provided manifests/helm charts contain no images")
 		return nil, nil
 	}
 


pkg/combustion/registry.go
```diff
@@ -61,11 +61,17 @@ func configureRegistry(ctx *image.Context) ([]string, error) {
 		return nil, fmt.Errorf("configuring helm: %w", err)
 	}
 
-	if err = configureEmbeddedArtifactRegistry(ctx, helmTemplatePath, helmManifestHolderDir); err != nil {
+	var configured bool
+	if configured, err = configureEmbeddedArtifactRegistry(ctx, helmTemplatePath, helmManifestHolderDir); err != nil {
```
atanasdinov, line 65: It's better to drop the inline if statement and use short form declaration `configured, err := ...`

pkg/combustion/registry.go
```diff
@@ -61,11 +61,17 @@ func configureRegistry(ctx *image.Context) ([]string, error) {
...
 		log.AuditComponentFailed(registryComponentName)
 		return nil, fmt.Errorf("configuring embedded artifact registry: %w", err)
 	}
 
+	if !configured {
+		log.AuditComponentSkipped(registryComponentName)
```
atanasdinov, line 71: Can we include a log statement here as well?

pkg/combustion/registry.go
```diff
@@ -211,7 +217,8 @@ func createRegistryCommand(ctx *image.Context, commandName string, args []string
 func IsEmbeddedArtifactRegistryConfigured(ctx *image.Context) bool {
 	return len(ctx.ImageDefinition.EmbeddedArtifactRegistry.ContainerImages) != 0 ||
 		len(ctx.ImageDefinition.Kubernetes.Manifests.URLs) != 0 ||
-		isComponentConfigured(ctx, filepath.Join(k8sDir, helmDir))
+		isComponentConfigured(ctx, filepath.Join(k8sDir, helmDir)) ||
+		isComponentConfigured(ctx, filepath.Join(k8sDir, k8sManifestsDir))
```
atanasdinov, line 221: This has already been merged. Not sure why it shows as a new change.

pkg/combustion/registry.go
```diff
@@ -61,11 +61,18 @@ func configureRegistry(ctx *image.Context) ([]string, error) {
...
-	if err = configureEmbeddedArtifactRegistry(ctx, helmTemplatePath, helmManifestHolderDir); err != nil {
+	configured, err := configureEmbeddedArtifactRegistry(ctx, helmTemplatePath, helmManifestHolderDir)
+	if err != nil {
 		log.AuditComponentFailed(registryComponentName)
 		return nil, fmt.Errorf("configuring embedded artifact registry: %w", err)
 	}
 
+	if !configured {
+		log.AuditComponentSkipped(registryComponentName)
+		log.Audit("Embedded Artifact Registry skipped: Provided manifests/helm charts contain no images.")
```
atanasdinov, line 72: I meant a log statement to the log file i.e. `zap.S()...` which we can later use. `log.AuditComponentSkipped(registryComponentName)` is more than enough for the user facing side.

atanasdinov (approved):

PR #218, opened 2024-02-21
Revamp registry / helm implementation

atanasdinov:
- Introduce helm package and type for the necessary Helm usages
- Extracts the Helm implementation out of the combustion package
- Splits the dependencies in the registry component
- Simplify container images parsing
- Improve testing


pkg/image/context.go
```diff
@@ -52,4 +54,5 @@ type Context struct {
 	// RPMResolver responsible for resolving rpm/package dependencies
 	RPMResolver    rpmResolver
 	RPMRepoCreator rpmRepoCreator
+	Helm           registry.Helm
```
atanasdinov, line 57: This is definitely not ideal. We have to improve it by introducing a Registry type and / or dropping all interfaces from the context in the first place.

jdob: Agreed. I can see a world where we go back to a Builder object which lets us wire up all of these implementation paths and keep them separate from the working data needed to build an image.

pkg/registry/helm_crd.go
```diff
@@ -0,0 +1,61 @@
+package registry
```
jdob, line 1: You'd know better than me, but using an underscore feels weird here. It's minor, but it separates helm.go from helm_tests.go because this and its test are in between. From what I could find [1], this doesn't seem like it passes their exception cases for underscores. 

It's not something I'd block the PR on, I'm mostly curious to see your rationale.

[1] https://google.github.io/styleguide/go/decisions.html

pkg/helm/helm_test.go
```diff
@@ -0,0 +1,244 @@
...
+		t.Run(test.name, func(t *testing.T) {
+			repoName := repositoryName(test.repoURL, test.chart)
+			assert.Equal(t, test.expectedOutput, repoName)
+		})
+	}
+}
+
+func TestAddRepoCommand(t *testing.T) {
+	var buf bytes.Buffer
+	cmd := addRepoCommand("apache", "oci://registry-1.docker.io/bitnamicharts/apache", &buf)
```
dbw7, line 42: Small nit: using `oci://registry-1.docker.io/bitnamicharts/apache` may be a bit misleading as `helm repo add repo-apache oci://registry-1.docker.io/bitnamicharts/apache` is invalid in practice.

jdob (approved):

dbw7 (approved):
Everything looks good to me, great work!

pkg/registry/helm_crd.go
atanasdinov: It's in fact fairly common to use underscores in filenames for compound words. I'm pretty sure I've seen such usages in all Go projects which I have followed / contributed to.

It's the package names where an underscore is definitely not appropriate. Filenames are generally fine (or at least in my opinion).

atanasdinov: We also already have it in EIB e.g. `pkg/kubernetes/artefact_downloader.go`.

Issue #214, opened 2024-02-21
[RFE] Add support for multiple ssh keys in the parameter file


e-minguez:
Currently only a single ssh key can be added to a particular user but it would be nice to have support for multiple keys instead.

PR #220, opened 2024-02-21
updated OS users to take a list of SSH keys

jdob:
closes: #214


pkg/image/validation/os.go
```diff
@@ -114,8 +114,8 @@ func validateUsers(os *image.OperatingSystem) []FailedValidation {
 			})
 		}
 
-		if user.EncryptedPassword == "" && user.SSHKey == "" {
-			msg := fmt.Sprintf("User '%s' must have either a password or SSH key.", user.Username)
+		if user.EncryptedPassword == "" && user.SSHKeys == nil {
```
atanasdinov, line 117:
```suggestion
		if user.EncryptedPassword == "" && len(user.SSHKeys) == 0 {
```

atanasdinov (approved):

PR #223, opened 2024-02-22
Added ability to define operating system groups

jdob:
This doesn't include the changes to be able to assign users to groups, I'll add that in the next PR. However, I did already rename the user combustion script file to make sure it runs after groups, which will be needed once the the ability to assign users to groups is added.


e-minguez:
Not sure if we want to specify group IDs as well (same for users).
It would be useful for some users that want to have certain IDs instead of autogenerated

jdob:
I left room for that option in the future by making the definition not just a list, but a keyed list (not sure that makes sense, but its why its "name:" instead of just a list of values). I figured that's one more moving part past code freeze and we can add it later if we get the request.

atanasdinov (commented):
Please update the RELEASE_NOTES file too.

PR #224, opened 2024-02-22
added user group assignment

jdob:
I explicitly removed the `-m` flag from useradd; I think the default should be no user home directory. After this, I'll add a flag to user creation to allow that option to be set.

### DIFF
--- a/docs/building-images.md
+++ b/docs/building-images.md
@@ -68,8 +68,14 @@ operatingSystem:
     sshKeys:
       - user1Key1
       - user1Key2
+    primaryGroup: groupPrimary
+    secondaryGroups:
+      - group1
+      - group2
   - username: user2
     encryptedPassword: 456
+    secondaryGroups:
+      - group3
   - username: user3
     sshKeys:
       - user3Key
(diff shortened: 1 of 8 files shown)


e-minguez:
I guess we can take some inspiration from the ignition spec https://coreos.github.io/ignition/configuration-v3_4/ as it is somehow the industry standard.
There are too many fields for users to be implemented right now but I guess groups and home creation are a good starting point.

docs/building-images.md
```diff
@@ -128,6 +134,11 @@ operatingSystem:
   * `encryptedPassword` - Optional; Encrypted password to set for the use (for example, using `openssl passwd -6 $PASSWORD`
     to generate the value for this field).
   * `sshKeys` - Optional; List of public SSH keys to configure for the user.
+  * `primaryGroup` - Optional; If specified, the user will be configured with this as the primary group. The group
```
e-minguez, line 137: If it is not specified, what would be the behavior? I will expect a group with the same name as the user would be created (same as `useradd` would do if not specified) but I guess it would be nice to explicitly document it.

dbw7 (approved):
Looks good to me!

docs/building-images.md
jdob: It ends up as the default for what the OS does, which is `users` in SLE Micro's case. But good call, I should explicitly mention that.

docs/building-images.md
```diff
@@ -128,6 +134,11 @@ operatingSystem:
...
+    must already exist, either as a default group or one defined in the `groups` field. May not be specified if the
+    user is "root".
```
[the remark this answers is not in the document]
jdob: On second thought, I'll update the docs to remove this. I can't see a reason _why_ someone would want to, but I also don't necessarily believe it's our place to stop them from doing something like that.

PR #225, opened 2024-02-22
Install Kubernetes VIP manifest as a Helm chart

atanasdinov:
- Creates a plugin for integrating Helm charts required by the different Combustion components
- Uses said plugin to process the Kubernetes VIP manifest


dbw7 (approved):
Looks good to me, great work!

pkg/combustion/templates/k8s-vip.yaml.tpl
```diff
@@ -27,7 +27,7 @@ metadata:
   name: endpoint-copier-operator
   namespace: endpoint-copier-operator
 spec:
-  repo: https://suse-edge.github.io/endpoint-copier-operator
+  repo: https://suse-edge.github.io/charts
```
jdob, line 30: This is interesting when we consider production v. dev. Last I heard (and I may be wrong), our charts will be pulled from *BS. But I also see the need to be able to pull from this sort of development branch too. 

You might already have looked into this and it's a non-issue. But if we're going to have a different location for our shipped charts, can you create an issue so we remember to go back and update this to pull from there?

pkg/combustion/templates/k8s-vip.yaml.tpl
```diff
@@ -27,7 +27,7 @@ metadata:
...
   chart: endpoint-copier-operator
```
jdob, line 31: This may be my lack of Helm knowledge, but is there a way to specify the version here? Feels fragile to not have it pinned for a production release.

jdob (approved):
Admittedly, I don't 100% know how the helm stuff works internally, but from a code perspective nothing looks wrong. A few comments in there about our production readiness around the endpoint copier operator, but if that's a non-issue and I'm missing something, go ahead and merge it.

Issue #197, opened 2024-02-19
Consider using ENTRYPOINT instead of CMD


hardys:
Currently eib container usage is like:

```
podman run -v $IMAGE_DIR:/eib eib:dev /bin/eib build ...
```

This requires knowledge of exactly where the eib binary is installed in the image, and in the case of the RPM based [OBS image](https://build.opensuse.org/package/show/isv:SUSE:Edge:EdgeImageBuilder/edge-image-builder-100rc0-image) it ends up in `/usr/bin/eib` instead.

We could change that in the OBS build so the paths align, but I'm wondering is instead we should be using ENTRYPOINT in both Dockerfiles instead, then the usage can be like this, which may be preferable unless we expect to support multiple entrypoints to the container?

```
podman run -v $IMAGE_DIR:/eib <EIB IMAGE> build ...
```

jdob:
Potentially related to #72, though I think even with the launcher this is still a good idea to consider.

PR #247, opened 2024-02-28
197 - Changed Dockerfile to use ENTRYPOINT instead of CMD

jdob:
closes: #197 

I tested it and it works like Steve suggested, that we don't need to indicate `/bin/eib` in the podman run line. `ENTRYPOINT` makes more sense here since the only argument it includes is the location of the executable, which isn't something that changes.


dbw7 (approved):
Changes make sense and look good to me.

atanasdinov (approved):

rdoxenham (approved):
Very much in favour of simplifying this! Thanks!

PR #248, opened 2024-02-28
consolidate zypper repo add and install layers

jdob:
Docker layers are only rebuilt if necessary, starting at the lowest one with a change. With our previous division, adding the SUSE Edge repo and calling `refresh` was in its own layer separate from the package install calls. While uncommon, this leaves the chance to use stale repo metadata from a cached layer. These two operations should be in the same layer since they are used in conjunction with each other to make sure the metadata and actual RPMs available in the repo match.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,9 +19,6 @@ RUN --mount=type=cache,id=gomod,target=/go/pkg/mod \
 # ----- Deliverable Image -----
 FROM opensuse/leap:15.5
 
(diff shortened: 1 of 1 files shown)


e-minguez (approved):

atanasdinov (approved):
I think this will still not completely fix the issue because we'd read from the cache layer even when a new version of an RPM is available.

jdob:
> I think this will still not completely fix the issue because we'd read from the cache layer even when a new version of an RPM is available.

Agreed. If the RPM gets updated, we won't automatically get it. But we shouldn't get any issues of the repo metadata referring to an old version that's deleted.

PR #257, opened 2024-03-01
Added verbose output for scripts being executed

e-minguez:
image

### DIFF
--- a/pkg/combustion/script.go
+++ b/pkg/combustion/script.go
@@ -36,6 +36,11 @@ func assembleScript(scripts []string, networkScript string) (string, error) {
 
 	// Add a call to each script that was added to the combustion directory
 	for _, filename := range scripts {
+		message := fmt.Sprintf("echo \"Running %s\"\n", filename)
+		_, err = b.WriteString(message)
+		if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/combustion/script.go
```diff
@@ -46,5 +46,5 @@ func assembleScript(scripts []string, networkScript string) (string, error) {
 }
 
 func scriptExecutor(name string) string {
-	return fmt.Sprintf("./%s\n", name)
+	return fmt.Sprintf("echo \"Running %s\"\n./%s\n", name, name)
```
atanasdinov, line 49: I'd suggest to revert this and instead add the following to the for loop above:

```
		message := fmt.Sprintf("echo \"Running %s\"\n", filename)
		_, err = b.WriteString(message)
		if err != nil {
			return "", fmt.Errorf("writing script caller message: %w", err)
		}
```

atanasdinov (approved):

jdob:
Please add a note under RELEASE_NOTES talking about this change.

e-minguez:
> Please add a note under RELEASE_NOTES talking about this change.

Done. PTAL

jdob (approved):

rdoxenham (approved):

PR #274, opened 2024-03-06
Helm Chart Definition

dbw7:


atanasdinov (commented):
A follow-up will adjust the Kubernetes VIP chart requirements to the new implementation. Only then would the leftovers be deleted.

atanasdinov (commented):
Looks good! Minor things to address and I'm happy to approve.

pkg/image/validation/kubernetes.go
```diff
@@ -134,3 +140,82 @@ func validateManifestURLs(k8s *image.Kubernetes) []FailedValidation {
...
+func validateHelmChartValues(valuesFile string, imageConfigDir string) string {
+	if valuesFile == "" {
+		return ""
+	}
+
+	if filepath.Ext(valuesFile) != ".yaml" && filepath.Ext(valuesFile) != ".yml" {
+		return "Helm Chart 'valuesFile' field must be the name of a valid yaml file ending in '.yaml' or '.yml'."
+	}
+
+	valuesFilePath := filepath.Join(imageConfigDir, "values", valuesFile)
```
atanasdinov, line 209: This should check for the file existing under `/helm/values`. Can we please export both of these directories from the combustion package and use them here?

atanasdinov (approved):

PR #285, opened 2024-03-08
corrected go version

jdob:
Not entirely sure what happened, but the project just stopped working on my Mac overnight with the following:

```
$ go mod tidy
go: downloading go1.22 (darwin/arm64)
go: download go1.22 for darwin/arm64: toolchain not available
```

I looked around and found [1] which talks about the issue. The fix works and I don't _think_ there are any other implications in changing this.

[1] https://github.com/golang/go/issues/65568

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/suse-edge/edge-image-builder
 
-go 1.22
+go 1.22.0
 
 require (
 	// version should match buildah version in the


atanasdinov (approved):

jdob:
Closed out; while this probably wouldn't cause an issue, my local go was behind. When I updated that it worked fine as is, and it's late in the dev cycle to dork with any of this if we don't need to.

PR #287, opened 2024-03-08
initial attempt at an examples directory

jdob:
I don't think this is the final form it will take once we get to more complex examples, but I would rather start small and iterate instead of trying to boil the ocean on the first try. The important part is that we don't have a directory of a bunch of files with no description around them outside of the filename, since that can be become pretty useless quickly. The goal was to describe in text what the definition does so users can find what they need without needing to be experts in our schema.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,0 +1,114 @@
+# Image Definition Examples
+
+This directory provides a number of example image definition files that can be used
+directly with Edge Image Builder. They are divided across two directories depending on the
+resulting image type (ISO or RAW image).
+
+Unless otherwise specified, these definitions may be used as-is with one caveat. The base
+image in your image configuration directory must match the name specified in the `baseImage` field
+of the definition. This can be achieved either through renaming your local image or changing this
+field in the definition itself.
+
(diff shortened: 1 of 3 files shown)


examples/README.md
```diff
@@ -0,0 +1,114 @@
...
+  --definition-file ./definitions/iso/basic.yaml
+```
+
+---
+
+## Simple Examples
+
+### `iso/basic.yaml`
+
+| Option       | Default Value                                                           |
```
atanasdinov, line 83: Why do we use the word `Default` here?

atanasdinov (approved):

examples/README.md
jdob: Honestly, we don't need either, but there's no way of doing a header-less table in markdown. What you're seeing is one man's struggle to name a column that didn't actually need a name.

jdob: Actually wait, I kinda lied. "Default" was only because I expect the user may want to change these, particularly the base image name. But like I said, it wasn't a critical decision on my part.

PR #288, opened 2024-03-08
New Helm Definition

dbw7:


pkg/helm/helm.go
```diff
@@ -8,6 +8,8 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/suse-edge/edge-image-builder/pkg/image"
+
```
atanasdinov, line 12: Nit: This empty line can be removed to keep the list clean.

pkg/helm/helm_test.go
```diff
@@ -4,48 +4,58 @@ import (
 	"bytes"
 	"testing"
 
+	"github.com/suse-edge/edge-image-builder/pkg/image"
+
```
atanasdinov, line 8: Nit: This empty line can be removed to keep the list clean.

pkg/image/validation/kubernetes.go
```diff
@@ -143,69 +142,115 @@ func validateManifestURLs(k8s *image.Kubernetes) []FailedValidation {
...
-			failures = append(failures, FailedValidation{
-				UserMessage: "Helm Chart 'repo' field must be defined.",
-			})
-		} else if !strings.HasPrefix(chart.Repo, "http") && !strings.HasPrefix(chart.Repo, "oci://") {
-			failures = append(failures, FailedValidation{
-				UserMessage: "Helm Chart 'repo' field must begin with either 'oci://', 'http://', or 'https://'.",
-			})
-		}
+		seenHelmCharts[chart.Name] = true
+		seenHelmRepos[chart.RepositoryName] = true
```
atanasdinov, line 167: I don't think this will work as expected since the `seenHelmCharts` is only appended once per iteration. Even if it's _fine_ I'd still suggest to extract the duplication catch out of this loop for simplicity.

pkg/image/validation/kubernetes_test.go
```diff
@@ -68,18 +76,27 @@ func TestValidateKubernetes(t *testing.T) {
...
+					Charts: []image.HelmChart{
+						{
+							Name:           "",
+							RepositoryName: "apache-repo",
+							Version:        "10.7.0",
+						},
+					},
+					Repositories: []image.HelmRepository{
+						{
+							Name: "invalid-repo",
```
atanasdinov, line 89: Nit: It's slightly better to use a valid value here e.g. `apache-repo` and another value for `RepositoryName` above e.g. `another-apache-repo`.

pkg/image/validation/kubernetes.go
```diff
@@ -143,69 +142,115 @@ func validateManifestURLs(k8s *image.Kubernetes) []FailedValidation {
...
 	var failures []FailedValidation
 
-	if len(k8s.HelmCharts) == 0 {
+	if len(k8s.Helm.Charts) == 0 {
+		return failures
+	}
+
+	if len(k8s.Helm.Repositories) == 0 {
+		failures = append(failures, FailedValidation{
+			UserMessage: "Helm Charts defined with no Helm Repository defined.",
```
atanasdinov, line 154: What do you think about using lower case for all `Chart`, `Charts`, `Repository` occurrences?

pkg/registry/helm.go
```diff
@@ -14,12 +14,14 @@ type HelmChart struct {
...
-func HelmCharts(helmCharts []image.HelmChart, valuesDir, buildDir, kubeVersion string, helm image.Helm) ([]*HelmChart, error) {
+func HelmCharts(helm *image.Helm, valuesDir, buildDir, kubeVersion string, helmClient image.HelmClient) ([]*HelmChart, error) {
 	var charts []*HelmChart
+	chartRepoMap := mapChartRepos(helm)
 
-	for _, helmChart := range helmCharts {
+	for _, helmChart := range helm.Charts {
 		c := helmChart
-		chart, err := handleChart(&c, valuesDir, buildDir, kubeVersion, helm)
+		r := chartRepoMap[c.RepositoryName]
```
atanasdinov, line 23: We must ensure that a value is there:
```suggestion
		r, ok := chartRepoMap[c.RepositoryName]
		if !ok {
			return nil, fmt.Errorf("repository not found for chart %s", c.Name)
		}
```

atanasdinov (changes requested):
My only concern is the missing check when reading values from the repository map.

atanasdinov (commented):
Release notes are not updated.

pkg/image/validation/kubernetes.go
```diff
@@ -143,80 +142,126 @@ func validateManifestURLs(k8s *image.Kubernetes) []FailedValidation {
...
 	_, err := os.Stat(valuesFilePath)
 	if err != nil {
 		if errors.Is(err, os.ErrNotExist) {
-			return fmt.Sprintf("Helm Chart Values File '%s' could not be found at '%s'.", valuesFile, valuesFilePath)
+			return fmt.Sprintf("Helm chart values file '%s' could not be found at '%s'.", valuesFile, valuesFilePath)
 		}
 
 		zap.S().Errorf("values file '%s' could not be read: %s", valuesFile, err)
-		return fmt.Sprintf("Helm Chart Values File '%s' could not be read.", valuesFile)
+		return fmt.Sprintf("Helm chart values File '%s' could not be read.", valuesFile)
```
atanasdinov, line 264:
```suggestion
		return fmt.Sprintf("Helm chart values file '%s' could not be read.", valuesFile)
```

pkg/image/validation/kubernetes.go
```diff
@@ -143,80 +142,137 @@ func validateManifestURLs(k8s *image.Kubernetes) []FailedValidation {
...
-	seenHelmCharts := make(map[string]bool)
-	for _, chart := range k8s.HelmCharts {
-		if chart.Name == "" {
-			failures = append(failures, FailedValidation{
-				UserMessage: "Helm Chart 'name' field must be defined.",
-			})
-		}
+	if len(k8s.Helm.Repositories) == 0 {
+		failures = append(failures, FailedValidation{
+			UserMessage: "Helm charts defined with no Helm Repository defined.",
```
atanasdinov, line 154:
```suggestion
			UserMessage: "Helm charts defined with no Helm repositories defined.",
```

atanasdinov (approved):

atanasdinov (approved):

PR #290, opened 2024-03-11
New Helm chart definition docs

dbw7:


docs/building-images.md
```diff
@@ -198,6 +218,18 @@ kubernetes:
 * `manifests` - Optional; manifests to apply to the cluster.
   Can be used separately or in combination with the [configuration directory](#kubernetes-1).
   * `urls` - Optional; list of HTTP(s) URLs to download the manifests from
+* `helm` - Optional; Helm charts to be deployed to the cluster.
+  * `charts` - Optional; Defines a list of Helm charts and configuration for each Helm chart.
+    * `name` - Required; This must match the name of the actual Helm chart.
+    * `repositoryName` - Required; This is the name of the corresponding `name` for a repository/registry specified within `repositories` that contains this Helm chart.
+    * `version` - Required; The version of the Helm chart to be deployed.
+    * `installationNamespace` - Optional; The namespace where the Helm installation is executed. The default is `kube-system`.
```
atanasdinov, line 226: This doesn't seem correct.
```suggestion
    * `installationNamespace` - Optional; The namespace where the Helm installation is executed. The default is `default`.
```

docs/building-images.md
```diff
@@ -198,6 +218,18 @@ kubernetes:
...
+    * `targetNamespace` - Optional; The namespace where the Helm chart will be deployed. The default is `default`.
+    * `createNamespace` - Optional; If `true` the `targetNamespace` will be created, if `false` it assumes the `targetNamespace` already exists.
+    * `valuesFile` - Optional; The name of the [Helm values files](https://helm.sh/docs/chart_template_guide/values_files/), placed under `kubernetes/helm/values`, for the specified chart. e.g. the input for `kubernetes/helm/values/metallb-values.yaml` is  `metallb-values.yaml`.
```
atanasdinov, line 229:
```suggestion
    * `valuesFile` - Optional; The name of the [Helm values files](https://helm.sh/docs/chart_template_guide/values_files/), placed under `kubernetes/helm/values`, for the specified chart e.g. the input for `kubernetes/helm/values/metallb-values.yaml` is  `metallb-values.yaml`.
```

atanasdinov (approved):

docs/building-images.md
```diff
@@ -198,6 +218,18 @@ kubernetes:
...
   Can be used separately or in combination with the [configuration directory](#kubernetes-1).
   * `urls` - Optional; list of HTTP(s) URLs to download the manifests from
+* `helm` - Optional; Helm charts to be deployed to the cluster.
+  * `charts` - Optional; Defines a list of Helm charts and configuration for each Helm chart.
+    * `name` - Required; This must match the name of the actual Helm chart.
+    * `repositoryName` - Required; This is the name of the corresponding `name` for a repository/registry specified within `repositories` that contains this Helm chart.
+    * `version` - Required; The version of the Helm chart to be deployed.
+    * `installationNamespace` - Optional; The namespace where the Helm installation is executed. The default is `default`.
+    * `targetNamespace` - Optional; The namespace where the Helm chart will be deployed. The default is `default`.
+    * `createNamespace` - Optional; If `true` the `targetNamespace` will be created, if `false` it assumes the `targetNamespace` already exists.
```
jdob, line 228: What happens if it assumes it exists but it doesn't? I'm guessing runtime error when the node is booted but might make sense to explicitly mention.

docs/building-images.md
```diff
@@ -198,6 +218,18 @@ kubernetes:
...
+    * `valuesFile` - Optional; The name of the [Helm values files](https://helm.sh/docs/chart_template_guide/values_files/), placed under `kubernetes/helm/values`, for the specified chart e.g. the input for `kubernetes/helm/values/metallb-values.yaml` is  `metallb-values.yaml`.
```
jdob, line 229: Nit: "Helm values files" should be "Helm values file".

docs/building-images.md
```diff
@@ -286,9 +318,8 @@ The following sections further describe optional directories that may be include
   * `manifests` - Contains locally provided manifests which will be applied to the cluster. Can be used separately or
     in combination with the manifests section in the definition file. All files in this directory will be parsed and
     the container images that they reference will be downloaded and served in an embedded artefact registry.
-  * `helm` - Contains locally provided ([K3s](https://docs.k3s.io/helm) / [RKE2](https://docs.rke2.io/helm)) Helm Custom Resources
-    which will be applied to the cluster as Helm charts. All files in this directory will be parsed and
-    the container images that they reference will be downloaded and served in an embedded artefact registry.
+  * `helm`
+    * `values` Contains [Helm values files](https://helm.sh/docs/chart_template_guide/values_files/). Each Helm chart that requires specified values must have its own values file.
```
jdob, line 322: Is the second sentence technically correct? It's unlikely, but possible that a single values file would apply to multiple charts, in which case would they need to make a copy or could it be reused?

docs/building-images.md
```diff
@@ -198,6 +218,18 @@ kubernetes:
 * `manifests` - Optional; manifests to apply to the cluster.
   Can be used separately or in combination with the [configuration directory](#kubernetes-1).
   * `urls` - Optional; list of HTTP(s) URLs to download the manifests from
+* `helm` - Optional; Helm charts to be deployed to the cluster.
+  * `charts` - Optional; Defines a list of Helm charts and configuration for each Helm chart.
```
jdob, line 222: Is the plan that this is actually optional? I feel like validation would see if the section was created but no charts defined, unless we explicitly want to allow repositories to be defined without a chart installed.

jdob (approved):

docs/building-images.md
dbw7: 👍 Will reword

docs/building-images.md
dbw7: Will add that 👍

PR #297, opened 2024-03-12
Helm Insecure Arg Handling

dbw7:


pkg/helm/helm.go
```diff
@@ -170,6 +178,13 @@ func pullCommand(chart string, repo *image.HelmRepository, version, destDir stri
 		args = append(args, "--destination", destDir)
 	}
 
+	if repo.SkipTLSVerify {
+		args = append(args, "--insecure-skip-tls-verify")
+	}
+	if repo.PlainHTTP {
+		args = append(args, "--plain-http")
+	}
```
atanasdinov, line 186: Aren't these two mutually exclusive? `else if` should be more appropriate to convey it.

pkg/helm/helm.go
```diff
@@ -114,6 +118,10 @@ func registryLoginCommand(host string, repo *image.HelmRepository, output io.Wri
 		args = append(args, "--username", repo.Authentication.Username, "--password", repo.Authentication.Password)
 	}
 
+	if repo.SkipTLSVerify || repo.PlainHTTP {
```
atanasdinov, line 121: Is a `PlainHTTP` check appropriate here? Official docs state:
```
      --insecure           allow connections to TLS registry without certs
```

pkg/helm/helm_test.go
```diff
@@ -127,6 +174,54 @@ func TestRegistryLoginCommand(t *testing.T) {
...
+			name: "Valid registry with auth and plain HTTP",
+			host: "registry-1.docker.io",
+			repo: &image.HelmRepository{
+				Name: "apache-repo",
+				URL:  "oci://registry-1.docker.io/bitnamicharts/apache",
+				Authentication: image.HelmAuthentication{
+					Username: "user",
+					Password: "pass",
+				},
+				SkipTLSVerify: true,
```
atanasdinov, line 211: This is duplicate of the previous test case.

pkg/helm/helm_test.go
```diff
@@ -213,6 +308,80 @@ func TestPullCommand(t *testing.T) {
...
+					Username: "user",
+					Password: "pass",
+				},
+				PlainHTTP: true,
+			},
+			expectedArgs: []string{
+				"helm",
+				"pull",
+				"oci://registry-1.docker.io/bitnamicharts/apache",
+				"--plain-http",
```
atanasdinov, line 382: Pardon my ignorance - does this even make sense? Isn't the OCI protocol completely separate?

pkg/image/validation/kubernetes.go
```diff
@@ -253,6 +276,36 @@ func validateRepo(repo *image.HelmRepository, seenHelmRepos map[string]bool) []F
...
+func validateHelmRepoArgs(repo *image.HelmRepository) []FailedValidation {
+	var failures []FailedValidation
+
+	if repo.SkipTLSVerify && repo.PlainHTTP {
+		failures = append(failures, FailedValidation{
+			UserMessage: fmt.Sprintf("Helm repository 'plainHTTP' and 'skipTLSVerify' fields for %q cannot both be true.", repo.Name),
+		})
+	}
+
+	if strings.HasPrefix(repo.URL, "http://") && !repo.PlainHTTP {
```
atanasdinov, line 288: This and the checks below seem quite complex. I'd generally suggest to `url.Parse()`, use a switch statement on the `Scheme` field. Feel free to ignore the suggestion if it is not applicable.

pkg/helm/helm_test.go
dbw7: I can’t say I fully understand it just yet but from what I understand at the moment is that it still uses HTTP(s) under the hood:
```
curl -u "user:root" "http://192.168.1.185:5000/v2/_catalog"
{"repositories":["apache/apache","rancher/rancher"]}
```

```
08:07:10 danny@ubuntu ~ → helm registry login http://192.168.1.185:5000
Username: user
Password:
INFO[0002] Error logging in to endpoint, trying next endpoint  error="Get \"https://192.168.1.185:5000/v2/\": http: server gave HTTP response to HTTPS client"
Error: Get "https://192.168.1.185:5000/v2/": http: server gave HTTP response to HTTPS client

08:07:27 danny@ubuntu ~ → helm registry login http://192.168.1.185:5000 --insecure
Username: user
Password:
WARN[0001] insecure registry http://192.168.1.185:5000 should not contain 'http://' and 'http://' has been removed from the insecure registry config
INFO[0001] Error logging in to endpoint, trying next endpoint  error="Get \"https://192.168.1.185:5000/v2/\": http: server gave HTTP response to HTTPS client"
Login Succeeded
```

pkg/helm/helm_test.go
dbw7: Good catch, meant for that to be `plainHTTP: true`.

pkg/image/validation/kubernetes.go
dbw7: I don't think a switch statement would work here but I did switch over to using `url.Parse()`.

pkg/image/validation/kubernetes.go
```diff
@@ -218,6 +222,25 @@ func validateChart(chart *image.HelmChart, imageConfigDir string) []FailedValida
 func validateRepo(repo *image.HelmRepository, seenHelmRepos map[string]bool) []FailedValidation {
 	var failures []FailedValidation
 
+	parsedURL, err := url.Parse(repo.URL)
+	if err != nil {
+		zap.S().Errorf("url '%s' could not be parsed: %s", repo.URL, err)
```
atanasdinov, line 227: We need to return here - `parsedURL` is nil.

PR #299, opened 2024-03-12
Add note to Kubernetes documentation for RAW images

ipetrov117:
* Adds an example for the `rawConfiguration` property
* Adds a note regarding building RAW images with Kubernetes configuration

### DIFF
--- a/docs/building-images.md
+++ b/docs/building-images.md
@@ -165,7 +165,7 @@ The Kubernetes configuration section is another entirely optional one.
 It contains all necessary settings to configure and bootstrap a Kubernetes cluster.
 The supported Kubernetes distributions are K3s and RKE2.
 
-> **_NOTE:_** In addition to the below configuration, if you are building a `raw` image, you must manually specify its disk size. This is done through the [`rawConfiguration`](#operating-system) property.
+> **_NOTE:_**  In addition to the configuration below, if you are building a `raw` image, you must manually specify its disk size. The disk size specification is needed in order to ensure that the `raw` image has enough space to host the Kubernetes tarball resources that EIB attempts to copy ontop of it. Increasing the `raw` image disk size is done through the [`rawConfiguration`](#operating-system) property.
 
 ```yaml
 kubernetes:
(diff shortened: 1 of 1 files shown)


docs/building-images.md
```diff
@@ -40,6 +40,8 @@ operatingSystem:
   isoConfiguration:
     installDevice: /path/to/disk
     unattended: false
+  rawConfiguration:
```
atanasdinov, line 43: This was initially intentionally omitted due to the conflict with `isoConfiguration`. Not sure if we want to keep it.

ipetrov117: I think we should have it in the examples and maybe specify in the parameter options that it should not be used in cohesion with the `isoConfiguration` parameter. Otherwise users might miss it (as did I). Nevertheless, if you think that this is not a good idea I do not mind removing it.

ipetrov117: Iterating over this, actually we do not need an example, as we ensure that the user will know about the `rawConfiguration ` property from the note in the `Kubernetes` section. Property example has been removed.

docs/building-images.md
```diff
@@ -163,6 +163,8 @@ The Kubernetes configuration section is another entirely optional one.
 It contains all necessary settings to configure and bootstrap a Kubernetes cluster.
 The supported Kubernetes distributions are K3s and RKE2.
 
+> **_NOTE:_** In addition to the configuration below, if you are building a `raw` image, you must manually specify its disk size. The disk size specification is needed in order to ensure that the `raw` image has enough space to host the Kubernetes tarball resources that EIB attempts to copy ontop of it. Increasing the `raw` image disk size is done through the [`rawConfiguration`](#operating-system) property.
```
jdob, line 166:
```suggestion
> **_NOTE:_** In addition to the configuration below, if you are building a `raw` image, you must manually specify its disk size. The disk size specification is needed in order to ensure that the `raw` image has enough space to host the Kubernetes tarball resources that EIB attempts to copy into it. Increasing the `raw` image disk size is done through the [`rawConfiguration`](#operating-system) property.
```

jdob: Minor rephrasing; "ontop" isn't a word.

jdob (approved):

atanasdinov (approved):

Issue #302, opened 2024-03-12
Definition validate command should be its own subcommand instead of a flag


jdob:
EIB uses subcommands (e.g. `eib build ...`) to trigger the build. Currently, the flag to only do definition validation is a flag to the build command. Before we release 1.0, we should migrate this to being its own subcommand (e.g. `eib validate`).

PR #309, opened 2024-03-13
Add validate subcommand

atanasdinov:
- Introduces new `validate` subcommand
- Drops the `--validate` argument from the `build` subcommand
- Improves CLI error handling
- Closes https://github.com/suse-edge/edge-image-builder/issues/302

**Successful validation:**
```
suse-edge380@edge380:~/atanas/eib> podman run --rm -it -v /home/suse-edge380/atanas/eib:/eib eib:dev validate --definition-file k3s-single-node.yaml
Checking image config dir...
Parsing image definition...
Validating image definition...
The specified image definition is valid.
```

**Unsuccessful validation:**
```
suse-edge380@edge380:~/atanas/eib> podman run --rm -it -v /home/suse-edge380/atanas/eib:/eib eib:dev validate --definition-file k3s-single-node.yaml
Checking image config dir...
Parsing image definition...
Validating image definition...
	Image
		The 'imageType' field must be one of: iso, raw
		The 'arch' field must be one of: aarch64, x86_64
	Operating System
		User 'root' must have either a password or at least one SSH key.
		Duplicate username found: root

Please check the log file under the validation directory for more information.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,13 +42,31 @@ For details on how to create the artifacts needed to build an image, see the
 
 The image configuration directory must be attached to the container at runtime. This serves as both the mechanism
 to introduce image definition files and provide a way to get the built image out of the container and onto
-the host machine. 
+the host machine.
 
-The following example command attaches the directory and runs EIB:
+#### Validating image definition
+
+The following example command attaches the image configuration directory and validates a definition:
 ```shell
-podman run --rm -it \
--v $IMAGE_DIR:/eib eib:dev build \
---definition-file $DEFINITION_FILE.yaml
+podman run --rm -it -v $IMAGE_DIR:/eib \
+eib:dev \
+validate --definition-file $DEFINITION_FILE.yaml
+```
+
+* `-v` - Used to mount a local directory (in this example, the value of $IMAGE_DIR) into the EIB container at `/eib`.
+* `--definition-file` - Specifies which image definition file to build. The path to this file will be relative to
+  the image configuration directory. If the definition file is in the root of the configuration directory, simply
+  specify the name of the configuration file.
+* `--config-dir` - (Optional) Specifies the image configuration directory. This path is relative to the running container, so its
+  value must match the mounted volume. It defaults to `/eib` which matches the mounted volume `$IMAGE_DIR:/eib` in the example above.
+
+#### Building image
+
+The following example command attaches the image configuration directory and builds an image:
+```shell
+podman run --rm -it -v $IMAGE_DIR:/eib \
+eib:dev \
+build --definition-file $DEFINITION_FILE.yaml
 ```
 
 **NOTE:**
@@ -66,8 +84,6 @@ which require it (e.g. Elemental, Kubernetes SELinux, etc.).
   for assembling/generating the components used in the build which will persist after EIB finishes. This may also be
   specified to another location within a mounted volume. The directory will contain subdirectories storing the
   respective artifacts of the different builds as well as cached copies of certain downloaded files.
-* `--validate` - If specified, the specified image definition and configuration directory will be checked to ensure
-  the build can proceed, however the image will not actually be built.
 
 
 ## Testing Images
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ The image configuration directory must be attached to the container at runtime. 
 to introduce image definition files and provide a way to get the built image out of the container and onto
 the host machine.
 
-#### Validating image definition
+#### Validating an image definition
 
 The following example command attaches the image configuration directory and validates a definition:
 ```shell
@@ -60,7 +60,7 @@ validate --definition-file $DEFINITION_FILE.yaml
 * `--config-dir` - (Optional) Specifies the image configuration directory. This path is relative to the running container, so its
   value must match the mounted volume. It defaults to `/eib` which matches the mounted volume `$IMAGE_DIR:/eib` in the example above.
 
-#### Building image
+#### Building an image
 
 The following example command attaches the image configuration directory and builds an image:
 ```shell
--- a/pkg/cli/build/validate.go
+++ b/pkg/cli/build/validate.go
@@ -84,13 +84,13 @@ func validateImageDefinition(ctx *image.Context) *cmd.Error {
 	slices.Sort(orderedComponentNames)
 
 	for _, componentName := range orderedComponentNames {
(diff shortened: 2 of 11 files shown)


pkg/cli/cmd/error.go
```diff
@@ -0,0 +1,23 @@
...
+
+	"github.com/suse-edge/edge-image-builder/pkg/log"
+)
+
+type Error struct {
+	UserMessage string
+	LogMessage  string
+}
+
+func LogError(err *Error, checkLogMessage string) {
```
jdob, line 14: Why pass in the check message instead of making the constant here? If this is the one responsible for actually printing the messages to the screen, IMO it feels more natural here instead of passing it in each time.

pkg/cli/cmd/error.go
```diff
@@ -0,0 +1,23 @@
...
+	if err.LogMessage == "" {
```
jdob, line 15: I don't quite understand this logic. If there is no message for the log, why do we instead print the user message as an error?

pkg/cli/cmd/build.go
```diff
@@ -10,10 +10,23 @@ type BuildFlags struct {
 	DefinitionFile string
 	ConfigDir      string
 	RootBuildDir   string
-	Validate       bool
 }
 
-var BuildArgs BuildFlags
+var (
```
jdob, line 15: This might be excessive, but it feels weird (to me, this is a personal preference sort of thing) to have these defined here even though they are used in another file. Did you consider something like `flags.go`?

pkg/log/audit.go
```diff
@@ -37,6 +37,10 @@ func AuditInfof(message string, args ...any) {
 	doAudit(auditMe, zap.S().Info)
 }
 
+func AuditError(message string) {
```
jdob, line 40: Lol, it's back!

pkg/log/log.go
```diff
@@ -0,0 +1,20 @@
+package log
+
+import (
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+)
+
+func ConfigureGlobalLogger(logFilename string) {
```
jdob, line 8: Splitting this out is a good addition, even outside of the need to because of the second subcommand.

README.md
```diff
@@ -42,13 +42,31 @@ For details on how to create the artifacts needed to build an image, see the
 
 The image configuration directory must be attached to the container at runtime. This serves as both the mechanism
 to introduce image definition files and provide a way to get the built image out of the container and onto
-the host machine. 
+the host machine.
 
-The following example command attaches the directory and runs EIB:
+#### Validating image definition
```
jdob, line 47:
```suggestion
#### Validating an image definition
```

README.md
```diff
@@ -42,13 +42,31 @@ For details on how to create the artifacts needed to build an image, see the
...
+```
+
+* `-v` - Used to mount a local directory (in this example, the value of $IMAGE_DIR) into the EIB container at `/eib`.
+* `--definition-file` - Specifies which image definition file to build. The path to this file will be relative to
+  the image configuration directory. If the definition file is in the root of the configuration directory, simply
+  specify the name of the configuration file.
+* `--config-dir` - (Optional) Specifies the image configuration directory. This path is relative to the running container, so its
+  value must match the mounted volume. It defaults to `/eib` which matches the mounted volume `$IMAGE_DIR:/eib` in the example above.
+
+#### Building image
```
jdob, line 63:
```suggestion
#### Building an image
```

pkg/cli/build/validate.go
```diff
@@ -0,0 +1,102 @@
...
+	for c := range failedValidations {
+		orderedComponentNames = append(orderedComponentNames, c)
+	}
+	slices.Sort(orderedComponentNames)
+
+	for _, componentName := range orderedComponentNames {
+		userMessageBuilder.WriteString("\t" + componentName + "\n")
+
+		for _, cf := range failedValidations[componentName] {
+			userMessageBuilder.WriteString("\t\t" + cf.UserMessage + "\n")
```
jdob, line 90: Personally, I'm a much bigger fan of 2 spaces instead of tabs (so 4 spaces in this instance); the depth to which tabs indent looks awkward to me. How strongly do you feel about moving to tabs instead?

I don't know if there's a standard per se, but anecdotally looking around 2 spaces seems to be more common:

```
$ ln --help
Usage: ln [OPTION]... [-T] TARGET LINK_NAME
  or:  ln [OPTION]... TARGET
  or:  ln [OPTION]... TARGET... DIRECTORY
  or:  ln [OPTION]... -t DIRECTORY TARGET...
In the 1st form, create a link to TARGET with the name LINK_NAME.

[SNIP]

Mandatory arguments to long options are mandatory for short options too.
      --backup[=CONTROL]      make a backup of each existing destination file
  -b                          like --backup but does not accept an argument
  -d, -F, --directory         allow the superuser to attempt to hard link
                                directories (note: will probably fail due to
                                system restrictions, even for the superuser)
```

```
$ podman -h
Manage pods, containers and images

Usage:
  podman [options] [command]

Available Commands:
  attach      Attach to a running container
  auto-update Auto update containers according to their auto-update policy
```

```
$ git status
On branch debugging
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   docs/debugging.md
```

pkg/cli/cmd/error.go
atanasdinov: Because it's not a constant. The different subcommands pass different `checkLogMessage`.

pkg/cli/build/validate.go
```diff
@@ -0,0 +1,102 @@
...
+)
+
+const (
+	checkValidationLogMessage = "Please check the log file under the validation directory for more information."
+)
+
+func Validate(_ *cli.Context) error {
+	args := &cmd.BuildArgs
+
+	validationDir := filepath.Join(args.ConfigDir, "_validation")
```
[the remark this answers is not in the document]
atanasdinov: It's not actually easy to plug it in the current setup. Each build has its own `build-<timestamp>` dir so this log file would have to live elsewhere. Having `_build/validate-<timestamp>.log` files hanging around the `cache` and build directories felt very weird so I put everything under a separate dir.

jdob (approved):
LGTM

PR #311, opened 2024-03-13
Helm CA file support

dbw7:


pkg/image/validation/kubernetes.go
```diff
@@ -317,9 +323,51 @@ func validateHelmRepoArgs(parsedURL *url.URL, repo *image.HelmRepository) []Fail
 		})
 	}
 
+	if repo.SkipTLSVerify && repo.CAFile != "" {
+		failures = append(failures, FailedValidation{
+			UserMessage: fmt.Sprintf("Helm repository 'caFile' field for %q should not be defined while 'skipTLSVerify' is true.", repo.Name),
```
atanasdinov, line 328: Nit:
```suggestion
			UserMessage: fmt.Sprintf("Helm repository 'caFile' field for %q cannot be defined while 'skipTLSVerify' is true.", repo.Name),
```

pkg/image/validation/kubernetes.go
```diff
@@ -317,9 +323,51 @@ func validateHelmRepoArgs(parsedURL *url.URL, repo *image.HelmRepository) []Fail
...
+		})
+	}
+
+	if repo.PlainHTTP && repo.CAFile != "" {
+		failures = append(failures, FailedValidation{
+			UserMessage: fmt.Sprintf("Helm repository 'caFile' field for %q should not be defined while 'plainHTTP' is true.", repo.Name),
```
atanasdinov, line 334:
```suggestion
			UserMessage: fmt.Sprintf("Helm repository 'caFile' field for %q cannot be defined while 'plainHTTP' is true.", repo.Name),
```

pkg/image/validation/kubernetes.go
```diff
@@ -317,9 +323,51 @@ func validateHelmRepoArgs(parsedURL *url.URL, repo *image.HelmRepository) []Fail
...
+
 	return failures
 }
 
+func validateHelmRepoCert(repoName, certFile string, imageConfigDir string) string {
+	if certFile == "" {
+		return ""
+	}
+
+	validExtensions := []string{".pem", ".crt", ".cer", ".der", ".p7b", ".p7c", ".pfx", ".p12"}
```
atanasdinov, line 352: @rdoxenham could you please confirm these?

pkg/image/validation/kubernetes.go
```diff
@@ -317,9 +323,51 @@ func validateHelmRepoArgs(parsedURL *url.URL, repo *image.HelmRepository) []Fail
...
+	if !slices.Contains(validExtensions, filepath.Ext(certFile)) {
+		return fmt.Sprintf("Helm chart 'caFile' field for %q must be the name of a valid cert file with the following possible endings: '.pem', '.crt', '.cer', '.der', '.p7b', '.p7c', '.pfx', '.p12'.", repoName)
+	}
+
+	certFilePath := filepath.Join(imageConfigDir, combustion.K8sDir, combustion.HelmDir, combustion.CertsDir, certFile)
+	_, err := os.Stat(certFilePath)
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return fmt.Sprintf("Helm repo cert file '%s' could not be found at '%s'.", certFile, certFilePath)
```
atanasdinov, line 361: I'm not  sure whether mixing up the terminology is the right path here... we're technically only referring to CA bundles and not cert files.

pkg/image/validation/kubernetes.go
atanasdinov: We're only talking about CA bundles here, I'm afraid I don't know enough on the topic. 

```
      --ca-file string     verify certificates of HTTPS-enabled servers using this CA bundle
      --cert-file string   identify registry client using this SSL certificate file
```

dbw7: Should note that `--cert-file` file isn't included in this, this would just be support for `--ca-file`. `--cert-file` is a separate thing that needs to be used in combination with `--key-file` (as far as I understand).

atanasdinov (approved):

Issue #315, opened 2024-03-14
If Elemental fails to register (for whatever reason) we drop to emergency shell


rdoxenham:
In the current implementation of the Elemental registration script (31-elemental-register.sh.tpl) we explicitly call `elemental-register` outside of systemd:

```
# Register --no-toolkit disables OS management
elemental-register --config-path /etc/elemental/config.yaml --state-path /etc/elemental/state.yaml --install --no-toolkit
```

See: https://github.com/suse-edge/edge-image-builder/blob/main/pkg/combustion/templates/31-elemental-register.sh.tpl#L7-L8

The problem with this approach is that if Elemental fails to register, for whatever reason, e.g. DNS issues, network issues, the system boot is halted due to Emergency Shell where it requires manual user intervention. We should never be in a position where we're halting the boot because of something that could be a temporary issue and certainly not something that's a catastrophic failure. The registration should be handled by systemd post-boot, where we also have the opportunity to continue to retry, e.g. when network becomes available or DNS records are correctly updated.

See example of a certificate issue on Rancher (although the example below was because of using the IP rather than hostname for Rancher/Elemental endpoint, but the point still stands):

![image](https://github.com/suse-edge/edge-image-builder/assets/1042177/06e0f52b-6090-4851-bff3-572883471882)

PR #316, opened 2024-03-14
Moved Elemental register to a systemd unit and updated temporary Elemental RPM location

rdoxenham:
In this PR we move Elemental register to a systemd unit instead, that attempts to continually register rather than halting the boot and dropping to an emergency shell that requires manual intervention. We also move the package repo to the *other* temporary repo for 5.5 maintenance from OBS. This will get replaced with the proper repo in the coming days, but this unblocks folks wanting to try EIB+Elemental in the meantime.

Fixes: #315

### DIFF
--- a/pkg/combustion/templates/31-elemental-register.sh.tpl
+++ b/pkg/combustion/templates/31-elemental-register.sh.tpl
@@ -4,12 +4,29 @@ set -euo pipefail
 mkdir -p /etc/elemental
 cp ./{{ .ConfigFile }} /etc/elemental/config.yaml
 
-# Register --no-toolkit disables OS management
-elemental-register --config-path /etc/elemental/config.yaml --state-path /etc/elemental/state.yaml --install --no-toolkit
+# Enable systemd based Elemental registration
+# Register --no-toolkit disables OS management in Rancher
+cat <<- EOF > /etc/systemd/system/elemental-register-systemd.service
+[Unit]
+Description=Elemental Register Install via Systemd
+Wants=network-online.target
+After=network-online.target
+ConditionPathExists=!/etc/rancher/elemental/agent/elemental_connection.json
+
+[Install]
+WantedBy=network-online.target
+
+[Service]
+EnvironmentFile=-/etc/sysconfig/proxy
+Type=oneshot
+ExecStart=/usr/sbin/elemental-register --debug --config-path /etc/elemental/config.yaml --state-path /etc/elemental/state.yaml --install --no-toolkit
+ExecStartPost=/usr/bin/cp /var/lib/elemental/agent/elemental_connection.json /etc/rancher/elemental/agent
+Restart=on-failure
+RestartSec=10
+EOF
 
 # Enable elemental-system-agent
-# On SLEMicro /var/lib is not persistent, so we copy elemental_connection.json in ExecStartPre
-cp /var/lib/elemental/agent/elemental_connection.json /etc/rancher/elemental/agent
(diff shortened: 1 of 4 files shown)


jdob:
One quick note: can you add a line for the bug fix into RELEASE_NOTES under root?

pkg/combustion/templates/31-elemental-register.sh.tpl
```diff
@@ -4,12 +4,29 @@ set -euo pipefail
...
+ExecStart=/usr/sbin/elemental-register --debug --config-path /etc/elemental/config.yaml --state-path /etc/elemental/state.yaml --install --no-toolkit
+ExecStartPost=/usr/bin/cp /var/lib/elemental/agent/elemental_connection.json /etc/rancher/elemental/agent
+Restart=on-failure
+RestartSec=10
+EOF
 
 # Enable elemental-system-agent
-# On SLEMicro /var/lib is not persistent, so we copy elemental_connection.json in ExecStartPre
-cp /var/lib/elemental/agent/elemental_connection.json /etc/rancher/elemental/agent
+# On SLE Micro /var/lib is not persistent, so we copy elemental_connection.json in ExecStartPre
```
e-minguez, line 29:
```suggestion
```

pkg/combustion/templates/31-elemental-register.sh.tpl
```diff
@@ -20,16 +37,17 @@ After=time-sync.target
...
 [Service]
 Type=simple
 Restart=always
 RestartSec=5s
-StandardOutput=journal+console
-StandardError=journal+console
+StandardOutput=journal
+StandardError=journal
 Environment="CATTLE_AGENT_CONFIG=/etc/rancher/elemental/agent/config.yaml"
 ExecStartPre=/bin/sh -c "mkdir -p /var/lib/elemental/agent && cp /etc/rancher/elemental/agent/elemental_connection.json /var/lib/elemental/agent"
```
e-minguez, line 48:
```suggestion
# On SLE Micro /var/lib is not persistent, so we copy elemental_connection.json in ExecStartPre
ExecStartPre=/bin/sh -c "mkdir -p /var/lib/elemental/agent && cp /etc/rancher/elemental/agent/elemental_connection.json /var/lib/elemental/agent"
```

pkg/combustion/templates/31-elemental-register.sh.tpl
```diff
@@ -20,16 +37,17 @@ After=time-sync.target
```
e-minguez, line 37: Does this service require the other being successful? if so, a Wants directive here won't hurt

pkg/combustion/templates/31-elemental-register.sh.tpl
rdoxenham: This still needs to persist I think, we don't copy the file during Combustion as this isn't created until post-creation. But the `ExecStartPre` is still there.

pkg/combustion/templates/31-elemental-register.sh.tpl
rdoxenham: See above. I don't think we want to put this into the systemd unit file either.

pkg/combustion/templates/31-elemental-register.sh.tpl
rdoxenham: I don't think this matters at all at this stage of the build. I'd prefer to leave it aligned with the Elemental team's systemd unit files.

jdob (approved):

atanasdinov (approved):
Thanks for taking the time to drop the leftovers!

Issue #289, opened 2024-03-11
The services for RPM dependency resolution failed to start


e-minguez:
It happened to me a couple of times already while building a raw image using EIB:

```
The services for RPM dependency resolution failed to start. Please check the eib-build.log file under the build directory for more information.
```

The eib-build.log shows:
```
2024-03-11T15:44:34.733Z	INFO	log/audit.go:58	The services for RPM dependency resolution failed to start. Please check the eib-build.log file under the build directory for more information.
2024-03-11T15:44:34.733Z	ERROR	build/build.go:287	creating new podman connection: unable to connect to Podman socket: Get "http://d/v4.8.3/libpod/_ping": dial unix /run/podman/podman.sock: connect: no such file or directory
github.com/suse-edge/edge-image-builder/pkg/cli/build.bootstrapDependencyServices
	/home/abuild/rpmbuild/BUILD/edge-image-builder-1.0.0.rc1/pkg/cli/build/build.go:287
github.com/suse-edge/edge-image-builder/pkg/cli/build.Run
	/home/abuild/rpmbuild/BUILD/edge-image-builder-1.0.0.rc1/pkg/cli/build/build.go:77
github.com/urfave/cli/v2.(*Command).Run
	/home/abuild/rpmbuild/BUILD/edge-image-builder-1.0.0.rc1/vendor/github.com/urfave/cli/v2/command.go:279
github.com/urfave/cli/v2.(*Command).Run
	/home/abuild/rpmbuild/BUILD/edge-image-builder-1.0.0.rc1/vendor/github.com/urfave/cli/v2/command.go:272
github.com/urfave/cli/v2.(*App).RunContext
	/home/abuild/rpmbuild/BUILD/edge-image-builder-1.0.0.rc1/vendor/github.com/urfave/cli/v2/app.go:337
github.com/urfave/cli/v2.(*App).Run
	/home/abuild/rpmbuild/BUILD/edge-image-builder-1.0.0.rc1/vendor/github.com/urfave/cli/v2/app.go:311
main.main
	/home/abuild/rpmbuild/BUILD/edge-image-builder-1.0.0.rc1/cmd/eib/main.go:18
runtime.main
	/usr/lib64/go/1.21/src/runtime/proc.go:267
```

And the podman-system-service.log:

```
time="2024-03-11T15:44:34Z" level=info msg="/usr/bin/podman filtering at log level debug"
[...]

PR #317, opened 2024-03-14
[Bug Fix] Add wait mechanism for Podman's API Listener service

ipetrov117:
Currently there is a race-condition that can happen where we start the [API Listener](https://github.com/suse-edge/edge-image-builder/blob/main/pkg/podman/listener.go#L29) and immediately attempt to make a [connection](https://github.com/suse-edge/edge-image-builder/blob/main/pkg/podman/podman.go#L44) with Podman. 

This sometimes results in the following error:
```bash
creating new podman connection: unable to connect to Podman socket: Get "http://d/v4.8.3/libpod/_ping": dial unix /run/podman/podman.sock: connect: no such file or directory
```
Mainly this happens, because we do not give enough time for the API Listener to create its `podman.sock` file. 

This PR introduces a waiting mechanism that waits for the `podman.sock` file to be created. The wait timeout for this file is 15 seconds, where the file is checked every 3 seconds. The `podman.sock` creation operation is very quick (hence why we didn't catch the race-condition sooner), so 15 seconds is a good timeout value for this file.

Fixes: #289

### DIFF
--- a/pkg/podman/listener.go
+++ b/pkg/podman/listener.go
@@ -7,17 +7,25 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"time"
+
+	"github.com/suse-edge/edge-image-builder/pkg/log"
+	"go.uber.org/zap"
 )
 
 const (
 	podmanArgsBase        = "--log-level=debug system service -t 0"
 	podmanExec            = "/usr/bin/podman"
 	podmanListenerLogFile = "podman-system-service.log"
+	podmanSocketPath      = "/run/podman/podman.sock"
 )
 
 // creates a listening service that answers API calls for Podman (https://docs.podman.io/en/v4.8.2/markdown/podman-system-service.1.html)
 // only way to start the service from within a container - https://github.com/containers/podman/tree/v4.8.2/pkg/bindings#starting-the-service-manually
 func setupAPIListener(out string) error {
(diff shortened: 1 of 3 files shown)


pkg/podman/listener.go
```diff
@@ -42,3 +49,23 @@ func preparePodmanCommand(out io.Writer) *exec.Cmd {
...
+
+	zap.S().Infof("Waiting for '%s' to be created", podmanSocketPath)
+	for i := 0; i < retries; i++ {
+		if _, err := os.Stat(podmanSocketPath); err == nil {
+			zap.S().Infof("'%s' file has been created successfully", podmanSocketPath)
+			return nil
+		}
+
+		zap.S().Infof("'%s' file is not yet created, retrying in %d seconds", podmanSocketPath, sleepSeconds)
+		time.Sleep(sleepSeconds * time.Second)
```
jdob, line 67: What happens if I ctrl+C during a sleep? Given the short sleep it's probably not an issue, I just want to double check that we're not going to lock the user out of exiting.

jdob (approved):

pkg/podman/listener.go
ipetrov117: ctrl+C breaks the sleep, so we wouldn't block the user out of exiting.

PR #322, opened 2024-03-14
Helm template namespace

### LINKED ISSUE #321: Helm chart templating fails in default namespace
Certain Helm charts fail when templated in the default namespace:
`helm-template.log`
```
command: /usr/bin/helm template --skip-crds nvidia-device-plugin /eib/_build/build-Mar14_19-01-19/helm/nvidia-device-plugin-0.14.5.tgz --version v0.14.5 -f /eib/kubernetes/helm/values/nvidia-values.yaml --kube-version v1.27.11+rke2r1
Error: execution error at (nvidia-device-plugin/templates/validation.yml:19:4): 
Running in the 'default' namespace is not recommended.
Set 'allowDefaultNamespace=true' to bypass this error.
Otherwise, use --namespace (with --create-namespace as necessary) to run in a specific namespace.
See: https://helm.sh/docs/helm/helm_install/#options

Use --debug flag to render out invalid YAML
```

Helm charts should be templated in the `targetNamespace` if one is provided.

dbw7:
Closes #321

### DIFF
--- a/pkg/image/context.go
+++ b/pkg/image/context.go
@@ -8,7 +8,7 @@ type HelmClient interface {
 	AddRepo(repository *HelmRepository) error
 	RegistryLogin(repository *HelmRepository) error
 	Pull(chart string, repository *HelmRepository, version, destDir string) (string, error)
-	Template(chart, repository, version, valuesFilePath, kubeVersion string) ([]map[string]any, error)
+	Template(chart, repository, version, targetNamespace, valuesFilePath, kubeVersion string) ([]map[string]any, error)
 }
 
 type networkConfigGenerator interface {
--- a/pkg/image/context.go
+++ b/pkg/image/context.go
@@ -8,7 +8,7 @@ type HelmClient interface {
 	AddRepo(repository *HelmRepository) error
 	RegistryLogin(repository *HelmRepository) error
 	Pull(chart string, repository *HelmRepository, version, destDir string) (string, error)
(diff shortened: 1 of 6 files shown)


pkg/registry/helm.go
```diff
@@ -54,7 +54,7 @@ func handleChart(chart *image.HelmChart, repo *image.HelmRepository, valuesDir,
 		return nil, fmt.Errorf("downloading chart: %w", err)
 	}
 
-	images, err := getChartContainerImages(chart, helmClient, chartPath, valuesPath, kubeVersion)
+	images, err := getChartContainerImages(chart, helmClient, chartPath, valuesPath, kubeVersion, chart.TargetNamespace)
```
atanasdinov, line 57: We already pass the full chart - there's no need to provide the namespace separately.

pkg/image/context.go
```diff
@@ -8,7 +8,7 @@ type HelmClient interface {
 	AddRepo(repository *HelmRepository) error
 	RegistryLogin(repository *HelmRepository) error
 	Pull(chart string, repository *HelmRepository, version, destDir string) (string, error)
-	Template(chart, repository, version, valuesFilePath, kubeVersion string) ([]map[string]any, error)
+	Template(chart, repository, version, targetNamespace, valuesFilePath, kubeVersion string) ([]map[string]any, error)
```
atanasdinov, line 11: Please use this exact positioning for both implementations.

atanasdinov (changes requested):
We must also update the release notes.

atanasdinov (approved):

Issue #233, opened 2024-02-26
Use different Helm chart sources for development and production builds


atanasdinov:
As of https://github.com/suse-edge/edge-image-builder/pull/225, the MetalLB and Endpoint Copier Operator charts will be pulled and installed directly from the GitHub charts repository.

This is fine and switching it to the charts produced in OBS should also not be a problem, however, we'd probably want to consume these charts from IBS in a production EIB container.

We need to figure out how we'd like to differentiate between the different environments and their artefacts.

PR #338, opened 2024-03-28
Extract configurable Helm chart repository variable

atanasdinov:
- Extracts a SUSE Edge chart repository variable to a new `env` package
- Modifying the variable is as simple as adding a flag to the build:
   ```
   MODULE=github.com/suse-edge/edge-image-builder
   go build -ldflags "-X $MODULE/pkg/env.SUSEEdgeRepository=<repository-url>" ./cmd/eib
   ```
- Closes https://github.com/suse-edge/edge-image-builder/issues/233

### DIFF
--- a/pkg/env/env.go
+++ b/pkg/env/env.go
@@ -1,0 +1,5 @@
+package env
+
+var (
+	SUSEEdgeRepository = "https://suse-edge.github.io/charts"
+)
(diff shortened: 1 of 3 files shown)


pkg/env/env.go
```diff
@@ -0,0 +1,5 @@
+package env
```
atanasdinov, line 1: Not quite sure about the name of the package. I guess it'll be more clear once we have more variables to configure.

dbw7 (approved):
Looks good to me!

PR #346, opened 2024-04-01
Unattended changes

dbw7:


atanasdinov (approved):
Please update the release notes too.

rdoxenham (approved):
Given internal discussions, I'm fine with how this has been cleaned up. Ultimately, if the user specifies the `installDevice` it should skip GRUB. Otherwise we've got the `unattended` flag that doesn't really add much without the `installDevice`. One should ultimately require the other.

PR #349, opened 2024-04-02
Fix local RPM dir reading

atanasdinov:


ipetrov117 (approved):
Works as expected. Tested for:
1. Missing GPG dir failure when GPG validation is enabled
2. Existing GPG dir failure when GPG validation is disabled
3. RPM resolution with side-loaded RPMs + GPG validation
4. RPM resolution with side-loaded RPMs without GPG validation

PR #356, opened 2024-04-05
Fix Helm chart OCI registry handling

### LINKED ISSUE #355: Helm fails getting charts stored in unauthenticated OCI registries
The current Helm implementation falsely assumes that all registries are authenticated and attempts to login to every Helm chart OCI registry with an empty string as the username and password, causing it to fail.

Example:
```
command: /usr/bin/helm registry login registry.suse.com
Username: Error: EOF
```

dbw7:
Closes #355 .

### DIFF
--- a/pkg/registry/helm_test.go
+++ b/pkg/registry/helm_test.go
@@ -197,6 +197,10 @@ func TestDownloadChart_FailedRegistryLogin(t *testing.T) {
 	helmChart := &image.HelmChart{}
 	helmRepo := &image.HelmRepository{
 		URL: "oci://registry-1.docker.io/bitnamicharts/apache",
+		Authentication: image.HelmAuthentication{
+			Username: "wrong",
+			Password: "creds",
+		},
 	}
 
 	helmClient := mockHelmClient{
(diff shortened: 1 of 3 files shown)


pkg/registry/helm_test.go
```diff
@@ -197,6 +197,10 @@ func TestDownloadChart_FailedRegistryLogin(t *testing.T) {
 	helmChart := &image.HelmChart{}
 	helmRepo := &image.HelmRepository{
 		URL: "oci://registry-1.docker.io/bitnamicharts/apache",
+		Authentication: image.HelmAuthentication{
```
atanasdinov, line 200: Can we add a test case without credentials?

dbw7: I would say this is covered in [TestDownloadChart](https://github.com/suse-edge/edge-image-builder/pull/356/files#diff-282f47019ad0f92da3ac65274bbb93573234e6faf7ccb50c65ba93a1b414ba84R242) as it uses an OCI repository with no credentials while all of the other failure cases of downloadChart are covered.

atanasdinov: Not really. In fact we need to split up the cases where we:
- return `nil` if there are credentials for successful operation
- return error if there are credentials for failed operation
- panic if there aren't credentials since the function should then never be called

atanasdinov (approved):

PR #357, opened 2024-04-05
Embedded artifact registry progress bar

dbw7:
A progress bar is now shown when populating the embedded artifact registry to demonstrate progress. Previously, the embedded artifact registry would just run in the background and it was difficult to tell if it was still going or if there was something wrong.

Example output:
```
Identifier ................... [SUCCESS]
Custom Files ................. [SUCCESS]
Time ......................... [SKIPPED]
Network ...................... [SKIPPED]
Groups ....................... [SKIPPED]
Proxy ........................ [SKIPPED]
Rpm .......................... [SKIPPED]
Systemd ...................... [SKIPPED]
Elemental .................... [SKIPPED]
Suma ......................... [SKIPPED]
Populating Embedded Artifact Registry... 100% |██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| (2/2, 19 it/min)  
```

### DIFF
--- a/pkg/combustion/registry.go
+++ b/pkg/combustion/registry.go
@@ -10,15 +10,13 @@ import (
 	"strings"
 
 	"github.com/schollz/progressbar/v3"
-
-	"gopkg.in/yaml.v3"
-
 	"github.com/suse-edge/edge-image-builder/pkg/fileio"
 	"github.com/suse-edge/edge-image-builder/pkg/image"
 	"github.com/suse-edge/edge-image-builder/pkg/log"
 	"github.com/suse-edge/edge-image-builder/pkg/registry"
 	"github.com/suse-edge/edge-image-builder/pkg/template"
 	"go.uber.org/zap"
+	"gopkg.in/yaml.v3"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


pkg/combustion/registry.go
```diff
@@ -373,6 +376,10 @@ func populateRegistry(ctx *image.Context, images []string) error {
 		if err := generateRegistryTar(ctx, imageTarDest); err != nil {
 			return fmt.Errorf("generating hauler store tar: %w", err)
 		}
+
+		if err := bar.Add(1); err != nil {
+			zap.S().Errorf("Error incrementing the progress bar: %s", err)
```
dbw7, line 381: I'm not sure if handling the error like this is the way to go, or if it's even necessary, open to input.

atanasdinov: We can use the debug level for this, it's not that imporant.

pkg/combustion/registry.go
```diff
@@ -9,6 +9,8 @@ import (
 	"slices"
 	"strings"
 
+	"github.com/schollz/progressbar/v3"
+
```
atanasdinov, line 13: Can we fix the import ordering? This empty line and the one below should be removed.

atanasdinov (approved):

Issue #374, opened 2024-04-11
Enable SELinux support for Kubernetes agents if servers enforce it


atanasdinov:
SELinux is an option which we'll enforce to be the same regardless of whether the Kubernetes node is a control plane or a worker.

PR #377, opened 2024-04-11
Use same SELinux config option for both server and agent nodes

atanasdinov:
- Set agents' selinux config option to the one provided for servers
- Closes https://github.com/suse-edge/edge-image-builder/issues/374


dbw7 (approved):
Beautiful 👍

dbw7 (approved):

Issue #381, opened 2024-04-12
Empty gpg-keys directory passes GPG enablement only to fail during the dependency resolution


atanasdinov:
We're currently only checking whether the [GPG keys directory exists](https://github.com/suse-edge/edge-image-builder/blob/main/pkg/combustion/rpm.go#L186) before running the dependency resolution which in turn delays the failure:

```
2024-04-12T14:58:01.735Z	FATAL	build/build.go:102	An error occurred building the image: configuring combustion: configuring component "RPM": resolving rpm/package dependencies: building resolver image: building image from context /eib/_build/build-Apr12_14-56-40/resolver-image-build: building at STEP "RUN ./rpm-resolution.sh": while running runtime: exit status 1
```

PR #385, opened 2024-04-15
Ensure rpms/gpg-keys is not empty if present

atanasdinov:
- Closes https://github.com/suse-edge/edge-image-builder/issues/381


ipetrov117 (approved):
Looks great 🚀

Issue #384, opened 2024-04-15
Improve RPM validation


atanasdinov:
- Missing empty values in `packageList`
- Consistency between messages

PR #391, opened 2024-04-16
Improve RPM validation

atanasdinov:
- Closes https://github.com/suse-edge/edge-image-builder/issues/384

### DIFF
--- a/pkg/image/validation/os.go
+++ b/pkg/image/validation/os.go
@@ -3,6 +3,7 @@ package validation
 import (
 	"fmt"
 	"regexp"
+	"slices"
 	"strings"
 
 	"github.com/suse-edge/edge-image-builder/pkg/image"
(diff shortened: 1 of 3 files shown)


pkg/image/validation/os.go
```diff
@@ -189,27 +206,24 @@ func validatePackages(os *image.OperatingSystem) []FailedValidation {
...
+	var repoURLs []string
 
-		if duplicates := findDuplicates(urlSlice); len(duplicates) > 0 {
-			duplicateValues := strings.Join(duplicates, ", ")
-			msg := fmt.Sprintf("The 'additionalRepos' field contains duplicate repos: %s", duplicateValues)
+	for _, repo := range os.Packages.AdditionalRepos {
+		if repo.URL == "" {
 			failures = append(failures, FailedValidation{
-				UserMessage: msg,
+				UserMessage: "The 'url' field is required for all entries under 'additionalRepos'.",
```
atanasdinov, line 214: This is only reworded for consistency with all other functions within the module.

pkg/image/validation/os.go
```diff
@@ -181,6 +182,22 @@ func validateSuma(os *image.OperatingSystem) []FailedValidation {
 func validatePackages(os *image.OperatingSystem) []FailedValidation {
 	var failures []FailedValidation
 
+	if len(os.Packages.PKGList) == 0 {
+		if len(os.Packages.AdditionalRepos) > 0 {
+			failures = append(failures, FailedValidation{
+				UserMessage: "The 'packageList' field is required if any entries are specified under 'additionalRepos'.",
```
ipetrov117, line 188: Isn't this conflicting with the use-case where we have a RPM (in the `rpms` dir) that has some dependencies that should be retrieved from a third-party repository?

Then we would have a configuration where `AdditionalRepos` is defined without an entry in the package list.

atanasdinov: Thanks, I didn't catch that at all. Added a comment to explain it for other developers.

ipetrov117 (approved):
Looks good.

PR #430, opened 2024-05-01
added execute bit to each file in custom/scripts

### LINKED ISSUE #429: automatically set execute bit on scripts
When putting a script that doesn't have the execute bit set in `custom/scripts` combustion will fail with a permission denied error. We should set the execute bit on the script on image creation.

vtrenton:
closes #429


jdob:
Thank you for this! We really appreciate contributions :)

For some background, we used to automatically do this for both custom scripts and files. We received feedback that it wasn't desired for files, but got a little overzealous and removed the executable from both files and scripts.

What I did in a new patch [1] was to resurrect the old handling for when the permissions are set at copy time, but only applying it to scripts and not files. While functionally the same as your patch, it's a little more in line with the general design and conventions used in EIB.

Thanks again :)

[1] https://github.com/suse-edge/edge-image-builder/pull/431

PR #431, opened 2024-05-02
set custom scripts to executable, but custom files retain their existing permissions

jdob:
closes: #429

### DIFF
--- a/pkg/combustion/custom.go
+++ b/pkg/combustion/custom.go
@@ -41,17 +41,18 @@ func configureCustomFiles(ctx *image.Context) ([]string, error) {
 
 func handleCustomFiles(ctx *image.Context) error {
 	fullFilesDir := generateComponentPath(ctx, filepath.Join(customDir, customFilesDir))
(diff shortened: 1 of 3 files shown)


dbw7 (approved):
Looks good!

pkg/combustion/custom.go
```diff
@@ -72,12 +73,18 @@ func copyCustomFiles(fromDir, toDir string) ([]string, error) {
...
 		copyTo := filepath.Join(toDir, entry.Name())
 
-		info, err := entry.Info()
-		if err != nil {
-			return nil, fmt.Errorf("reading file info: %w", err)
+		var mode os.FileMode
+		if filePermissions == nil {
+			info, infoErr := entry.Info()
+			if infoErr != nil {
+				return nil, fmt.Errorf("reading file info: %w", err)
```
atanasdinov, line 80: This is using the wrong error:
```suggestion
				return nil, fmt.Errorf("reading file info: %w", infoErr)
```

atanasdinov (approved):

Issue #447, opened 2024-05-17
EIB doesn't support >512 byte sector size base images


rdoxenham:
The standard SLE Micro SelfInstall ISO contains a raw disk image (that gets copied onto the target block device) that has a 512 byte sector size. When written to a disk that has a 4096 byte sector size, it breaks the partition table and the system is not able to see the partitions that were just written to the disk, leading to dracut looping trying to find the disk. This behaviour is not witnessed with the standard SLE Micro wizard-based (YaST) installer because in that scenario, new partitions are created based on the detected sector size, whereas the SelfInstall ISO uses already created partitions from the original raw disk image layout.

For some customers we're able to build 4096 byte sector size images, and for this, EIB needs to be able to accommodate this, and it will need to detect the appropriate sector size and act accordingly.

PR #448, opened 2024-05-17
Added support for 4096 byte sector size input base images

rdoxenham:
Fixes: #447

### DIFF
--- a/pkg/build/templates/modify-raw-image.sh.tpl
+++ b/pkg/build/templates/modify-raw-image.sh.tpl
@@ -13,6 +13,13 @@ set -euo pipefail
 #
 # Guestfish Command Documentation: https://libguestfs.org/guestfish.1.html
 
+# Test the block size of the base image and adapt to suit either 512/4096 byte images
+BLOCKSIZE=512
+if ! guestfish -i --blocksize=$BLOCKSIZE -a {{.ImagePath}} echo "[INFO] 512 byte sector check successful."; then
+        echo "[WARN] Failed to access image with 512 byte sector size, trying 4096 bytes."
+        BLOCKSIZE=4096
+fi
+
 # Resize the raw disk image to accommodate the users desired raw disk image size
 # This is also required if embedding content into /combustion, especially for airgap.
 # Should *only* execute if the user is building a raw disk image.
@@ -24,7 +31,7 @@ cp {{.ImagePath}}.expanded {{.ImagePath}}
 rm -f {{.ImagePath}}.expanded
 {{ end }}
 
-guestfish --format=raw --rw -a {{.ImagePath}} -i <<'EOF'
+guestfish --blocksize=$BLOCKSIZE --format=raw --rw -a {{.ImagePath}} -i <<'EOF'
   # Enables write access to the read only filesystem
   sh "btrfs property set / ro false"
 
--- a/pkg/rpm/resolver/templates/prepare-tarball.sh.tpl
+++ b/pkg/rpm/resolver/templates/prepare-tarball.sh.tpl
@@ -45,7 +45,7 @@ virt-tar-out --blocksize=$BLOCKSIZE -a $RAW_FILE / - | pigz --best > $WORK_DIR/{
 
 # Test the block size of the base image and adapt to suit either 512/4096 byte images
 BLOCKSIZE=512
-if ! guestfish -i --blocksize=$BLOCKSIZE -a $RAW_FILE echo "[INFO] 512 byte sector check successful."; then
+if ! guestfish -i --blocksize=$BLOCKSIZE -a $IMG_PATH echo "[INFO] 512 byte sector check successful."; then
         echo "[WARN] Failed to access image with 512 byte sector size, trying 4096 bytes."
         BLOCKSIZE=4096
 fi
--- a/pkg/rpm/resolver/templates/prepare-tarball.sh.tpl
+++ b/pkg/rpm/resolver/templates/prepare-tarball.sh.tpl
@@ -33,7 +33,23 @@ if [ `wc -w <<< $RAW_FILE` -ne 1 ]; then
 	exit 1
 fi
 
-virt-tar-out -a $RAW_FILE / - | pigz --best > $WORK_DIR/{{.ArchiveName}}
+# Test the block size of the base image and adapt to suit either 512/4096 byte images
+BLOCKSIZE=512
+if ! guestfish -i --blocksize=$BLOCKSIZE -a $RAW_FILE echo "[INFO] 512 byte sector check successful."; then
+        echo "[WARN] Failed to access image with 512 byte sector size, trying 4096 bytes."
+        BLOCKSIZE=4096
+fi
+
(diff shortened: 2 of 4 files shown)


pkg/rpm/resolver/templates/prepare-tarball.sh.tpl
```diff
@@ -33,7 +33,23 @@ if [ `wc -w <<< $RAW_FILE` -ne 1 ]; then
...
+
+virt-tar-out --blocksize=$BLOCKSIZE -a $RAW_FILE / - | pigz --best > $WORK_DIR/{{.ArchiveName}}
 {{ else }}
-virt-tar-out -a $IMG_PATH / - | pigz --best > $WORK_DIR/{{.ArchiveName}}
-{{ end }}
\ No newline at end of file
+
+# Test the block size of the base image and adapt to suit either 512/4096 byte images
+BLOCKSIZE=512
+if ! guestfish -i --blocksize=$BLOCKSIZE -a $RAW_FILE echo "[INFO] 512 byte sector check successful."; then
```
atanasdinov, line 48: $RAW_FILE is not set here. Are we sure this is working as expected?

rdoxenham: Wow! Great eye.

rdoxenham:
Tested this again with different image types with both 512 and 4096 byte sector sizes. The first two runs are an ISO followed by a RAW disk image with 4096, and the last two are the same but with 512:

```
rdo@woden:~/work/eib-testing/_build% for i in $(ls); do cat $i/prepare-resolver-base-tarball-image.log | grep byte; done
[WARN] Failed to access image with 512 byte sector size, trying 4096 bytes.
[WARN] Failed to access image with 512 byte sector size, trying 4096 bytes.
[INFO] 512 byte sector check successful.
[INFO] 512 byte sector check successful.
```

pkg/build/templates/modify-raw-image.sh.tpl
```diff
@@ -13,6 +13,13 @@ set -euo pipefail
 #
 # Guestfish Command Documentation: https://libguestfs.org/guestfish.1.html
 
+# Test the block size of the base image and adapt to suit either 512/4096 byte images
+BLOCKSIZE=512
+if ! guestfish -i --blocksize=$BLOCKSIZE -a {{.ImagePath}} echo "[INFO] 512 byte sector check successful."; then
+        echo "[WARN] Failed to access image with 512 byte sector size, trying 4096 bytes."
+        BLOCKSIZE=4096
```
e-minguez, line 20: Shall we do the same check here instead of assuming 4096?

pkg/rpm/resolver/templates/prepare-tarball.sh.tpl
```diff
@@ -33,7 +33,23 @@ if [ `wc -w <<< $RAW_FILE` -ne 1 ]; then
 	exit 1
 fi
 
-virt-tar-out -a $RAW_FILE / - | pigz --best > $WORK_DIR/{{.ArchiveName}}
+# Test the block size of the base image and adapt to suit either 512/4096 byte images
+BLOCKSIZE=512
+if ! guestfish -i --blocksize=$BLOCKSIZE -a $RAW_FILE echo "[INFO] 512 byte sector check successful."; then
+        echo "[WARN] Failed to access image with 512 byte sector size, trying 4096 bytes."
+        BLOCKSIZE=4096
```
e-minguez, line 40: Shall we do the same check here instead of assuming 4096?

pkg/rpm/resolver/templates/prepare-tarball.sh.tpl
```diff
@@ -33,7 +33,23 @@ if [ `wc -w <<< $RAW_FILE` -ne 1 ]; then
...
 {{ else }}
-virt-tar-out -a $IMG_PATH / - | pigz --best > $WORK_DIR/{{.ArchiveName}}
-{{ end }}
\ No newline at end of file
+
+# Test the block size of the base image and adapt to suit either 512/4096 byte images
+BLOCKSIZE=512
+if ! guestfish -i --blocksize=$BLOCKSIZE -a $IMG_PATH echo "[INFO] 512 byte sector check successful."; then
+        echo "[WARN] Failed to access image with 512 byte sector size, trying 4096 bytes."
+        BLOCKSIZE=4096
```
e-minguez, line 50: Shall we do the same check here instead of assuming 4096?

pkg/build/templates/modify-raw-image.sh.tpl
rdoxenham: @e-minguez thanks for checking this, but this statement *is* the check. We start with assuming 512 bytes, we then run `guestfish` with a 512 byte sector size, and if that fails (it will show up as `operating system not found` in the logs), we throw a warning and assume 4096. This value then feeds into the next call in the script, the full `guestfish` run where we run the necessary modifications on the raw disk image.

pkg/rpm/resolver/templates/prepare-tarball.sh.tpl
rdoxenham: @e-minguez as per the previous comment, this is the check. This then feeds into the next call `virt-tar-out` which needs to also be aware of the block size. It will use 4096 *only* if a simple echo check with 512 fails.

pkg/rpm/resolver/templates/prepare-tarball.sh.tpl
rdoxenham: @e-minguez as before. Thanks again!

atanasdinov (approved):

pkg/build/templates/modify-raw-image.sh.tpl
e-minguez: what I was saying is something like:

```
BLOCKSIZE=512
if ! guestfish -i --blocksize=$BLOCKSIZE -a {{.ImagePath}} echo "[INFO] 512 byte sector check successful."; then
  echo "[WARN] Failed to access image with 512 byte sector size, trying 4096 bytes."
  BLOCKSIZE=4096
  if ! guestfish -i --blocksize=$BLOCKSIZE -a {{.ImagePath}} echo "[INFO] 4096 byte sector check successful."; then
    echo "[ERROR] Failed to access image with 512 or 4096 byte sector sizes."
fi
```

(this can be improved of course)

agracey: Are there other blocksizes we need to accomodate for? 

Is there any way the check could return a false positive?

e-minguez: not that I'm aware of, but _just in case_ I guess it won't hurt to check both :)

rdoxenham: I hear what you're saying, Edu, but each `guestfish` call adds to the image build time. I think we should assume 512 (as the vast majority of situations will be 512) and if that fails, try with 4096 (the only other somewhat common block size). I don't think it makes sense to check for 4096 before proceeding, at this point we should just error out.

e-minguez: Agree!

e-minguez (approved):

PR #449, opened 2024-05-21
Adding Elemental Reset capabilities to nodes provisioned via EIB

rdoxenham:
Currently, SLE Micro hosts that register against Elemental (and are provisioned via EIB) cannot utilise the `elemental.cattle.io/resettable: 'true'` capability to wipe Kubernetes cluster information and restore the machine to the inventory so it can be re-used. The current behaviour is that the node is deleted, and manual action needs to take place to clean up data and re-register the machine, which is undesirable.

In https://github.com/rancher/elemental-operator/pull/731 I've made a suggestion on how we can implement OS-agnostic system markers that can be watched to trigger any desired cleanup actions, bypassing the need to have Elemental be concerned with the type of OS being utilised. As part of Edge we should look for this marker and take appropriate action. In this PR we introduce a `systemd.path` unit that watches for the presence of `/var/lib/elemental/.unmanaged_reset` and executes a slightly stripped down version of https://github.com/suse-edge/misc/blob/main/elemental/edge-node-cleanup.sh. 

This PR is considered WIP because it relies on the elemental-operator patch landing, but some additional thoughts-

1. Should we make this pluggable, i.e. via an EIB switch to enable node reset, rather than having it enabled by default as this PR suggests?
2. Should we also reset the SLE Micro snapshot back to original state as part of the cleanup procedure? Currently we only care about Kubernetes data and uninstalling k3s/rke2.

Fixes: #450

### DIFF
--- a/pkg/combustion/elemental_test.go
+++ b/pkg/combustion/elemental_test.go
@@ -66,5 +66,5 @@ func TestWriteElementalCombustionScript(t *testing.T) {
 	assert.Contains(t, found, "/etc/systemd/system/elemental-reset.path")
 	assert.Contains(t, found, "/etc/systemd/system/elemental-reset.service")
 	assert.Contains(t, found, "mkdir -p /opt/edge/")
-	assert.Contains(t, found, "cat <<- EOF > /opt/edge/node_cleanup.sh")
+	assert.Contains(t, found, "cat <<- \\EOF > /opt/edge/node_cleanup.sh")
 }
--- a/pkg/combustion/elemental_test.go
+++ b/pkg/combustion/elemental_test.go
@@ -63,4 +63,8 @@ func TestWriteElementalCombustionScript(t *testing.T) {
 	require.NoError(t, err)
 	found := string(foundBytes)
 	assert.Contains(t, found, "/usr/sbin/elemental-register --debug --config-path /etc/elemental/config.yaml --state-path /etc/elemental/state.yaml --install --no-toolkit")
+	assert.Contains(t, found, "/etc/systemd/system/elemental-reset.path")
+	assert.Contains(t, found, "/etc/systemd/system/elemental-reset.service")
+	assert.Contains(t, found, "mkdir -p /opt/edge/")
+	assert.Contains(t, found, "cat <<- EOF > /opt/edge/node_cleanup.sh")
 }
--- a/pkg/combustion/elemental_test.go
+++ b/pkg/combustion/elemental_test.go
@@ -66,5 +66,5 @@ func TestWriteElementalCombustionScript(t *testing.T) {
 	assert.Contains(t, found, "/etc/systemd/system/elemental-reset.path")
 	assert.Contains(t, found, "/etc/systemd/system/elemental-reset.service")
 	assert.Contains(t, found, "mkdir -p /opt/edge/")
-	assert.Contains(t, found, "cat <<- \\EOF > /opt/edge/node_cleanup.sh")
+	assert.Contains(t, found, "cat <<- \\EOF > /opt/edge/elemental_node_cleanup.sh")
 }
--- a/pkg/combustion/templates/31-elemental-register.sh.tpl
+++ b/pkg/combustion/templates/31-elemental-register.sh.tpl
@@ -155,3 +155,5 @@ rm -f /etc/elemental/state.yaml
 # by restarting the Elemental registration service via systemd
 systemctl restart elemental-register-systemd.service
 EOF
+
+chmod a+x /opt/edge/node_cleanup.sh
--- a/pkg/combustion/templates/31-elemental-register.sh.tpl
+++ b/pkg/combustion/templates/31-elemental-register.sh.tpl
@@ -63,7 +63,7 @@ After=network-online.target
 
 [Service]
 Type=simple
-ExecStart=/opt/edge/node_cleanup.sh -u
+ExecStart=/opt/edge/elemental_node_cleanup.sh -u
 ExecStartPost=/usr/bin/rm -f /var/lib/elemental/.unmanaged_reset
 EOF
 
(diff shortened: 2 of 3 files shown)


pkg/combustion/templates/31-elemental-register.sh.tpl
```diff
@@ -47,5 +47,113 @@ Environment="CATTLE_AGENT_CONFIG=/etc/rancher/elemental/agent/config.yaml"
...
+
+cat <<- EOF > /etc/systemd/system/elemental-reset.service
+[Unit]
+Description=Elemental Reset for Unmanaged Hosts
+Wants=network-online.target
+After=network-online.target
+
+[Service]
+Type=simple
+ExecStart=/opt/edge/node_cleanup.sh -u
```
e-minguez, line 66:
```suggestion
ExecStart=/opt/edge/elemental_node_cleanup.sh -u
```

pkg/combustion/templates/31-elemental-register.sh.tpl
```diff
@@ -47,5 +47,113 @@ Environment="CATTLE_AGENT_CONFIG=/etc/rancher/elemental/agent/config.yaml"
...
+ExecStartPost=/usr/bin/rm -f /var/lib/elemental/.unmanaged_reset
+EOF
+
+systemctl enable elemental-reset.path || true
 systemctl enable elemental-register-systemd.service || true
 systemctl enable elemental-system-agent.service || true
+
+mkdir -p /opt/edge/
+cat <<- \EOF > /opt/edge/node_cleanup.sh
```
e-minguez, line 75:
```suggestion
cat <<- \EOF > /opt/edge/elemental_node_cleanup.sh
```

pkg/combustion/elemental_test.go
```diff
@@ -63,4 +63,8 @@ func TestWriteElementalCombustionScript(t *testing.T) {
 	require.NoError(t, err)
 	found := string(foundBytes)
 	assert.Contains(t, found, "/usr/sbin/elemental-register --debug --config-path /etc/elemental/config.yaml --state-path /etc/elemental/state.yaml --install --no-toolkit")
+	assert.Contains(t, found, "/etc/systemd/system/elemental-reset.path")
+	assert.Contains(t, found, "/etc/systemd/system/elemental-reset.service")
+	assert.Contains(t, found, "mkdir -p /opt/edge/")
+	assert.Contains(t, found, "cat <<- \\EOF > /opt/edge/node_cleanup.sh")
```
e-minguez, line 69:
```suggestion
	assert.Contains(t, found, "cat <<- \\EOF > /opt/edge/elemental_node_cleanup.sh")
```

pkg/combustion/templates/31-elemental-register.sh.tpl
```diff
@@ -47,5 +47,113 @@ Environment="CATTLE_AGENT_CONFIG=/etc/rancher/elemental/agent/config.yaml"
 ExecStart=/usr/sbin/elemental-system-agent sentinel
 EOF
 
+cat <<- EOF > /etc/systemd/system/elemental-reset.path
+[Path]
+PathModified=/var/lib/elemental/.unmanaged_reset
```
e-minguez, line 52: I didn't tested it but I think it would be better to use "PathExists" so the service is only triggered once, otherwise as the service deletes it, it will be maybe trigger it again (?)
Quoting https://www.freedesktop.org/software/systemd/man/latest/systemd.path.html#PathExists=:

```
PathExists= may be used to watch the mere existence of a file or directory. If the file specified exists, the configured unit is activated.
...
PathChanged= may be used to watch a file or directory and activate the configured unit whenever it changes.
It is not activated on every write to the watched file but it is activated if the file which was open for writing gets closed.
```

rdoxenham: Thanks, Edu. I actually tested this quite extensively, and it behaves as expected here... when the file is created, it's triggered. Then the `elemental-reset.service` deletes the file as part of `ExecStartPost`.

rdoxenham:
https://github.com/rancher/elemental-operator/pull/731 landed and will be backported to 1.4.z. We should start the review on our side.

pkg/combustion/templates/31-elemental-register.sh.tpl
atanasdinov, line 75: Backslash intended?

atanasdinov, line 75: Embedding the script like this doesn't sit right with me but I'm fine with merging this and seeing if future modifications complicate it enough to warrant extracting it out.

atanasdinov (commented):
+1 on Edu's suggestion to rename the cleanup script in order to be more intuitive.

agracey (approved):
I don't have hardware with me to test with but the script makes sense. 

Longer term, I think it could make sense to package the service/script up in an RPM that gets installed alongside the elemental-register package

pkg/combustion/templates/31-elemental-register.sh.tpl
agracey: Agreed and I didn't see your comment before making mine

rdoxenham: Yep, backslash intended, and tested :)

rdoxenham: Yeah, I agree, we should revisit this later and package it properly. Reason being, it gives us a way of answering the two questions at the start. But for now, this gives us a way forward.

rdoxenham:
> I don't have hardware with me to test with but the script makes sense.
> 
> Longer term, I think it could make sense to package the service/script up in an RPM that gets installed alongside the elemental-register package

Agreed. We should definitely look to improve the handling here.

pkg/combustion/elemental_test.go
rdoxenham: Done! Thanks!

atanasdinov (approved):

PR #451, opened 2024-05-22
Introduce Combustion type

atanasdinov:


pkg/image/context.go
```diff
@@ -1,48 +1,19 @@
 package image
 
-import (
-	"io"
-)
-
 type HelmClient interface {
```
atanasdinov, line 3: Cannot be moved out until the registry component is refactored.

pkg/combustion/combustion.go
```diff
@@ -21,9 +22,44 @@ import (
...
+
+type rpmResolver interface {
+	Resolve(packages *image.Packages, localRPMConfig *image.LocalRPMConfig, outputDir string) (rpmDirPath string, pkgList []string, err error)
+}
+
+type rpmRepoCreator interface {
+	Create(path string) error
+}
+
+type Combustion struct {
```
jdob, line 50: I went back and forth on this a few times. My initial idea in my head was that these component configuration objects would live under Builder, with it passing them into the combustion methods as necessary.

But now seeing this I kinda like the idea of all of these being encapsulated under the combustion object since, at least for now, they are SUPER scoped to generating combustion scripts and configuration. So +1 from me on this approach.

pkg/cli/build/build.go
```diff
@@ -241,12 +240,16 @@ func appendHelm(ctx *image.Context) {
 	ctx.ImageDefinition.Kubernetes.Helm.Repositories = append(ctx.ImageDefinition.Kubernetes.Helm.Repositories, componentRepos...)
 }
 
-// If the image definition requires it, starts the necessary services, returning an error in the event of failure.
-func bootstrapDependencyServices(ctx *image.Context, rootDir string) *cmd.Error {
+func buildCombustionConfigurator(ctx *image.Context, rootDir string) (*combustion.Combustion, *cmd.Error) {
```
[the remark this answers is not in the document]
atanasdinov: Having of all this in the `combustion` package is not possible due to circular dependencies. We must do it outside in order not to have to import any of the implementations and only rely on the interfaces.

+1 on the `Configurator` comment. I couldn't find a better word - both `Generator` and even `Handler` are better. Will adjust.

jdob (approved):

PR #455, opened 2024-05-27
Make embedded-registry script reentrant

tu1h:
Make embedded-registry script reentrant since it would be failed if `/opt/hauler` already exists.


atanasdinov (approved):

atanasdinov:
Thanks!

PR #467, opened 2024-06-04
Registry component refactoring

atanasdinov:
- Introduces a new Registry type
- Initialising a Registry will now copy all local manifests, download all manifests from listed URLs and pull all Helm charts
- Decouples and drops the dependencies between the Kubernetes and Registry combustion components


pkg/registry/registry.go
```diff
@@ -0,0 +1,211 @@
...
+		helmValuesDir:  helmValuesDir,
+		kubeVersion:    ctx.ImageDefinition.Kubernetes.Version,
+	}, nil
+}
+
+func (r *Registry) ManifestsPath() string {
+	return r.manifestsDir
+}
+
+func storeManifests(ctx *image.Context, localManifestsDir string) (string, error) {
```
jdob, line 65: Not PR Blocking: Assuming I wasn't misreading the `storeManifestImages` method above, this makes me +1 my previous comment. This one actually does a storing operation, which to me is an accurate usage in the method name.

The issue might be as simple as "store" being both a verb and a noun. If that's the case, that's one drawback of not using a `Get` type prefix on retrieval types of methods.

pkg/registry/registry.go
```diff
@@ -0,0 +1,211 @@
...
+	var manifestsPathPopulated bool
+	manifestsDestDir := filepath.Join(ctx.BuildDir, "manifests")
```
jdob, line 67: Personal Preference: I imagine this directory name is only used in this instance, hence it's not a constant. I prefer the consistency of having all of the directory structure names as constants. 

To be clear, I have zero programmatic reason for liking this, it just feels clean to me :)

pkg/registry/registry.go
```diff
@@ -0,0 +1,211 @@
...
+	kubeVersion    string
+}
+
+func New(ctx *image.Context, localManifestsDir string, helmClient helmClient, helmValuesDir string) (*Registry, error) {
+	manifestsDir, err := storeManifests(ctx, localManifestsDir)
+	if err != nil {
+		return nil, fmt.Errorf("storing manifests: %w", err)
+	}
+
+	charts, err := storeHelmCharts(ctx, helmClient)
```
jdob, line 46: This is an interesting philosophical discussion. I was always taught to not to heavyweight operations in a constructor by default. Granted, this is more of a factory than a constructor, and Go has plenty of other differences from Java & Python that I'm still adjusting to. So I'm certainly not pushing you to change it, but if you were at all on the fence that this felt like the wrong place to do it, then you're not alone.

pkg/registry/registry.go
```diff
@@ -0,0 +1,211 @@
...
+			HelmChart:     c,
+			localPath:     localPath,
+			repositoryURL: repository.URL,
+		})
+	}
+
+	return charts, nil
+}
+
+func mapChartRepos(helm *image.Helm) map[string]*image.HelmRepository {
```
jdob, line 141: Nit: This is mostly covered by the return type, but `mapChartsToRepos` would be IMO a bit more clear.

jdob (approved):
A few comments, but none of them blocking if you decide you don't want to make any of the suggested changes.

pkg/combustion/registry.go
```diff
@@ -249,89 +236,6 @@ func (c *Combustion) configureEmbeddedArtifactRegistry(ctx *image.Context) (bool
 	return true, nil
```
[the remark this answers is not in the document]
dbw7: If there are no images we don't send back and error but we use the bool so that we don't continue configuring the EAR
```
configured, err := c.configureEmbeddedArtifactRegistry(ctx)
if err != nil {
	return nil, fmt.Errorf("configuring embedded artifact registry: %w", err)
}

if !configured {
	log.AuditComponentSkipped(registryComponentName)
	zap.S().Info("Skipping embedded artifact registry since the provided manifests/helm charts contain no images")
	return nil, nil
}
```

pkg/registry/registry.go
```diff
@@ -0,0 +1,211 @@
...
+	"path/filepath"
+	"strings"
+
+	"github.com/suse-edge/edge-image-builder/pkg/fileio"
+	"github.com/suse-edge/edge-image-builder/pkg/http"
+	"github.com/suse-edge/edge-image-builder/pkg/image"
+	"go.uber.org/zap"
+)
+
+type helmClient interface {
```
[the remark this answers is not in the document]
atanasdinov: There's no hard rule as to where it should be defined - I did it here since it's only used in `registry.go` in the first place. I tend to keep as many type definitions as possible in the same file, especially when one is embedding others. It's a personal preference, though.

pkg/registry/registry.go
atanasdinov: I'd want to see how it turns out. I'm not quite happy with it exactly due to the reasons you listed, but at the same time I need to do this first because the combustion component runs.

The alternative would be for the constructor  to only initialise some of the struct fields and have another function e.g. `Prepare()` to do the downloading right after, however I find this even more confusing.

dbw7 (approved):
Changes look clean and I ran the HA k3s test case and it seemed to work fine👍

pkg/registry/registry.go
jdob: I've had the same debate in my own head, where I've split the constructor and an init() function. The benefits were there, but minor. In any event, I'm totally cool with your rationale.

Issue #200, opened 2024-02-20
Add support to "drop files into the filesystem" at combustion phase


e-minguez:
Imagine you want to have a dedicated sshd configuration in the /etc/ssh/sshd_config. Currently you need to DIY a combustion script that copies the file to the final destination. It would be nice to add support to something like a 'files' directory where files are copied at combustion time:

```
files/etc/ssh/sshd_config -> /etc/ssh/sshd_config
```

PR #474, opened 2024-06-11
added ability to copy files into the built image's filesystem

jdob:
Closes: #200

### DIFF
--- a/pkg/combustion/osfiles.go
+++ b/pkg/combustion/osfiles.go
@@ -1,0 +1,101 @@
+package combustion
+
+import (
+	_ "embed"
+	"fmt"
+	"io"
+	"os"
+	"os/exec"
+	"path/filepath"
+
+	"github.com/suse-edge/edge-image-builder/pkg/fileio"
+	"github.com/suse-edge/edge-image-builder/pkg/image"
+	"github.com/suse-edge/edge-image-builder/pkg/log"
+	"go.uber.org/zap"
+)
+
+const (
+	osFilesComponentName = "os files"
+	osFilesConfigDir     = "os-files"
+	osFilesScriptName    = "19-copy-os-files.sh"
+	osFilesLogFile       = "copy-os-files.log"
+)
+
+var (
+	//go:embed templates/19-copy-os-files.sh
+	osFilesScript string
+)
+
+func configureOSFiles(ctx *image.Context) ([]string, error) {
+	if !isComponentConfigured(ctx, osFilesConfigDir) {
+		log.AuditComponentSkipped(osFilesComponentName)
+		zap.S().Info("skipping os files component, no files provided")
+		return nil, nil
+	}
+
+	if err := copyOSFiles(ctx); err != nil {
+		log.AuditComponentFailed(osFilesComponentName)
+		return nil, err
+	}
+
+	if err := writeOSFilesScript(ctx); err != nil {
+		log.AuditComponentFailed(osFilesComponentName)
+		return nil, err
+	}
+
+	log.AuditComponentSuccessful(osFilesComponentName)
+	return []string{osFilesScriptName}, nil
+}
+
+func copyOSFiles(ctx *image.Context) error {
+	srcDirectory := filepath.Join(ctx.ImageConfigDir, osFilesConfigDir)
+	destDirectory := filepath.Join(ctx.CombustionDir, osFilesConfigDir)
+
+	dirEntries, err := os.ReadDir(srcDirectory)
+	if err != nil {
+		return fmt.Errorf("reading the os files directory at %s: %w", srcDirectory, err)
+	}
+
+	// If the directory exists but there's nothing in it, consider it an error case
+	if len(dirEntries) == 0 {
+		return fmt.Errorf("no files found in directory %s", srcDirectory)
+	}
+
+	logFilename := filepath.Join(ctx.BuildDir, osFilesLogFile)
+	logFile, err := os.Create(logFilename)
+	if err != nil {
+		return fmt.Errorf("creating log file: %w", err)
+	}
+
+	defer func() {
+		if err = logFile.Close(); err != nil {
+			zap.S().Warnf("failed to close copy os-files log file properly: %s", err)
+		}
+	}()
+
+	cmd := generateOSFilesCopyCommand(srcDirectory, destDirectory, logFile)
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("running copy os-files command: %w", err)
+	}
+
+	return nil
+}
+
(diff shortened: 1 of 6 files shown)


pkg/combustion/templates/19-copy-os-files.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+set -euo pipefail
+
+cp -R ./os-files/* /
```
jdob, line 4: Admittedly, it feels a bit weird to shell out for a copy command. My rationale is that:

* I couldn't find a great library that did it, and implementing it in Go is tedious.
  * Even if I found a library, it's one more dependency to run through the legal and auditing process
* We already shell out for a number of things, so it felt less dirty. If this was the first instance of shelling out in EIB, I would have tried harder to avoid it.

e-minguez: Selinux context needs to be maintained I guess... I'm AFK but I think there is a flag on cp for that purpose.
Also what about rsync instead? 😅 (Maybe it just increases the footprint and make things more complex)

ipetrov117: But, don't we already have a [CopyFiles](https://github.com/suse-edge/edge-image-builder/blob/main/pkg/fileio/file_io.go#L80) function in EIB that can do exactly this?

pkg/combustion/osfiles.go
```diff
@@ -0,0 +1,101 @@
...
+	cmd := generateOSFilesCopyCommand(srcDirectory, destDirectory, logFile)
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("running copy os-files command: %w", err)
+	}
+
+	return nil
+}
+
+func generateOSFilesCopyCommand(srcDirectory string, destDirectory string, output io.Writer) *exec.Cmd {
+	cmd := exec.Command("cp", "-R", srcDirectory, destDirectory)
```
ipetrov117, line 85: Instead of doing this through `exec.Command` isn't it possible to do it with the existing [CopyFiles](https://github.com/suse-edge/edge-image-builder/blob/main/pkg/fileio/file_io.go#L80) function?

jdob: Oh shit, I didn't even realize we had that. Will fix.

pkg/combustion/osfiles.go
```diff
@@ -0,0 +1,89 @@
...
+		return fmt.Errorf("reading the os files directory at %s: %w", srcDirectory, err)
+	}
+
+	// If the directory exists but there's nothing in it, consider it an error case
+	if len(dirEntries) == 0 {
+		return fmt.Errorf("no files found in directory %s", srcDirectory)
+	}
+
+	logFilename := filepath.Join(ctx.BuildDir, osFilesLogFile)
+	logFile, err := os.Create(logFilename)
```
atanasdinov, line 63: This file is never used. Can we remove it?

pkg/combustion/osfiles.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	defer func() {
+		if err = logFile.Close(); err != nil {
+			zap.S().Warnf("failed to close copy os-files log file properly: %s", err)
+		}
+	}()
+
+	if err := fileio.CopyFiles(srcDirectory, destDirectory, "", true); err != nil {
+		return fmt.Errorf("running copy os-files command: %w", err)
```
atanasdinov, line 75:
```suggestion
		return fmt.Errorf("copying os-files: %w", err)
```

pkg/combustion/osfiles_test.go
```diff
@@ -0,0 +1,73 @@
...
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/suse-edge/edge-image-builder/pkg/image"
+)
+
+func setupOsFilesConfigDir(t *testing.T, empty bool) (ctx *image.Context, teardown func()) {
+	ctx, teardownCtx := setupContext(t)
```
atanasdinov, line 14: We have already defined these variables as returned ones so we can just do:
```suggestion
	ctx, teardown = setupContext(t)
```

pkg/combustion/osfiles_test.go
```diff
@@ -0,0 +1,73 @@
...
+	// -- Combustion Script
+	expectedCombustionScript := filepath.Join(ctx.CombustionDir, osFilesScriptName)
+	contents, err := os.ReadFile(expectedCombustionScript)
+	require.NoError(t, err)
+	assert.Contains(t, string(contents), "cp -R")
+
+	// -- Files
+	expectedFile := filepath.Join(ctx.CombustionDir, osFilesConfigDir, "etc", "ssh", "test-config-file")
+	_, err = os.Stat(expectedFile)
+	require.NoError(t, err)
```
atanasdinov, line 57: Nit: These can easily be replaced:
```suggestion
	assert.FileExists(t, expectedFile)
```

pkg/combustion/osfiles_test.go
```diff
@@ -0,0 +1,73 @@
...
+	defer teardown()
+
+	// Test
+	scriptName, err := configureOSFiles(ctx)
+
+	// Verify
+	assert.Nil(t, scriptName)
+
+	srcDirectory := filepath.Join(ctx.ImageConfigDir, osFilesConfigDir)
+	assert.Errorf(t, err, "no files found in directory %s", srcDirectory)
```
atanasdinov, line 72: I assume you want to do the following:
```suggestion
	assert.EqualError(t, err, fmt.Sprintf("no files found in directory %s", srcDirectory))
```

The current `no files found ...` message will only be displayed in case the assertion fails - we're not verifying the actual content of the error with `.Errorf()`

pkg/combustion/osfiles.go
jdob: I copied the line in manually during other changes, but good pick up.

pkg/combustion/osfiles_test.go
jdob: Ah right. I initially started out without the named return values but hit some sort of hiccup and reverted back to that without fixing this.

pkg/combustion/osfiles_test.go
jdob: Oh crap, I feel like I made this mistake elsewhere too. I'll keep an eye out :)

pkg/combustion/osfiles.go
jdob: Yup, its a holdover from when I was shelling out to a script instead of using the Go implementation.

atanasdinov (approved):

Issue #193, opened 2024-02-16
Add support for building image on an aarch64 host machine


jdob:
This is potentially something we are considering for a future release.

e-minguez:
related #68

orangedeng:
Just FYI on this
ISO build success in ubuntu:22.04, aarch64, aws with manual eib built binary.

orangedeng:
There are two problem to run EIB on aarch64:

1. qemu-ipxe must be installed in aarch64. Then guestfish can detect the kvm disabled state.
2. As guestfish requires https://github.com/libguestfs/libguestfs/blob/master/generator/UEFI.ml#L67-L77
```
    "aarch64",
    "/usr/share/AAVMF/AAVMF_CODE.fd",
    Some "/usr/share/AAVMF/AAVMF_CODE.verbose.fd",
    "/usr/share/AAVMF/AAVMF_VARS.fd",
    [];

    "aarch64",
    "/usr/share/edk2/aarch64/QEMU_EFI-pflash.raw",
    None,
    "/usr/share/edk2/aarch64/vars-template-pflash.raw",
    [];
```
  
efi device is required for booting aarch64 vm via qemu-system-aarch64. Which can't be found in opensuse/sles 15.5 packages.
In the ubuntu 22.04, The efi device can be installed by `qemu-efi-aarch64`.

orangedeng:
Marking a aarch working Dockerfile for EIB here: https://github.com/orangedeng/edge-image-builder/blob/add-aarch-build-dockerfile/Dockerfile.arm

There are some problems while the ISO got built:
- OS level, guestfish virtual-tar-out has proformance problem, It takes 20-30 mins to run parsing VM disk to container image. Using machine spec AWS `c6g.2xlarge`. It only use 1 core so `tar -C /sysroot/ -cf - .` takes forever to process.
- Kubernetes/K3s: downloaded wrong arch airgap package, amd64 instead of arm64
- registry: downloaded wrong arch images, amd64 instead of arm64.(hauler binary is fine)

Hope the information above can help.

PR #475, opened 2024-06-12
Implementing aarch64 image build capabilities

rdoxenham:
In this PR I introduce the ability to use an aarch64 system with EIB, that allows the user to generate either RAW or ISO images with the *same* architecture, i.e. it's not possible to cross-build with these patches (e.g. x86_64 image on aarch64, or vice versa) as it's a) not possible to currently use the target architecture for the type of images that we cache, nor the hauler/nmc binary architectures, and b) not possible to currently spin up guestfish with a different qemu architecture to be able to run the image modification tools, but I think this is a good step forward to enabling customers to build aarch64 images at this time.

Note that we will need to rebase on `leap:15.6` for ISO images, as the aarch64 ISO's need a newer version of `xorriso` to be able to repack the aarch64 ISO. I've tested these changes on both aarch64 and x86_64 based systems for both raw and iso (albeit with the previous comment in mind, I tested that pathway on 15.6).

Credit for this largely goes to @dbw7 and @diconico07 for their prior investigation and inputs.

Fixes: #193

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,6 +27,7 @@ FROM opensuse/leap:15.5
 # 4. RPM resolution logic
 # 5. Embedded artefact registry
 # 6. Network configuration
+# 7. RAW image modification on aarch64
 RUN zypper addrepo https://download.opensuse.org/repositories/isv:SUSE:Edge:EdgeImageBuilder/SLE-15-SP5/isv:SUSE:Edge:EdgeImageBuilder.repo && \
     zypper --gpg-auto-import-keys refresh && \
     zypper install -y \
@@ -35,8 +36,21 @@ RUN zypper addrepo https://download.opensuse.org/repositories/isv:SUSE:Edge:Edge
     podman \
     createrepo_c \
     helm hauler \
-    nm-configurator && \
-    zypper clean -a
+    nm-configurator
+
+# Make adjustments for running guestfish and image modifications on aarch64
+# guestfish looks for very specific locations on the filesystem for UEFI firmware
+# and also expects the boot kernel to be a portable executable (PE), not ELF.
+RUN if [[ $(uname -m) == "aarch64" ]]; then \
+	zypper install -y qemu-uefi-aarch64; \
+	mkdir -p /usr/share/edk2/aarch64; \
+	cp /usr/share/qemu/aavmf-aarch64-code.bin /usr/share/edk2/aarch64/QEMU_EFI-pflash.raw; \
+	cp /usr/share/qemu/aavmf-aarch64-vars.bin /usr/share/edk2/aarch64/vars-template-pflash.raw; \
+	mv /boot/vmlinux* /boot/backup-vmlinux; \
+fi
+
+# Clean up the repos to reduce size
+RUN zypper clean -a
 
 COPY --from=0 /src/eib /bin/eib
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,7 +27,6 @@ FROM opensuse/leap:15.5
 # 4. RPM resolution logic
 # 5. Embedded artefact registry
 # 6. Network configuration
-# 7. RAW image modification on aarch64
 RUN zypper addrepo https://download.opensuse.org/repositories/isv:SUSE:Edge:EdgeImageBuilder/SLE-15-SP5/isv:SUSE:Edge:EdgeImageBuilder.repo && \
     zypper --gpg-auto-import-keys refresh && \
     zypper install -y \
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -27,6 +27,7 @@ FROM opensuse/leap:15.5
 # 4. RPM resolution logic
 # 5. Embedded artefact registry
 # 6. Network configuration
+# 7. RAW image modification on aarch64
```
jdob, line 30: This isn't needed here. The intention was to explain the contents of this one big ass RUN line, but the aarch64 stuff is done at a different level after this.

jdob (commented):
Absolutely insane to me how small this patch is. I haven't tested it yet but will shortly. I just added the one quick comment about the comments in the Dockerfile.

pkg/build/templates/modify-raw-image.sh.tpl
```diff
@@ -13,6 +13,17 @@ set -euo pipefail
 #
 # Guestfish Command Documentation: https://libguestfs.org/guestfish.1.html
 
+# In x86_64, the default root partition is the third partition
+ROOT_PART=/dev/sda3
+
+# Make the necessarry adaptations for aarch64
+if [[ $(uname -m) == "aarch64" ]]; then
```
dbw7, line 20: I think we should use the variable that’s set to the arch in the image definition file and then template it in as it allows us to remove this if check and just rely on a single variable which is a little bit safer.

pkg/build/templates/modify-raw-image.sh.tpl
```diff
@@ -13,6 +13,17 @@ set -euo pipefail
...
+	if ! test -f /dev/kvm; then
+		export LIBGUESTFS_BACKEND_SETTINGS=force_tcg
```
dbw7, line 22: I personally believe that this should be the default behavior. Getting KVM to work in a container (in my experience) has been very error prone and while investigating the x86 EIB builds, I found that it resorts to “tcg” anyway so my opinion is that we just make use force_tcg as the default and avoid KVM.

dbw7 (commented):
Changes look good, added some comments which I’m curious to hear your thoughts on. The big unknown here is testing this on a variety of builds (especially with RPMs and embedded images) to see what breaks and what doesn’t.

Dockerfile
```diff
@@ -35,8 +36,21 @@ RUN zypper addrepo https://download.opensuse.org/repositories/isv:SUSE:Edge:Edge
...
+RUN if [[ $(uname -m) == "aarch64" ]]; then \
+	zypper install -y qemu-uefi-aarch64; \
+	mkdir -p /usr/share/edk2/aarch64; \
+	cp /usr/share/qemu/aavmf-aarch64-code.bin /usr/share/edk2/aarch64/QEMU_EFI-pflash.raw; \
+	cp /usr/share/qemu/aavmf-aarch64-vars.bin /usr/share/edk2/aarch64/vars-template-pflash.raw; \
+	mv /boot/vmlinux* /boot/backup-vmlinux; \
+fi
+
+# Clean up the repos to reduce size
+RUN zypper clean -a
```
e-minguez, line 53: This needs to be done in the same layer as the installation, otherwise it will still be there.

ipetrov117:
Not sure whether this should be covered by this PR, but have we tested whether the RPM resolution logic works with the suggested implementation?

I did a quick test on a aarch64 Tumbleweed with your changes + an EIB image using `leap:15.6` as base and the RPM resolution logic failed with a `libguestfs` error. Could be doing something wrong though 🤷

Dockerfile
```diff
@@ -38,6 +38,18 @@ RUN zypper addrepo https://download.opensuse.org/repositories/isv:SUSE:Edge:Edge
     nm-configurator && \
     zypper clean -a
 
+# Make adjustments for running guestfish and image modifications on aarch64
+# guestfish looks for very specific locations on the filesystem for UEFI firmware
+# and also expects the boot kernel to be a portable executable (PE), not ELF.
+RUN if [[ $(uname -m) == "aarch64" ]]; then \
+	zypper install -y qemu-uefi-aarch64; \
```
agracey, line 45: How much space does this package take up? If it's trivial, it may make sense to go ahead and install the aarch64 emulator in the x86 image as well to make working on cross complications easier in the future.

ipetrov117:
> Not sure whether this should be covered by this PR, but have we tested whether the RPM resolution logic works with the suggested implementation?

Tested RPM resolution logic on an arm system using arm SLE Micro. Image was built successful.

Dockerfile
atanasdinov: This is something we already do in the downstream builds.

pkg/build/templates/modify-raw-image.sh.tpl
atanasdinov: I vote for doing this too - we should be able to simplify the script and move everything to Go code.

atanasdinov:
Resolved in https://github.com/suse-edge/edge-image-builder/pull/490.

PR #480, opened 2024-07-01
Extract artifacts config

### LINKED ISSUE #370: [RFE] Add a setting to deploy downstream/upstream embedded bits
Currently EIB deploys metallb and e-c-o from upstream charts (suse-edge/charts) if the container image is being built by the user but if using the downstream EIB image, metallb and e-c-o (something else?) are deployed using downstream charts (IBS).
This is ok but if someone wants to run a test with downstream bits based on a PR here or while the changes haven't been backported, it is currently not possible.

### LINKED ISSUE #419: Make component Helm charts configurable at build time
Names and versions of Helm charts can change and those are currently being [hardcoded](https://github.com/suse-edge/edge-image-builder/blob/main/pkg/combustion/helm.go). We need to provide the option to configure them either by extracting global variables which can be modified at build time or via a configuration file of some sort.

atanasdinov:
Introduces a new configuration file (`artifacts.yaml`) which will serve to contain all options for the artefacts deployed via EIB. The file will be shipped with each release in both upstream and downstream builds. ~~It is possible to override these values via the newly introduced CLI args.~~

~~Listing all options:~~
```
$ podman run --rm --privileged -it -v .:/eib eib:dev build -h
NAME:
   eib build - Build new image

USAGE:
   eib build [OPTIONS]

OPTIONS:
   --build-dir value        Full path to the directory to store build artifacts
   --config-dir value       Full path to the image configuration directory (default: "/eib")
   --definition-file value  Name of the image definition file

   Artifact sources

   --elemental-register.repository value        Address of the elemental-register RPM repository
   --elemental-system-agent.repository value    Address of the elemental-system-agent RPM repository
   --endpoint-copier-operator.chart value       Name of the Endpoint Copier Operator Helm chart
   --endpoint-copier-operator.repository value  Address of the Endpoint Copier Operator Helm repository
   --endpoint-copier-operator.version value     Version of the Endpoint Copier Operator Helm chart
   --metallb.chart value                        Name of the MetalLB Helm chart
   --metallb.repository value                   Address of the MetalLB Helm repository
   --metallb.version value                      Version of the MetalLB Helm chart
```

~~Building an image with custom source values:~~

```
$ podman run --rm --privileged -it -v .:/eib eib:dev build --definition-file definition.yaml --metallb.repository oci://registry.suse.com/edge --metallb.chart metallb-chart

$ cat _build/build-Jul01_15-12-10/helm-repo-add.log
command: /usr/bin/helm repo add suse-edge-endpoint-copier-operator https://suse-edge.github.io/charts
"suse-edge-endpoint-copier-operator" has been added to your repositories

$ cat _build/build-Jul01_15-12-10/helm-pull.log
command: /usr/bin/helm pull oci://registry.suse.com/edge/metallb-chart --version 0.14.3 --destination /eib/_build/build-Jul01_15-12-10/helm/metallb-chart
Pulled: registry.suse.com/edge/metallb-chart:0.14.3
Digest: sha256:8d2987490f7321b3271c950525a203ac4c5b54520fc22c8d5e5cd17e182b2990
command: /usr/bin/helm pull suse-edge-endpoint-copier-operator/endpoint-copier-operator --version 0.2.0 --destination /eib/_build/build-Jul01_15-12-10/helm/endpoint-copier-operator
```

**EDIT:** Customising the default artifact sources will be introduced at a later point when we have more context for a better design decision.

Closes #419


pkg/image/context.go
```diff
@@ -18,4 +18,23 @@ type Context struct {
 	ArtefactsDir string
 	// ImageDefinition contains the image definition properties.
 	ImageDefinition *Definition
+	// ArtifactSources contains the information necessary for the deployment of external artifacts.
+	ArtifactSources *ArtifactSources
+}
+
+type ArtifactSources struct {
```
atanasdinov, line 25: Not convinced this is the best package to have the definition at but it felt weird to keep `env` for it. Thoughts?

jdob: I don't hate it. The image context describes everything needed to build a particular image. In this case, the sources are inputs for that, in the same way the image configuration directory is.

jdob:
I worry about adding each override as an individual flag. I think it has the potential to get out of control, and the flags chosen are pretty verbose (though I totally see why you chose them).

I think we'd be better served with a single flag, something like `--artifacts-override-file` (not wed to that name, but come with me on the concept). In that, the user can use the same structure as the default `artifacts.yaml` file. We can merge the two in a pre-processing step, with any user-specified value taking priority over the defaults.

That has a few effects:
* It makes it simpler for people using overrides. Yes, it means more files, but I think that is a better option -- and less error prone -- than an extremely verbose series of command arguments.
* It puts the configuration into a file which can be version controlled, instead of it existing ephemerally at build time.
* I suspect we're going to need the option to override more things in the future. We've talked about resisting the idea of .eibrc or eib-config.yaml as long as we can, but if we have that need in the future, if we start that model with the artifacts, it'll be consistent if we end up adding the general EIB config file. Admittedly, this is too speculative to use as a rationale in and of itself, but since I think the previous points are justification enough, I'm mentioning this as well.

artifacts.yaml
jdob, line 1: Is there a reason this is in the root and not `/configs`? Admittedly, I'm drawing this from a textbook-level description of a typical Go project, so it's not like I'm speaking from experience. But I prefer to keep root as clean as possible if there's a way to nest it somewhere.

pkg/combustion/helm.go
```diff
@@ -11,40 +10,50 @@ func ComponentHelmCharts(ctx *image.Context) ([]image.HelmChart, []image.HelmRep
 	}
 
 	const (
-		suseEdgeRepositoryName = "suse-edge"
-		installationNamespace  = "kube-system"
+		metallbRepositoryName = "suse-edge-metallb"
+		metallbNamespace      = "metallb-system"
+
+		endpointCopierOperatorRepositoryName = "suse-edge-endpoint-copier-operator"
```
jdob, line 16: The chances of someone defining a helm repository with this name are incredibly small, but do we want to document that certain names are reserved? Or to make it simpler, maybe just a pattern (i.e. "Don't use the prefix `suse-edge` in any of your user-defined Helm repositories"). I am probably overthinking things, so if you think the chances are low enough to ignore this as a problem, I'm not going to die on this hill.

atanasdinov: I was thinking about using an underscore prefix but I felt like it's not worth it. Not only the chances of someone using this are astronomical, but the bigger issue here would be someone trying to install the same Helm chart via the definition file.

I believe documenting this would only bring confusion and distraction and the flow would naturally evolve once we flesh out the "first class citizen" deployments.

artifacts.yaml
atanasdinov: It felt _right_ to have this on the same level as the Dockerfile but I don't mind adjusting it.

jdob (approved):

Issue #483, opened 2024-07-05
Edge image builder cannot find a compatible glibc version


agterdenbos:
I built the edge image builder from the main branch. This was working before. Since today it cannot find a compatible glibc. `eib` fails with the following message:
```
/bin/eib: /lib64/libc.so.6: version `GLIBC_2.38' not found (required by /bin/eib)
/bin/eib: /lib64/libc.so.6: version `GLIBC_2.33' not found (required by /bin/eib)
/bin/eib: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by /bin/eib)
/bin/eib: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by /bin/eib)
error: Recipe `image-config-validate` failed with exit code 1
```
The [Dockerfile](https://github.com/suse-edge/edge-image-builder/blob/550260ab1bfd729a1f9ee14163bae22a5a24745c/Dockerfile) uses `registry.suse.com/bci/golang:1.22` and `opensuse/leap:15.5` as base images. It seems that there is a mismatch in glibc version. But I do not understand why this worked before.

golang:1.22 has glibc 2.38, however, leap 15.5 still has 2.31, which is not a version that eib is looking for:
```
podman run -it registry.suse.com/bci/golang:1.22 ldd --version
ldd (GNU libc) 2.38
```
```
podman run -it opensuse/leap:15.5 ldd --version
ldd (GNU libc) 2.31 
```

The issue might be resolved by upgrading leap to version 15.6
```
podman run -it opensuse/leap:15.6 ldd --version
ldd (GNU libc) 2.38
```

agterdenbos:
@atanasdinov this is related to #476

atanasdinov:
This is weird. Seems like the BCI image was bumped, perhaps pinning to a specific patch version would be better (if possible).

agterdenbos:
Indeed, and this caused a golang image update. That is now based on SLES 15.6. Version 1.22-1.11.6 still has glibc 2.31 because it was based on SLES 15.5.
```
podman run -it registry.suse.com/bci/golang:1.22-1.11.6 ldd --version
ldd (GNU libc) 2.31
```

We can also temporarily downgrade golang to 1.22-1.11.6 if there is less risk.

atanasdinov:
Yes, if you could please downgrade it and confirm that it's working, we should be able to merge the PR and close the issue right away.

agterdenbos:
[...]

PR #486, opened 2024-07-08
Enable Dependabot for Dockerfile

agterdenbos:
Adds a "docker" ecosystem to Dependabot to automatically create pull requests for base container image updates.
Also adds a RUN instruction to the Dockerfile to verify the compatibility between the Golang BCI and the Leap image. Such incompatibilities can occur due to glibc version mismatches, like in #483.


atanasdinov (approved):
Thanks for this contribution!

PR #490, opened 2024-07-10
Arm support

dbw7:
Adds support for building Arm images on an Arm machine

Cross architecture build capabilities (building Arm images on x86_64 and vice versa) will come in a future PR

Closes: #193

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,6 +27,7 @@ FROM opensuse/leap:15.5
 # 4. RPM resolution logic
 # 5. Embedded artefact registry
 # 6. Network configuration
+# 7. RAW image modification on aarch64
 RUN zypper addrepo https://download.opensuse.org/repositories/isv:SUSE:Edge:EdgeImageBuilder/SLE-15-SP5/isv:SUSE:Edge:EdgeImageBuilder.repo && \
     zypper --gpg-auto-import-keys refresh && \
     zypper install -y \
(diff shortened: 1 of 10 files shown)


pkg/rpm/resolver/templates/prepare-tarball.sh.tpl
```diff
@@ -10,6 +10,11 @@ set -euo pipefail
 WORK_DIR={{.WorkDir}}
 IMG_PATH={{.ImgPath}}
 
+# Make the necessarry adaptations for aarch64
+if [[ $(uname -m) == "aarch64" ]]; then
```
atanasdinov, line 14: Don't we want to use the `Arch` variable here too?

Dockerfile
```diff
@@ -38,6 +38,16 @@ RUN zypper addrepo https://download.opensuse.org/repositories/isv:SUSE:Edge:Edge
     nm-configurator && \
     zypper clean -a
 
+# Make adjustments for running guestfish and image modifications on aarch64
+# guestfish looks for very specific locations on the filesystem for UEFI firmware
+# and also expects the boot kernel to be a portable executable (PE), not ELF.
+RUN zypper install -y qemu-uefi-aarch64 && \
```
atanasdinov, line 44: I'd suggest to include the package in the zypper installation layer above. This way we can simplify this to only make the file changes.

RELEASE_NOTES.md
```diff
@@ -8,6 +8,7 @@
 * Kubernetes manifests are now applied in a systemd service
 * Artifact sources origin and metadata are now extracted from a configuration file (`config/artifacts.yaml`)
 * Dropped `-chart` suffix from installed Helm chart names
+* Added ability to build Arm images on an Arm host machine
```
atanasdinov, line 11: Nit: Specifying the exact architecture `aarch64` sounds better to me.

atanasdinov (commented):
Great work!

atanasdinov (approved):

Issue #498, opened 2024-07-16
kernelArgs broken with Leap Micro 6.0


hardys:
I'm trying to build a raw image with a configuration that worked for 5.5 e.g:

```
apiVersion: 1.0
image:
  imageType: raw
  arch: x86_64
  baseImage: openSUSE-Leap-Micro.x86_64-6.0-Default-Build22.15.raw
  outputImageName: openSUSE-Leap-Micro.x86_64-6.0-Default-Build22.15.metal3.raw
operatingSystem:
  kernelArgs:
  - ignition.platform.id=openstack
  - net.ifnames=1
```

I built a local container image based on 058860173952a654f4ada0afc905339b93b07c31 which builds the image without errors but when booting the image we don't see the cmdline args added:

```
controlplane-0:~ # cat /proc/cmdline 
BOOT_IMAGE=/boot/vmlinuz-6.4.0-17-default root=UUID=7429bc86-6e9b-4d30-ad99-4c620001b3fa rd.timeout=60 rd.retry=45 console=ttyS0,115200 console=tty0 security=selinux selinux=1 quiet net.ifnames=0
1
```

It seems this may be a similar issue to https://github.com/suse-edge/edge-image-builder/issues/487 but it's impacting the deployed image, not the install device for the installer iso

hardys:
This was mentioned by @agterdenbos in https://github.com/suse-edge/edge-image-builder/pull/488 - I suspect we'll have to align with a similar sed-fix as an interim solution, but IMO we should consider a more robust approach in future e.g parsing the cmdline args into a map then merging user-specified overrides to render the new cmdline string (this could also resolve https://github.com/suse-edge/edge-image-builder/issues/264 which for some reason was closed without actually fixing anything)

hardys:
I'm not sure we actually need to sed the grub.cfg at all, I think it can be re-generated based on the `/etc/default/grub` e.g:

```
AFAICS we don't have to sed the grub.cfg at all, we can just update /etc/default/grub then re-generate the config via a guestfish shell e.g
guestfish <<EOF
add ${IMAGE}
run
mount /dev/sda3 /
sh "btrfs property set / ro false"
sh "grub2-mkconfig -o /boot/grub2/grub.cfg"
sh "btrfs property set / ro true"
download /boot/grub2/grub.cfg /tmp/grub.cfg.after
exit
EOF
```

```
cat /tmp/grub.cfg.after | grep ignition
[...]

PR #499, opened 2024-07-16
Fix kernelArgs issue with LeapMicro 6.0

hardys:
Due to changes in the grub cmdline the sed approach no longer works, so instead we can run grub2-mkconfig via guestfish which should re-generate the grub.cfg with the defaults applied

Closes: #498

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -20,6 +20,8 @@
 
 ## Bug Fixes
 
+* [#498](https://github.com/suse-edge/edge-image-builder/issues/498) - Fix kernelArgs issue with Leap Micro 6.0
+
 ---
 
 # v1.0.2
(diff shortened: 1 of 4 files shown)


hardys:
I've lightly tested this with both LeapMicro 6.0 and 5.5 - AFAICS it works fine and should hopefully be less fragile than sedding the grub.cfg directly

dbw7 (dismissed):
I tested this with a fairly complex configuration using SL Micro 6.0 and it seems to be working. I'm fine with approving this as is

Can you update the release notes as well to mention the bug fix?

hardys:
> I tested this with a fairly complex configuration using SL Micro 6.0 and it seems to be working. I'm fine with approving this as is
> 
> Can you update the release notes as well to mention the bug fix?

Thanks for the review - release note commit added!

dbw7 (approved):

rdoxenham (approved):

atanasdinov (approved):

hardys:
3 approvals so I'll merge this after discussion with @atanasdinov - thanks for the reviews!

PR #500, opened 2024-07-18
Dockerfile: Switch to Leap-15.6 repo

hardys:
Using SLE15-SP5 with Leap 15.6 may result in dependency issues, so I added a repo to OBS which builds packages Leap-15.6

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,7 +27,7 @@ FROM opensuse/leap:15.6
 # 4. RPM resolution logic
 # 5. Embedded artefact registry
 # 6. Network configuration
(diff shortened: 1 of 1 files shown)


atanasdinov (approved):
Can we please ensure that we're able to add the aarch64 arch and that the artefacts for it are properly built?

hardys:
> Can we please ensure that we're able to add the aarch64 arch and that the artefacts for it are properly built?

Thanks - that was a mistake, aarch packages are now present (thanks for fixing that!) https://download.opensuse.org/repositories/isv:/SUSE:/Edge:/EdgeImageBuilder/Leap-15.6/

PR #505, opened 2024-07-26
Container Image Caching

### LINKED ISSUE #479: Implement caching for container images
Consecutive builds should reuse the container images from a cache directory similarly to how we store the Kubernetes artefacts.

dbw7:
Closes #479 

Container images for air-gapping manifests and Helm charts are now cached to greatly decrease build times.

### DIFF
--- a/pkg/combustion/combustion.go
+++ b/pkg/combustion/combustion.go
@@ -62,7 +62,6 @@ type Combustion struct {
 	RPMResolver                  rpmResolver
 	RPMRepoCreator               rpmRepoCreator
 	Registry                     embeddedRegistry
-	CacheDir                     string
 }
 
 // Configure iterates over all separate Combustion components and configures them independently.
--- a/pkg/combustion/combustion.go
+++ b/pkg/combustion/combustion.go
@@ -62,6 +62,7 @@ type Combustion struct {
 	RPMResolver                  rpmResolver
 	RPMRepoCreator               rpmRepoCreator
 	Registry                     embeddedRegistry
+	CacheDir                     string
 }
 
 // Configure iterates over all separate Combustion components and configures them independently.
--- a/pkg/combustion/registry.go
+++ b/pkg/combustion/registry.go
@@ -2,6 +2,7 @@ package combustion
 
 import (
 	_ "embed"
+	"errors"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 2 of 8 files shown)


pkg/combustion/combustion.go
```diff
@@ -62,6 +62,7 @@ type Combustion struct {
 	RPMResolver                  rpmResolver
 	RPMRepoCreator               rpmRepoCreator
 	Registry                     embeddedRegistry
+	CacheDir                     string
```
atanasdinov, line 65: Pretty sure this is unused. Can we remove it?

pkg/combustion/registry.go
```diff
@@ -227,22 +233,50 @@ func populateRegistry(ctx *image.Context, images []string) error {
...
 	return nil
 }
+
+func imageExists(imagePath string) (bool, error) {
+	if _, err := os.Stat(imagePath); err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return false, nil
+		}
+
+		return false, fmt.Errorf("looking for cached image: %w", err)
```
atanasdinov, line 278: We can log these and just return a boolean. This will simplify the flow.

pkg/combustion/registry.go
```diff
@@ -227,22 +233,50 @@ func populateRegistry(ctx *image.Context, images []string) error {
...
+		} else {
+			if err = storeImage(img, arch, logFile); err != nil {
+				return fmt.Errorf("adding image to registry store: %w", err)
+			}
+
+			if err = generateRegistryTar(imageTarDest, logFile); err != nil {
+				return fmt.Errorf("generating registry store tarball: %w", err)
+			}
+
+			if err = fileio.CopyFile(imageTarDest, imageCacheLocation, fileio.ExecutablePerms); err != nil {
```
atanasdinov, line 260: Shouldn't this be stored with non executable permissions?

pkg/combustion/registry.go
```diff
@@ -227,22 +233,50 @@ func populateRegistry(ctx *image.Context, images []string) error {
...
+		}
 		if err = bar.Add(1); err != nil {
 			zap.S().Debugf("Error incrementing the progress bar: %s", err)
 		}
 	}
 
 	return nil
 }
+
+func imageExists(imagePath string) (bool, error) {
```
atanasdinov, line 272: Nothing about this function is related to an image. It only checks whether a file exists. Perhaps it'd be a good addition to the `fileio` package?

dbw7: I think that'd be good

atanasdinov (approved):

PR #513, opened 2024-07-31
Leverage the netip package when dealing with IP addresses

mchiappero:
When configuring Kubernetes with multiple nodes, EIB requires an IP address to be used as a target for load balancing purposes. However, it assumes this address to be in an IPv4 format, resulting in a broken mangling when IPv6 is used instead.

Correct this problem by letting netip package do the handling of IP addresses, hiding version specific issues. This also allows for easier and additional input validation with minimal effort.


mchiappero:
I have added a couple of additional sanity checks, however I don't think it is worth to create additional unit tests for this specific case, but in case let me know and I will add them. It might make sense to validate the function with an IPv6 too though, let me know if you would like it.

pkg/kubernetes/cluster.go
```diff
@@ -196,13 +197,26 @@ func setClusterCNI(config map[string]any) {
 	config[cniKey] = cniDefaultValue
 }
 
-func setClusterAPIAddress(config map[string]any, apiAddress string, port int) {
-	if apiAddress == "" {
-		zap.S().Warn("Attempted to set an empty cluster API address")
+func setClusterAPIAddress(config map[string]any, apiAddress string, port uint16) {
```
atanasdinov, line 200: We must ensure that this function is successful, because if we don't, then a cluster will not be formed at all. The checks below should be part of the [validation logic](https://github.com/suse-edge/edge-image-builder/blob/main/pkg/image/validation/kubernetes.go) instead.

atanasdinov (changes requested):
Thanks for contributing. Unfortunately, we have to move this logic outside of the modified function.

Please ensure that we also update the [release notes](https://github.com/suse-edge/edge-image-builder/blob/main/RELEASE_NOTES.md), as well as the docs referring to this field.

pkg/image/validation/kubernetes_test.go
```diff
@@ -140,33 +140,38 @@ func TestIsKubernetesDefined(t *testing.T) {
...
+				Nodes: []image.Node{
+					{
+						Hostname: "host",
+						Type:     image.KubernetesNodeTypeServer,
+					},
+				},
 			},
 		},
-		`with nodes - no network config`: {
+		`invalid multi-node - no network config`: {
```
mchiappero, line 174: I have moved this test from Node validation to Network validation as I also have removed the node test from there too. I think it makes more sense here, but let me know what you think.

pkg/kubernetes/cluster.go
```diff
@@ -202,7 +203,9 @@ func setClusterAPIAddress(config map[string]any, apiAddress string, port uint16)
 		return
 	}
 
-	config[serverKey] = fmt.Sprintf("https://%s:%d", apiAddress, port)
+	ip, _ := netip.ParseAddr(apiAddress)
```
mchiappero, line 206: It would be nice to store the Addr directly into APIVIP and skip this additional step already done in validation. But I don't know how easy it is with gopkg.in/yaml.v3 to do so and if it's worth it.

atanasdinov: Let's do the parsing in `NewCluster()` and return the error if it fails. Then we can pass it from there.

pkg/image/validation/kubernetes.go
```diff
@@ -44,21 +46,49 @@ func isKubernetesDefined(k8s *image.Kubernetes) bool {
...
 
-	numNodes := len(k8s.Nodes)
-	if numNodes <= 1 {
-		// Single node cluster, node configurations are not required
 		return failures
 	}
 
-	if k8s.Network.APIVIP == "" {
+	ip, err := netip.ParseAddr(vipAddress)
+	if err != nil {
```
atanasdinov, line 66: Can we also log the error to the log file via `zap.S().Errorf`?

pkg/image/validation/kubernetes.go
```diff
@@ -44,21 +46,49 @@ func isKubernetesDefined(k8s *image.Kubernetes) bool {
...
+		msg := fmt.Sprintf("Invalid cluster API address ('%s')", strings.ToLower(vipAddress))
 		failures = append(failures, FailedValidation{
-			UserMessage: "The 'apiVIP' field is required in the 'network' section when defining entries under 'nodes'.",
+			UserMessage: msg,
```
atanasdinov, line 69: We must return after this otherwise we'll run into nil pointer dereference below.

pkg/image/validation/kubernetes_test.go
```diff
@@ -140,33 +140,38 @@ func TestIsKubernetesDefined(t *testing.T) {
 	assert.False(t, result)
 }
 
-func TestValidateNodes(t *testing.T) {
+func TestValidateNetwork(t *testing.T) {
+	var validV4Network = validNetwork
```
atanasdinov, line 144: Nit: Why do we copy the value instead of just using it?

pkg/image/validation/kubernetes.go
```diff
@@ -44,21 +46,49 @@ func isKubernetesDefined(k8s *image.Kubernetes) bool {
...
 		})
 	}
 
+	if ip.IsUnspecified() || ip.IsMulticast() {
```
atanasdinov, line 73: I'm wondering if these two should be separate and include the actual issue in the validation message e.g. `Cluster API address cannot be a multicast address`. What do you think?

atanasdinov (commented):
Thanks for looking into this. Happy to approve after we resolve these comments.

pkg/kubernetes/cluster.go
mchiappero: This call cannot fail, the value has been validated already, isn't it? By parsing in NewCluster(), do you mean at the unmarshalling stage? Otherwise I can't see how it would be better.

atanasdinov: This is indeed more of a sanity check. While the value has been validated earlier, it has been done in a very different part of the code. There's no guarantee that future changes will not invalidate the flow we're establishing now.

If we ignore the error, then we risk "successfully" building an image that will fail during boot. And we want to limit the chances of such scenarios as much as possible.

Doing something as simple as the following in `NewCluster()` would be more than enough to prevent unexpected outcomes:
```
	virtualIP, err := netip.ParseAddr(kubernetes.Network.APIVIP)
	if err != nil {
		return nil, fmt.Errorf("parsing kubernetes vip address: %w", err)
	}
```

mchiappero: > This is indeed more of a sanity check. While the value has been validated earlier, it has been done in a very different part of the code. 

I'm new to the codebase and I might very well miss something... but I'm quite puzzled by this fact: we have an input validation we cannot rely on? What would the different coda paths be that can lead to a non validated input here? And/or would it be feasible to just store Addr in the Network struct directly?

> There's no guarantee that future changes will not invalidate the flow we're establishing now.

I'm afraid I don't understand what you mean here and how this is related to the above.

> Doing something as simple as the following in `NewCluster()` would be more than enough to prevent unexpected outcomes:

But then you end up with:

`func setMultiNodeConfigDefaults(kubernetes *image.Kubernetes, **address netip.Addr**, config map[string]any)`

despite having `*image.Kubernetes`, right? If so, it's a bit of a code smell.

By the way, I would like to check with you if the VIP is actually optional or rather forbidden in a single host scenario. I'm not sure I fully understand the code, but shouldn't `setClusterAPIAddress()` also be invoked when the value is set in a single host scenario?

atanasdinov: > but I'm quite puzzled by this fact: we have an input validation we cannot rely on? 

We, as users, should of course rely on the validation logic. However, the code itself shouldn't. Each statement should be responsible for achieving its intended result.

We're not talking about duplicating a validation here. We're just ensuring that the specific package and function can behave as expected regardless of its input. If we're unable to parse a string into an IP address, then the received value is invalid, right?

> What would the different coda paths be that can lead to a non validated input here?

We don't know the future. We can just ensure that the code will not produce an erroneous image build regardless of how input validation and modification changes.

> And/or would it be feasible to just store Addr in the Network struct directly?

We can't. At least not in a pretty way. I wouldn't want to duplicate a struct only for this use case, and modifying it in place is also not the way to go here. We're only parsing YAML into it. If we have specific and more sophisticated transformations in the future, then a separate struct would make sense.

> But then you end up with ... despite having *image.Kubernetes, right? If so, it's a bit of a code smell.

One could argue that ignoring an error that could possibly result into an image failing to bootstrap a cluster is a code smell itself. I don't see anything inherently wrong with passing additional argument to a function, though.

> By the way, I would like to check with you if the VIP is actually optional or rather forbidden in a single host scenario. I'm not sure I fully understand the code, but shouldn't setClusterAPIAddress() also be invoked when the value is set in a single host scenario?

It is not allowed to store the value as a [server flag](https://docs.rke2.io/reference/server_config#cluster), however, the purpose of this field is not only limited to this.

It enables the deployment of MetalLB and configuring a virtual IP address which can then later be used from within and outside the cluster. We ensure that it is recognised by the [TLS cert](https://github.com/suse-edge/edge-image-builder/blob/main/pkg/kubernetes/cluster.go#L141). Perhaps these cases may be very rare but we haven't considered this a forbidden scenario. What do you think?

atanasdinov: On the last note, we also disable ServiceLB and use MetalLB in its place for k3s when a VIP is provided. For reference: https://docs.k3s.io/networking/networking-services#disabling-servicelb

mchiappero: > We, as users, should of course rely on the validation logic. However, the code itself shouldn't. Each statement should be responsible for achieving its intended result.
> 
> We're not talking about duplicating a validation here. We're just ensuring that the specific package and function can behave as expected regardless of its input. If we're unable to parse a string into an IP address, then the received value is invalid, right?
> 
> We don't know the future. We can just ensure that the code will not produce an erroneous image build regardless of how input validation and modification changes.
> 
> We can't. At least not in a pretty way. I wouldn't want to duplicate a struct only for this use case, and modifying it in place is also not the way to go here. We're only parsing YAML into it. If we have specific and more sophisticated transformations in the future, then a separate struct would make sense.

Ok, let's step back a bit and go more directly to the point I have been trying to understand: what is the design? What are the various aspects and implications of the design? I'd like to understand what the structure looks like to review my assumptions, have an opinion and understand yours.

You basically say "it's a good practice to write functions that validate their input"; yes, pretty much everybody knows that; but still, it depends, we can make different decisions if it makes sense to do so. Say that the in memory representation is immutable or effectively immutable, or that nothing else is supposed to touch it and such a thing is clear and well documented, and the many possible different cases in between that could exist. 
For example, I'd avoid changing the transparent Network struct too and trying to do it was the last thing on my mind despite the question, but it also sucks have to pass around a string or not being able to use the Addr type following the validation step, which suggest the design could be better refined. So what do we do? :)

> > But then you end up with ... despite having *image.Kubernetes, right? If so, it's a bit of a code smell.
> 
> One could argue that ignoring an error that could possibly result into an image failing to bootstrap a cluster is a code smell itself. I don't see anything inherently wrong with passing additional argument to a function, though.

Sure, but we could argue many other things too, I could argue that the Network struct (or the whole definition) should have encapsulation or that it shouldn't contain a string before making this change; I could argue that `setMultiNodeConfigDefaults` should only receive Version and Network according to the least privilege principle; I could argue that if we have proper validation and unit tests chances are incredibly slim if any... The review is a space to consider and weight them here. I'm not against your change, but are we really sure it's the best decision? Are you sure about it?

Personal opinion, passing an additional parameter to share something that is already available through another parameter is indicative of something off with either the function API design or the code design. Again, what do we do? We don't have time to address the rest, ok, what is a reasonable compromise?

> > By the way, I would like to check with you if the VIP is actually optional or rather forbidden in a single host scenario. I'm not sure I fully understand the code, but shouldn't setClusterAPIAddress() also be invoked when the value is set in a single host scenario?
> 
> It is not allowed to store the value as a [server flag](https://docs.rke2.io/reference/server_config#cluster), however, the purpose of this field is not only limited to this.
> 
> It enables the deployment of MetalLB and configuring a virtual IP address which can then later be used from within and outside the cluster. We ensure that it is recognised by the [TLS cert](https://github.com/suse-edge/edge-image-builder/blob/main/pkg/kubernetes/cluster.go#L141). Perhaps these cases may be very rare but we haven't considered this a forbidden scenario. What do you think?

I have no opinion, but I'd say that if it is specified it should be honored even for a single node clusters (which could grow?). That said, the main reason for asking is this test:
https://github.com/suse-edge/edge-image-builder/blob/ipv6/pkg/image/validation/kubernetes.go#L55
Does it actually make sense? If not I will change it.

Thank you!

atanasdinov: I believe we're going too far with this discussion. We shouldn't be blocking a PR review over whether an argument should be passed to a function or not.

I basically want to ensure that a failure is caught. As for reasonable compromise, I guess just throw a `panic("invalid kubernetes vip address")` in the case an error is raised. It would never be thrown in 99.99% of the cases and even if it is, we're able to recover from it gracefully.

As far as design decisions go - both me, you and everyone else understands that the bigger the codebase, the easier it is to miss guidelines which are not enforced by a linter or other static analysing tool. Just documenting something doesn't mean that it will be followed. Not to mention we'd have to keep more documentation up to date and people would very rarely (if ever) revisit it.

As far as coding principles go - that's a very subjective take. Yes, none of these functions utilise the `Helm` or `Manifests` subfields of the `Kubernetes` struct but we're still only passing a single pointer, as opposed to a `version` string and a pointer to a `Network` struct. I'm fine with both and don't see major issues with either.

Just note that I'm not claiming that the code is perfect. No code is. We're not trying to follow the "best" coding practices that are out there - we're trying to create and maintain a stable product. And dare I say some of the commonly accepted good practices should not be followed all the time. :)

On the single node scenario - yes, let's keep it. This is intended and I missed the k3s LoadBalancer justification which I briefly mentioned above.

mchiappero: To be clear, I have never meant neither to withdraw the PR nor to avoid applying the changes if needed. I wanted to make an attempt to bring the review to the topics that matter to me, to complete the change and be happy to sign-off, which had/have not been addressed.

I am going to push the changes and address the rest of comments later today.

pkg/image/validation/kubernetes.go
mchiappero: The rationale is to exclude anything that has special meaning and allow all the rest of the space, not just public/private address spaces. It is not that useful to distinguish them, and after adding a couple more groups and checking the source code of IsGlobalUnicast() I have realized it does exactly that. Initially I thought it only included IANA defined GlobalUnicast addresses for IPv6. Bottom line, we can just say that it's not a unicast address, it's more that understandable to people dealing with networking.

pkg/image/validation/kubernetes_test.go
mchiappero: it's actually a leftover, as I initially had v4 & v6 variables. I have removed it, I'd change `validNetwork ` to `validV4Network `though.

pkg/image/validation/kubernetes.go
mchiappero: Good catch, it got lost during the copy to validation!

pkg/image/validation/kubernetes.go
```diff
@@ -44,19 +46,53 @@ func isKubernetesDefined(k8s *image.Kubernetes) bool {
...
-	if k8s.Network.APIVIP == "" {
+	ip, err := netip.ParseAddr(vipAddress)
+	if err != nil {
+		msg := fmt.Sprintf("Invalid cluster API address ('%s')", strings.ToLower(vipAddress))
 		failures = append(failures, FailedValidation{
-			UserMessage: "The 'apiVIP' field is required in the 'network' section when defining entries under 'nodes'.",
+			UserMessage: msg,
 		})
+
+		zap.S().Error(msg)
```
atanasdinov, line 72: Sorry for misleading you here - it's been a while since I took a look at this portion. I only envisioned we log the error value itself in order not to lose it, but in order to do that we can just populate the `Error` field of the `FailedValidation` and can get rid of this log here.

pkg/image/validation/kubernetes.go
```diff
@@ -44,19 +46,53 @@ func isKubernetesDefined(k8s *image.Kubernetes) bool {
...
+		return failures
+	}
+
+	if !ip.IsGlobalUnicast() {
+		msg := fmt.Sprintf("Invalid non-unicast cluster API address (%s)", ip.String())
+		failures = append(failures, FailedValidation{
+			UserMessage: msg,
+		})
+
+		zap.S().Error(msg)
```
atanasdinov, line 83: Here we don't have an actual error value but this message is going to be logged both to the console, as well as the log file so this statement is not necessary.

pkg/kubernetes/cluster.go
```diff
@@ -196,13 +197,18 @@ func setClusterCNI(config map[string]any) {
...
+func setClusterAPIAddress(config map[string]any, apiAddress string, port uint16) {
 	if apiAddress == "" {
 		zap.S().Warn("Attempted to set an empty cluster API address")
 		return
 	}
 
-	config[serverKey] = fmt.Sprintf("https://%s:%d", apiAddress, port)
+	ip, err := netip.ParseAddr(apiAddress)
+	if err != nil {
+		panic("Invalid Kubernetes VIP address")
```
atanasdinov, line 208: Can we add a unit test covering this scenario?

atanasdinov (commented):
LGTM! Happy to approve once these minor comments are resolved.

PR #514, opened 2024-07-31
Combustion Cleanup

dbw7:
This PR adds handling for removing leftover combustion artifacts during the first boot of the built image.


pkg/combustion/cleanup.go
```diff
@@ -0,0 +1,37 @@
...
+	"path/filepath"
+
+	"github.com/suse-edge/edge-image-builder/pkg/fileio"
+	"github.com/suse-edge/edge-image-builder/pkg/image"
+	"github.com/suse-edge/edge-image-builder/pkg/log"
+	"go.uber.org/zap"
+)
+
+const (
+	cleanupScriptName    = "99-remove-combustion.sh"
```
atanasdinov, line 16: Nit: I'd suggest `cleanup-combustion` instead.

pkg/combustion/templates/99-cleanup-combustion.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -euo pipefail
+
+rm -r /combustion
+
+if test -d /artefacts; then
+		rm -r /artefacts
```
atanasdinov, line 7: Can we improve the indentation here?

pkg/combustion/cleanup.go
dbw7: It's funny, I force pushed this change right before you commented this because I saw it and thought the same thing

pkg/combustion/templates/99-cleanup-combustion.sh
dbw7: I have no idea why it does that, it looked like 1 tab in Goland but I just fixed it

atanasdinov: hint: you should be able to configure what a tab does in the settings

pkg/combustion/cleanup.go
jdob: I wonder if we want to use something like `zz-cleanup-combustion.sh` instead of numbers. We currently don't force users into our numbering prefix scheme, but `99-` will run before any non-number-prefixed script, so anyone with custom scripts is gonna have a bad time when combustion is deleted before their script runs.

jdob: Actually, we may want to treat this entirely outside of the alphabetical scheme and instead of using the list of scripts to reference from the Combustion `script` file (worst name ever), we just append this to the end of that list regardless.

pkg/combustion/cleanup.go
```diff
@@ -0,0 +1,37 @@
...
+const (
+	cleanupScriptName    = "99-cleanup-combustion.sh"
+	cleanupComponentName = "cleanup"
+)
+
+//go:embed templates/99-cleanup-combustion.sh
+var cleanupScript string
+
+func configureCleanup(ctx *image.Context) ([]string, error) {
+	if ctx.ImageDefinition.Image.ImageType != "raw" {
```
jdob, line 24: Can you compare this to the constant `TypeRAW` in `definition` instead? I prefer to avoid inline constants like this.

jdob (commented):
We spoke about it in chat, but I think the `99-` prefix isn't going to work out. It'll be annoying to circumvent the rest of the structure we set up for adding combustion scripts in order to append it at the end, but I think it's worth it.

jdob (changes requested):
(see my previous comment; I forgot to select "Request changes")

pkg/combustion/combustion.go
```diff
@@ -163,6 +163,17 @@ func (c *Combustion) Configure(ctx *image.Context) error {
 		combustionScripts = append(combustionScripts, scripts...)
 	}
 
+	// We manually add the cleanup component as to always make sure it is last
+	cleanupComponent := componentWrapper{
```
jdob, line 167: It feels a bit overkill to mimic the wrapper struct we use for the full list of them. Am I missing something, or can you simplify it to:

```
s, err := configureCleanup(ctx)
if err != nil {
    return fmt.Errorf("configuring cleanup component %q: %w", cleanupComponentName, err)
}
combustionScripts = append(combustionScripts, s...)
```

dbw7: That works

pkg/combustion/cleanup.go
```diff
@@ -0,0 +1,37 @@
...
+
+const (
+	cleanupScriptName    = "cleanup-combustion.sh"
+	cleanupComponentName = "cleanup"
+)
+
+//go:embed templates/cleanup-combustion.sh
+var cleanupScript string
+
+func configureCleanup(ctx *image.Context) ([]string, error) {
```
atanasdinov, line 23: I'm wondering if we should still treat this as a component. Now that it's out of the loop, it's just a file that we write on the filesystem.

jdob: I wondered this too but didn't have strong opinions in either direction, for whatever it's worth.

jdob (approved):

PR #515, opened 2024-08-01
moved API schema version verification out of validation

jdob:
This is the first step towards enabling EIB 1.1 to differentiate between schema versions 1.0 and 1.1. 
* Moved the apiVersion checking out of validation; it needs to occur before that
* Added the supported schema versions to the output of the `eib version` command


pkg/version/version.go
```diff
@@ -3,11 +3,19 @@ package version
...
+	version10 = "1.0"
+	version11 = "1.1"
+)
+
+var SupportedSchemaVersions = []string{version10, version11}
+
 var version string
 
-func GetVersion() string {
+func GetEibVersion() string {
```
atanasdinov, line 18: nit: can we use `EIB` here since it is an acronym?

pkg/cli/build/build.go
```diff
@@ -126,6 +128,15 @@ func parseImageDefinition(configDir, definitionFile string) (*image.Definition,
 
 	imageDefinition, err := image.ParseDefinition(configData)
 	if err != nil {
+		if errors.Is(err, image.ErrorInvalidSchemaVersion) {
+			m := "Invalid schema version specified. This version of Edge Image Builder supports the following schema versions: %s"
+			msg := fmt.Sprintf(m, strings.Join(version.SupportedSchemaVersions, ", "))
+			return nil, &cmd.Error{
+				UserMessage: msg,
+				LogMessage:  msg,
```
atanasdinov, line 136: nit: I'd suggest using the actual error value here, similarly to what we do for other errors.

atanasdinov (approved):

PR #527, opened 2024-08-12
Directory handling in custom/files

### LINKED ISSUE #478: Folders in custom/files
- Now, if you're including some folders in the eib/custom/files, building the image you've got the following error:

```
2024-06-27T12:06:09.709Z	FATAL	build/build.go:102	An error occurred building the image: configuring combustion: configuring component "custom files": copying file to /eib/_build/build-Jun27_12-06-06/combustion/airgap-resources: copying file: write /eib/_build/build-Jun27_12-06-06/combustion/airgap-resources: copy_file_range: is a directory
github.com/suse-edge/edge-image-builder/pkg/cli/build.Run
```

- I think it could be useful to be able to include folders, not only files. For example, for airgap scenarios the customer maybe could want to add some folders to the image

dbw7:
Closes #478


RELEASE_NOTES.md
```diff
@@ -4,6 +4,8 @@
 
 ## General
 
+* Directories and their contents placed in custom/files are now copied
```
jdob, line 7:
```suggestion
* The "custom files" functionality may now include directories, which will be maintained when copied to the image
```

pkg/combustion/custom.go
```diff
@@ -41,30 +41,52 @@ func configureCustomFiles(ctx *image.Context) ([]string, error) {
...
 	fullScriptsDir := generateComponentPath(ctx, filepath.Join(customDir, customScriptsDir))
 	executablePerms := fileio.ExecutablePerms
-	scripts, err := copyCustomFiles(fullScriptsDir, ctx.CombustionDir, &executablePerms)
+	scripts, err := copyCustomScripts(fullScriptsDir, ctx.CombustionDir, &executablePerms)
 	return scripts, err
 }
 
-func copyCustomFiles(fromDir, toDir string, filePermissions *os.FileMode) ([]string, error) {
+func copyCustomFiles(fromDir, toDir string) error {
+	if _, err := os.Stat(fromDir); os.IsNotExist(err) {
```
jdob, line 56: Is this just a safety check, or is there an expected situation where we'd run into this and want to fail gracefully?

pkg/combustion/custom.go
```diff
@@ -41,30 +41,52 @@ func configureCustomFiles(ctx *image.Context) ([]string, error) {
...
+		return nil
+	}
+
+	dirEntries, err := os.ReadDir(fromDir)
+	if err != nil {
+		return fmt.Errorf("reading the custom files directory at %s: %w", fromDir, err)
+	}
+
+	// If the directory exists but there's nothing in it, consider it an error case
```
jdob, line 65: I don't think we should consider this an error case. It's possible, however incredibly unlikely, that there would be a reason the user would want an empty directory structure (e.g. they are going to run a script that generates something). It's a stretch, but in this case -- as compared to the image config directory sub-directories -- I don't see a huge value in this check.

pkg/combustion/custom.go
```diff
@@ -41,30 +41,52 @@ func configureCustomFiles(ctx *image.Context) ([]string, error) {
...
+	if len(dirEntries) == 0 {
+		return fmt.Errorf("no files found in directory %s", fromDir)
+	}
+
+	if err = fileio.CopyFiles(fromDir, toDir, "", true); err != nil {
+		return fmt.Errorf("copying custom: %w", err)
```
jdob, line 71:
```suggestion
		return fmt.Errorf("copying custom files and directories: %w", err)
```

jdob:
One more thing: can you update building-images.md to mention that directories are also possible now?

pkg/combustion/custom.go
dbw7: This is the clean way to way to just check if the directory exists and proceed if it does

pkg/combustion/custom.go
dbw7: I don't mind this, previously this was here when we would only handle files so I kept it but I can see the situation where they want to maintain the subdirectory structure

dbw7: Scratch that, this check should remain because it checks to make sure the `custom/files` directory itself isn't empty. This aligns with our practice where if any "top" level directory is present but empty we handle it as an error

jdob: Gotcha, that works. Eventually I want to potentially revisit that policy across the board, but better to stay consistent for now.

jdob (approved):

PR #528, opened 2024-08-13
arm64 RKE2 CNI download logic

### LINKED ISSUE #492: Allow the downloads of specific RKE2 arm64 artefacts
At the time of initial Kubernetes development, RKE2 was not shipping arm64 CNI installation tarballs. This has now changed and we need to adjust our [downloading logic](https://github.com/suse-edge/edge-image-builder/blob/main/pkg/kubernetes/artefact_downloader.go#L81).

dbw7:
Closes #492

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -4,6 +4,8 @@
 
 ## General
 
+* Allow RKE2 deployments with Calico, Cilium and Multus on aarch64 platforms
+
 ## API
 
 ### Image Definition Changes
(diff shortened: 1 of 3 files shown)


RELEASE_NOTES.md
```diff
@@ -4,6 +4,8 @@
 
 ## General
 
+* Changed download logic to allow arm64 RKE2 CNI installation tarballs
```
atanasdinov, line 7: I'd suggest a different wording. The target audience for these notes is usually not familiar with the codebase. Something along the lines of "Allow RKE2 deployments with Calico, Cilium and Multus on aarch64 platforms" would be more suitable.

atanasdinov (commented):
Looks great. Let's merge after the release note is adjusted.

PR #530, opened 2024-08-14
Kubernetes version validation

### LINKED ISSUE #469: Add missing Kubernetes version validation
We must ensure Kubernetes version is set when either of the following options is configured:
- manifest URLs
- local manifests
- Helm charts

dbw7:
Closes #469

### DIFF
--- a/pkg/image/validation/kubernetes.go
+++ b/pkg/image/validation/kubernetes.go
@@ -30,6 +30,7 @@ func validateKubernetes(ctx *image.Context) []FailedValidation {
 	var failures []FailedValidation
 
 	if !isKubernetesDefined(&def.Kubernetes) {
+		failures = append(failures, validateAdditionalArtifacts(ctx)...)
 		return failures
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/image/validation/kubernetes.go
```diff
@@ -414,3 +415,35 @@ func validateHelmChartDuplicates(charts []image.HelmChart) string {
 
 	return ""
 }
+
+func validateAdditionalArtifacts(ctx *image.Context) []FailedValidation {
+	var failures []FailedValidation
+
+	if _, err := os.Stat(combustion.KubernetesManifestsPath(ctx)); !os.IsNotExist(err) {
```
atanasdinov, line 422: Let's instead just `ReadDir` directly - this call doesn't make it simpler.

pkg/image/validation/kubernetes.go
```diff
@@ -414,3 +415,35 @@ func validateHelmChartDuplicates(charts []image.HelmChart) string {
...
+		dirEntries, err := os.ReadDir(combustion.KubernetesManifestsPath(ctx))
+		if err != nil {
+			failures = append(failures, FailedValidation{
+				Error: fmt.Errorf("kubernetes manifests directory could not be read: %w", err),
```
atanasdinov, line 426: You can use this description for the UserMessage and only set Error to the err value. Check its usages for reference.

pkg/image/validation/kubernetes_test.go
```diff
@@ -14,7 +18,47 @@ var validNetwork = image.Network{
...
+	defer teardown()
+
+	valuesDir := filepath.Join(ctx.ImageConfigDir, "kubernetes", "helm", "values")
+	err := os.MkdirAll(valuesDir, os.ModePerm)
+	require.NoError(t, err)
+
+	apacheValuesPath := filepath.Join(valuesDir, "apache-values.yaml")
+	err = os.WriteFile(apacheValuesPath, []byte(""), 0o600)
+	require.NoError(t, err)
+
```
atanasdinov, line 61: What's the point of all of this?

pkg/image/validation/kubernetes_test.go
```diff
@@ -14,7 +18,47 @@ var validNetwork = image.Network{
 	APIVIP:  "127.0.0.1",
 }
 
+func setupContext(t *testing.T) (ctx *image.Context, teardown func()) {
```
atanasdinov, line 21: We're barely using any of these directories - can't we just create it inline?

pkg/image/validation/kubernetes_test.go
dbw7: This is the only real way to test the Helm values file logic when it's working and in a non-error case

pkg/image/validation/kubernetes_test.go
dbw7: How about I greatly simplify the function?

atanasdinov (approved):

Issue #537, opened 2024-08-20
Rebuild EIB v1.0.2 and got error when using it


erpz22:
When we git clone the project and checkout v1.0.2 then ran podman build we got a successful eib docker image, then we use it to run it to validate image-config.yaml, the following error popup, when I changed "opensuse/leap:15.5" to "opensuse/leap:15.6" it works.  We did rebuild the EIB tools before and not have those error, so it is kind of error need to be fixed? or we should always use the latest code(that does not make sense)?

errors:
/bin/eib: /lib64/libc.so.6: version `GLIBC_2.38' not found (required by /bin/eib)
/bin/eib: /lib64/libc.so.6: version `GLIBC_2.33' not found (required by /bin/eib)
/bin/eib: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by /bin/eib)
/bin/eib: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by /bin/eib)

atanasdinov:
Thanks for reporting this. It's basically the same issue as https://github.com/suse-edge/edge-image-builder/issues/483, however, we didn't fix it in the `v1.0` branch as we're only using that Dockerfile for dev purposes.

Is there any reason you aren't using the container image we list in the README `registry.opensuse.org/isv/suse/edge/edgeimagebuilder/containerfile/suse/edge-image-builder:1.0.2`?

erpz22:
> Thanks for reporting this. It's basically the same issue as #483, however, we didn't fix it in the `v1.0` branch as we're only using that Dockerfile for dev purposes.
> 
> Is there any reason you aren't using the container image we list in the README `registry.opensuse.org/isv/suse/edge/edgeimagebuilder/containerfile/suse/edge-image-builder:1.0.2`?
The reason to rebuild EIB is to use a newer hauler to build ISO to fix another issue https://github.com/suse-edge/edge-image-builder/issues/531

jdob:
I'm going to close out this issue in favor of tracking it via #531 for the 1.1 release. There are no plans to backport the Hauler version change to the 1.0.z stream.

jdob:
In case you saw my previous comment that I deleted, ignore it, I misread the issue :)

I'll get the go image version fix backported into the `release-1.0` branch. Based on what we saw in #483, I'm going to start testing with the `1.22-1.11.6` version of the image. If you want to take a look at it before I get the fix in, you can try replacing the second line of the Dockerfile with:

```
[...]

PR #538, opened 2024-08-20
made go base image version more explicit

jdob:
It turns out that `golang:1.22` is a rolling release. A recent change/upgrade to this image caused our 1.0.z stream to stop building. This PR pins the go base image to a very specific version which is compatible with the rest of EIB 1.0.z.

closes: #537


jdob:
I did a quick test, fully rebuilding EIB itself and then generating a simple raw image. The build and subsequent boot ran successfully.

atanasdinov (approved):

PR #540, opened 2024-08-23
Schema 1.1 Validation

dbw7:
This is a preliminary attempt at a very simple and temporary solution to the schema validation for the new image definition version.

### DIFF
--- a/pkg/image/validation/validation.go
+++ b/pkg/image/validation/validation.go
@@ -15,6 +15,7 @@ func ValidateDefinition(ctx *image.Context) map[string][]FailedValidation {
 	failures := map[string][]FailedValidation{}
 
 	validations := map[string]validateComponent{
+		versionComponent:  validateVersion,
 		imageComponent:    validateImage,
 		osComponent:       validateOperatingSystem,
 		registryComponent: validateEmbeddedArtifactRegistry,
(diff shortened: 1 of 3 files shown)


pkg/image/validation/validation.go
```diff
@@ -15,6 +15,7 @@ func ValidateDefinition(ctx *image.Context) map[string][]FailedValidation {
 	failures := map[string][]FailedValidation{}
 
 	validations := map[string]validateComponent{
+		versionComponent:  validateVersion,
```
jdob, line 18: My gut reaction was that this is some sort of pre-step before the other validations, since it doesn't make sense to run them if there's a version incompatibility. But seeing this, I think it makes sense. In the future this may lead to some confusing validation failure messages since a failure doesn't stop subsequent checks, but we'll burn that bridge when we get to it.

pkg/image/validation/version.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"github.com/suse-edge/edge-image-builder/pkg/image"
+)
+
+const (
+	versionComponent = "Version"
+)
+
+func validateVersion(ctx *image.Context) []FailedValidation {
+	var failures []FailedValidation
```
jdob, line 12: Might not be a bad idea to drop a comment in here indicating this is a temporary implementation until we find something cleaner. That way someone new to the codebase won't be as confused.

jdob (approved):

jdob (approved):

PR #541, opened 2024-08-23
CopyFiles keep permissions

### LINKED ISSUE #534: Copying files should keep permissions
Files moved via [CopyFiles()](https://github.com/suse-edge/edge-image-builder/blob/main/pkg/fileio/file_io.go#L110) will always default to setting 0644 perms. Investigate all usages, confirm that this default behaviour is not desired and adjust the code to use the input perms instead.

dbw7:
Closes #534

### DIFF
--- a/pkg/fileio/file_io.go
+++ b/pkg/fileio/file_io.go
@@ -77,7 +77,7 @@ func CopyFileN(src io.Reader, dest string, perms os.FileMode, n int64) error {
 //
 // If `copySubDir` is used with 'ext', iterates through all sub-directories
 // and only copies files with the specified extension.
-func CopyFiles(src, dest, ext string, copySubDir bool) error {
+func CopyFiles(src, dest, ext string, copySubDir, maintainPerms bool) error {
 	files, err := os.ReadDir(src)
 	if err != nil {
 		return fmt.Errorf("reading source dir: %w", err)
@@ -97,7 +97,7 @@ func CopyFiles(src, dest, ext string, copySubDir bool) error {
 				continue
 			}
 
-			err = CopyFiles(sourcePath, destPath, ext, true)
+			err = CopyFiles(sourcePath, destPath, ext, true, maintainPerms)
 			if err != nil {
 				return fmt.Errorf("copying files from sub-directory '%s': %w", destPath, err)
 			}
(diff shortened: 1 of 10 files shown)


pkg/fileio/file_io.go
```diff
@@ -107,7 +107,17 @@ func CopyFiles(src, dest, ext string, copySubDir bool) error {
 				continue
 			}
 
-			err := CopyFile(sourcePath, destPath, NonExecutablePerms)
+			perms := NonExecutablePerms
+			if maintainPerms {
```
atanasdinov, line 111: I'm not quite convinced - why do we need this param?

dbw7: So my idea here is that for things that we use internally, we don't need to maintain the original permissions, but for things that the user provides to be on the final image that they manage and we don’t, it should be the original permissions they specified. To this, I thought having that parameter would be a clean solution to allow us to easily specify how we want to use the function wherever we need to. The usage itself though is up for discussion.

atanasdinov: I still think this is unnecessary complexity - none of the files that we transfer are executable in the first place.

agracey: Isn't this for the custom files that a user can place in a tree?

jdob: Another consideration here is that we'll eventually need to maintain SELinux contexts. In that light, I think the approach of "always maintain whatever is on the filesystem" is probably a better approach. If I try to step back as a completely new user, I think that's what I'd expect the behavior to be.

jdob (changes requested):
I think I'm with Atanas here. Let's maintain permissions in every instance. If we find an exception case in the future, we can add it back. Even then, I think the behavior should be more of "maintain permissions by default unless explicit permissions are specified".

pkg/fileio/file_io.go
```diff
@@ -106,12 +106,20 @@ func CopyFiles(src, dest, ext string, copySubDir bool) error {
...
 			}
-			info, infoErr := file.Info()
-			if infoErr != nil {
-				return fmt.Errorf("reading file info %w", infoErr)
+
+			var perms os.FileMode
+			if overridePerms != nil {
+				perms = *overridePerms
+			} else {
+				fileInfo, InfoErr := file.Info()
```
atanasdinov, line 114: Can we revert the upper case for `InfoErr`?

atanasdinov (approved):

atanasdinov (approved):

jdob (approved):

PR #542, opened 2024-08-23
Modify Helm backOffLimit

### LINKED ISSUE #504: Add backoff limit for helm charts
When creating a HelmChart resource we should specify a backoff limit. This value corresponds to the max number of retries in case the installation is unsuccessful. The default value is 1000(!) which means the installation job would have to exit with an error 1000 times before it is considered failed.

Reference: https://docs.k3s.io/helm#helmchart-field-definitions

dbw7:
Closes #504

### DIFF
--- a/pkg/registry/helm_crd.go
+++ b/pkg/registry/helm_crd.go
@@ -10,6 +10,7 @@ const (
 	helmChartAPIVersion = "helm.cattle.io/v1"
 	helmChartKind       = "HelmChart"
 	helmChartSource     = "edge-image-builder"
+	helmBackoffLimit    = 20
 )
 
 type HelmCRD struct {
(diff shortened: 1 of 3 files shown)


pkg/registry/helm_crd.go
```diff
@@ -53,12 +54,14 @@ func NewHelmCRD(chart *image.HelmChart, chartContent, valuesContent, repositoryU
...
 			TargetNamespace string `yaml:"targetNamespace,omitempty"`
 			CreateNamespace bool   `yaml:"createNamespace,omitempty"`
+			BackOffLimit    int    `yaml:"backOffLimit"`
 		}{
 			Version:         chart.Version,
 			ValuesContent:   valuesContent,
 			ChartContent:    chartContent,
 			TargetNamespace: chart.TargetNamespace,
 			CreateNamespace: chart.CreateNamespace,
+			BackOffLimit:    20,
```
atanasdinov, line 64: Can we extract this as a const?

atanasdinov (approved):

atanasdinov (approved):

PR #545, opened 2024-08-23
Adding support for enabling FIPS mode via an operating system boolean

rdoxenham:
In this PR we add support for enabling FIPS functionality with a single boolean in the image definition. In SLE Micro 5.5 it was sufficient to simply add `fips=1` to the kernel cmdline as the default image shipped with the required components, however in SL Micro 6.0 we have to perform some additional steps, at which point it's more convenient to have a single boolean. This boolean takes care of the following steps:

1. Ensures that the `patterns-base-fips` package is installed through RPM dependency resolution
2. Adds the `fips=1` kernel command line argument to the list
3. Runs the FIPS setup script at firstboot to ensure that the FIPS crypto is enabled

The resulting system has the kernel command line option, and persists reboots:

```
node1:~ # cat /proc/cmdline
BOOT_IMAGE=/boot/vmlinuz-6.4.0-17-default root=UUID=b5811ddf-7bcf-4b4d-9b67-c078c9d3657e rd.timeout=60 rd.retry=45 console=ttyS0,115200 console=tty0 security=selinux selinux=1 quiet net.ifnames=0 fips=1
node1:~ # grep fips /boot/grub2/grub.cfg
	linux	/boot/vmlinuz-6.4.0-17-default root=UUID=b5811ddf-7bcf-4b4d-9b67-c078c9d3657e  ${extra_cmdline} console=ttyS0,115200 console=tty0 security=selinux selinux=1 quiet net.ifnames=0 fips=1
		linux	/boot/vmlinuz-6.4.0-17-default root=UUID=b5811ddf-7bcf-4b4d-9b67-c078c9d3657e  ${extra_cmdline} console=ttyS0,115200 console=tty0 security=selinux selinux=1 quiet net.ifnames=0 fips=1
```

It installs the following dependent packages within the `patterns-base-fips`:

```
├── SUSE_Linux_Micro_6.0_x86_64:SL-Micro-6.0-Pool
│   └── x86_64
│       ├── dracut-fips-059+suse.557.g8a62bf73-1.8.x86_64.rpm
│       ├── libkcapi1-1.4.0-1.44.x86_64.rpm
│       ├── libkcapi-tools-1.4.0-1.44.x86_64.rpm
│       ├── libopenssl-3-fips-provider-3.1.4-5.6.x86_64.rpm
│       ├── openssh-fips-9.6p1-1.4.x86_64.rpm
│       └── patterns-base-fips-6.0-1.1.x86_64.rpm
```

It also has the FIPS crypto selection set correctly:

```
node1:~ # cat /proc/sys/crypto/fips_enabled
1

node1:~ # fips-mode-setup --check
FIPS mode is enabled.
Initramfs fips module is enabled.
The current crypto policy (FIPS) is based on the FIPS policy.
```

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -15,6 +15,8 @@
 
 ### Image Definition Changes
 
+* Introduced a dedicated FIPS mode option, adding the required packages, kernel arguments, and crypto selection
+
 ### Image Configuration Directory Changes
 
 ## Bug Fixes
--- a/docs/building-images.md
+++ b/docs/building-images.md
@@ -75,6 +75,7 @@ operatingSystem:
   kernelArgs:
   - arg1
   - arg2
+  enableFIPS: true
   groups:
     - name: group1
     - name: group2
@@ -153,6 +154,7 @@ The remainder of the operating system customizations may be applied regardless o
   parameter is omitted. If this option is set, the default entries will need to be manually added if they are
   still in use.
 * `kernelArgs` - Provides a list of flags that should be passed to the kernel on boot.
+* `enableFIPS` - Specifies whether to setup the operating system for FIPS mode (Default: false)
 * `groups` - Defines a list of operating system groups to create. This will not fail if the 
 group already exists. Each entry is made up of the following fields:
   * `name` - Required; Name of the group to create.
--- a/pkg/combustion/fips.go
+++ b/pkg/combustion/fips.go
@@ -1,0 +1,49 @@
+package combustion
+
+import (
+	_ "embed"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/suse-edge/edge-image-builder/pkg/fileio"
+	"github.com/suse-edge/edge-image-builder/pkg/image"
+	"github.com/suse-edge/edge-image-builder/pkg/log"
+)
+
+const (
+	fipsComponentName = "fips"
+	fipsScriptName    = "15-fips-setup.sh"
+)
+
+var (
+	//go:embed templates/15-fips-setup.sh
+	fipsScript     string
+	FipsPackages   = []string{"patterns-base-fips"}
+	FipsKernelArgs = []string{"fips=1"}
+)
+
+func configureFips(ctx *image.Context) ([]string, error) {
+	fips := ctx.ImageDefinition.OperatingSystem.EnableFips
+	if !fips {
+		log.AuditComponentSkipped(fipsComponentName)
+		return nil, nil
+	}
+
+	if err := writeFipsCombustionScript(ctx); err != nil {
+		log.AuditComponentFailed(fipsComponentName)
+		return nil, err
+	}
+
+	log.AuditComponentSuccessful(fipsComponentName)
+	return []string{fipsScriptName}, nil
+}
+
+func writeFipsCombustionScript(ctx *image.Context) error {
+	fipsScriptFilename := filepath.Join(ctx.CombustionDir, fipsScriptName)
+
+	if err := os.WriteFile(fipsScriptFilename, []byte(fipsScript), fileio.ExecutablePerms); err != nil {
+		return fmt.Errorf("writing file %s: %w", fipsScriptFilename, err)
+	}
+	return nil
+}
--- a/pkg/eib/eib.go
+++ b/pkg/eib/eib.go
@@ -30,6 +30,7 @@ func Run(ctx *image.Context, rootBuildDir string) error {
 	}
 
 	appendElementalRPMs(ctx)
+	appendFips(ctx)
 	appendHelm(ctx)
 
 	c, err := buildCombustion(ctx, rootBuildDir)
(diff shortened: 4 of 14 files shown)


pkg/eib/eib.go
```diff
@@ -107,6 +108,14 @@ func appendElementalRPMs(ctx *image.Context) {
 	}, combustion.ElementalPackages...)
 }
 
+func appendFipsRPMs(ctx *image.Context) {
+	log.AuditInfo("FIPS mode is configured. The necessary RPM packages will be downloaded.")
+	fips := ctx.ImageDefinition.OperatingSystem.Fips
+	if fips {
+		appendRPMs(ctx, nil, "patterns-base-fips")
```
atanasdinov, line 115: Can we extract the name of the RPM as a constant in the combustion package similarly to Elemental ones?

pkg/combustion/fips.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	if err := writeFipsCombustionScript(ctx); err != nil {
+		log.AuditComponentFailed(fipsComponentName)
+		return nil, err
+	}
+
+	kernelArgList := ctx.ImageDefinition.OperatingSystem.KernelArgs
+	kernelArgList = append(kernelArgList, "fips=1")
+	ctx.ImageDefinition.OperatingSystem.KernelArgs = kernelArgList
```
atanasdinov, line 36: Modifications to the input context shouldn't happen here. I'd suggest to:
- rename `appendFipsRPMs()` to `appendFIPS()` in the `eib` package
- extract an `appendKernelArgs()` function accepting variadic string args
- call `appendKernelArgs()` from `appendFIPS()`

atanasdinov (changes requested):
I think we should also enforce validation here. If the kernel args contain `fips=1` only (without the `fips` option being set), then it wouldn't work, right?

RELEASE_NOTES.md
```diff
@@ -12,6 +12,8 @@
 
 ### Image Definition Changes
 
+* Introduced a dedicated FIPS mode option, adding the required packages, kernel arguments, and crypto selection
```
jdob, line 15: When #540 lands, that functionality will need to be updated with this new flag as well.

docs/building-images.md
```diff
@@ -75,6 +75,7 @@ operatingSystem:
   kernelArgs:
   - arg1
   - arg2
+  fips: true
```
jdob, line 78: This looks a bit odd to me, but not awful. @atanasdinov what do you prefer, `fips: true` or `enableFips: true`? I think I like the latter, but it's not a hill I'm gonna die on.

jdob:
Don't forget to update the release notes for the new definition flag too.

jdob (changes requested):
On top of Atanas' concerns about editing the definition, I want to hold off on merging until we've discussed the new definition field name and the schema version validation stuff lands and is integrated (it should be a really simple addition to that validation).

RELEASE_NOTES.md
rdoxenham: Got it... fixed in latest push.

pkg/eib/eib.go
rdoxenham: Yep, done. I've left it as a list, just incase we need to add to it in the future.

docs/building-images.md
rdoxenham: I don't care either way, I've updated so that it's `enableFips`, as it does look a bit cleaner that way.

pkg/image/validation/os.go
```diff
@@ -46,6 +46,12 @@ func validateKernelArgs(os *image.OperatingSystem) []FailedValidation {
 					UserMessage: "Kernel arguments must be specified as 'key=value'.",
 				})
 			}
+
+			if key == "fips" && !os.EnableFips {
```
atanasdinov, line 50: Should we also check the value? Isn't `fips=0` something that might be intentionally set?

docs/building-images.md
```diff
@@ -75,6 +75,7 @@ operatingSystem:
   kernelArgs:
   - arg1
   - arg2
+  enableFips: true
```
atanasdinov, line 78: I'm wondering if we should use capitalise the param e.g. `enableFIPS` - WDYT?

atanasdinov (approved):

docs/building-images.md
rdoxenham: I'd prefer to have it capitalised, but I was just going with other examples :)

pkg/image/validation/os.go
rdoxenham: Hmm, good call.

docs/building-images.md
atanasdinov: My comment is pretty much based on the fact that it's an acronym.

atanasdinov (approved):

docs/building-images.md
jdob: It's tricky. We have

```
httpProxy
sshKeys
```

which are technically acronyms, but we don't capitalize. But we also have:

```
noGPGCheck
```

... should we roll the EIB Decision Dice?

jdob: We spoke about it in chat and let's go with `enableFederalInformationProcessingStandard` to make it simpler

jdob (approved):
I'm good with `enableFIPS`

Issue #276, opened 2024-03-07
Empty directory checks (such as custom scripts/files and rpms) should occur at validation, not build


jdob:
As of 1.0, the logic to check the custom files/scripts directories to see if they are empty occurs in the actual build step instead of validation. There's no real driving reason for this; it was largely missed in the last major validation rework which was primarily focused on the definition itself.

It's not entirely trivial to migrate this into the validation, which exists in the image package and, as previously stated, focuses more on the definition. The logic for what these sorts of directories are live in the combustion-specific component areas and I'm hesitant to introduce that dependency outside of a larger post-1.0 architecture revisit (we already know of a number of things that we need to rethink given new information, in which case this may become moot).

This issue tracks the movement of that logic, along with the logic checking the `rpms` directory for emptiness. The latter directly stems from #242 which will follow the current model the custom files uses when its fix lands.

atanasdinov:
Also applies to `rpms/gpg-keys`.

PR #549, opened 2024-08-28
Elemental validation and improvements when using SL Micro 6.0

dbw7:
This PR changes the way Elemental is handled depending on the API Version of the image definition. The reason being: 1.1 only supports SL Micro 6.0 and 1.0 only supports 5.5 so we can operate under that assumption when working with Elemental.

What happens now is that when using 1.0, the Elemental handling is the same and it uses the repositories found in `artifacts.yaml`, however, when using 1.1, those repositories are no longer used and instead the necessary Elemental packages are found through RPM resolution inside of the SL Micro 6.0 image. This now means that when using Elemental with 6.0, you need to specify a package registration code.

Additionally, to supplement this, I added validation for the Elemental configuration directory and for the registration code.


atanasdinov (changes requested):
Discussed offline - only putting out the conclusions for visibility.

1. Switching up Elemental source requires a change to `artifacts.yaml` which also means some of the code here could (or would) be removed
2. Elemental RPMs shipped with SL Micro 6.0 are not latest yet and we're unable to rely on them at this point in time
3. Validating empty directories should be part of https://github.com/suse-edge/edge-image-builder/issues/276

jdob:
Quick update: The proper RPMs are now available in the SL Micro 6.0 repo, so concern #2 from Atanas' comment is addressed.

pkg/eib/eib.go
```diff
@@ -98,14 +98,19 @@ func appendElementalRPMs(ctx *image.Context) {
...
 
-	appendRPMs(ctx, []image.AddRepo{
-		{
-			URL: ctx.ArtifactSources.Elemental.RegisterRepository,
-		},
-		{
-			URL: ctx.ArtifactSources.Elemental.SystemAgentRepository,
-		},
-	}, combustion.ElementalPackages...)
+	if ctx.ImageDefinition.APIVersion == "1.0" {
```
atanasdinov, line 101: This isn't a good indicator. You can build SL Micro 6.0 images with both API versions 1.0 and 1.1. Do we bite the bullet and only allow Elemental builds for 6.0 in EIB v1.1?

atanasdinov (commented):
I feel like this must land at this point... There's still no response from the ECM team.

pkg/image/validation/elemental_test.go
```diff
@@ -0,0 +1,172 @@
...
+	require.NoError(t, os.MkdirAll(elementalDir, os.ModePerm))
+
+	validElementalConfig := filepath.Join(elementalDir, "elemental_config.yaml")
+	require.NoError(t, os.WriteFile(validElementalConfig, []byte(""), 0o600))
+
+	tests := map[string]struct {
+		ImageDefinition        *image.Definition
+		ExpectedFailedMessages []string
+	}{
+		`valid 1.1`: {
```
atanasdinov, line 32: What does 1.1 indicate here?

pkg/image/validation/elemental_test.go
```diff
@@ -0,0 +1,172 @@
...
+			ImageDefinition: &image.Definition{
+				OperatingSystem: image.OperatingSystem{
+					Packages: image.Packages{
+						RegCode: "registration-code",
+					},
+				},
+			},
+		},
+		`1.1 no registration code`: {
```
atanasdinov, line 41: Same?

pkg/image/validation/elemental_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+			for _, expectedMessage := range test.ExpectedFailedMessages {
+				assert.Contains(t, foundMessages, expectedMessage)
+			}
+
+		})
+	}
+}
+
+func TestValidateElementalConfigDir(t *testing.T) {
```
atanasdinov, line 71: I'd suggest to split this up. It's unreadable at this point and this is not what table tests are suited for.

pkg/image/validation/elemental.go
```diff
@@ -0,0 +1,72 @@
...
+
+	return failures
+}
+
+func validateElementalDir(elementalConfigDir string) []FailedValidation {
+	var failures []FailedValidation
+
+	elementalConfigDirEntries, err := os.ReadDir(elementalConfigDir)
+	if err != nil {
+		failures = append(failures, FailedValidation{
```
atanasdinov, line 48: We need to return after this. No point in trying to make assertions on what's in the directory if we're unable to read it.

pkg/image/validation/elemental_test.go
```diff
@@ -0,0 +1,166 @@
...
+	elementalDir := filepath.Join(configDir, "elemental")
+	require.NoError(t, os.MkdirAll(elementalDir, os.ModePerm))
+
+	failures := validateElementalDir(elementalDir)
+	assert.Len(t, failures, 1)
+
+	var foundMessages []string
+	for _, foundValidation := range failures {
+		foundMessages = append(foundMessages, foundValidation.UserMessage)
+	}
```
atanasdinov, line 112: We don't need all of this for single value assertion.

pkg/image/validation/elemental_test.go
```diff
@@ -0,0 +1,166 @@
...
+	secondElementalConfig := filepath.Join(elementalDir, "elemental_config2.yaml")
+	require.NoError(t, os.WriteFile(secondElementalConfig, []byte(""), 0o600))
+
+	failures := validateElementalDir(elementalDir)
+	assert.Len(t, failures, 1)
+
+	var foundMessages []string
+	for _, foundValidation := range failures {
+		foundMessages = append(foundMessages, foundValidation.UserMessage)
+	}
```
atanasdinov, line 139: Same.

atanasdinov:
Please drop the unused properties from `artifacts.yaml`.

RELEASE_NOTES.md
```diff
@@ -9,6 +9,8 @@
 * Added the ability to automatically copy files into the built images filesystem (see Image Configuration Directory Changes below)
 * Kubernetes manifests are now applied in a systemd service instead of using the `/manifests` directory 
 * Helm chart installation backOffLimit changed from 1000(default) to 20
+* Improved Elemental handling when using SL Micro 6.0
```
atanasdinov, line 12: Can we rephrase this and move it to the API section below? `Elemental configuration now requires a registration code in order to install the necessary RPMs from the official sources` or something similar.

atanasdinov (approved):

PR #564, opened 2024-09-18
RPM installation fix

dbw7:


fdegir:
This was tested with an EIB image containing the fix and confirmed working so LGTM.

rdoxenham:
Fixes #565

PR #589, opened 2024-10-18
IPV6 APIVIP Handling

dbw7:


pkg/image/validation/kubernetes.go
```diff
@@ -117,6 +113,48 @@ func validateNodes(k8s *image.Kubernetes) []FailedValidation {
 	return failures
 }
 
+func validateNetwork(k8s *image.Kubernetes) []FailedValidation {
+	var failures []FailedValidation
+	ip := k8s.Network.APIVIP
```
atanasdinov, line 118: Can we move this next to where it is used?

pkg/image/validation/kubernetes.go
```diff
@@ -117,6 +113,48 @@ func validateNodes(k8s *image.Kubernetes) []FailedValidation {
...
+	}
+
+	if ip == "" {
+		failures = append(failures, FailedValidation{
+			UserMessage: "The 'apiVIP' field is required in the 'network' section when defining entries under 'nodes'.",
+		})
+	}
+
+	if ip != "" {
+		parsedIP, err := netip.ParseAddr(ip)
```
atanasdinov, line 133: Nit: This shows really well why extracting a new variable for something that's easily readable (`k8s.NetworkAPIVIP`) is suboptimal. Now we have to define `parsedIP` instead of using `ip` here which would be simpler, imo.

pkg/image/validation/kubernetes.go
```diff
@@ -117,6 +113,48 @@ func validateNodes(k8s *image.Kubernetes) []FailedValidation {
...
+		if err != nil {
+			failures = append(failures, FailedValidation{
+				UserMessage: fmt.Sprintf("Invalid APIVIP address %q for field 'apiVIP'.", ip),
```
atanasdinov, line 136:
```suggestion
				UserMessage: fmt.Sprintf("Invalid address value %q for field 'apiVIP'.", ip),
```

pkg/image/validation/kubernetes.go
```diff
@@ -117,6 +113,48 @@ func validateNodes(k8s *image.Kubernetes) []FailedValidation {
...
+				Error:       err,
+			})
+		}
+
+		if !parsedIP.Is4() && !parsedIP.Is6() {
+			failures = append(failures, FailedValidation{
+				UserMessage: fmt.Sprintf("%q is not an IPV4 or IPV6 address, only IPV4 and IPV6 addresses are valid for field 'apiVIP'.", ip),
```
atanasdinov, line 143: Can we reduce the repetition?
```suggestion
				UserMessage: "Only IPv4 and IPv6 addresses are valid values for field 'apiVIP'.",
```

pkg/image/validation/kubernetes.go
```diff
@@ -117,6 +113,48 @@ func validateNodes(k8s *image.Kubernetes) []FailedValidation {
...
+		return failures
+	}
+
+	if ip == "" {
+		failures = append(failures, FailedValidation{
+			UserMessage: "The 'apiVIP' field is required in the 'network' section when defining entries under 'nodes'.",
+		})
+	}
+
+	if ip != "" {
```
atanasdinov, line 132: Shouldn't all of the validations below also apply for single node clusters? If the value is there, it should be valid.

pkg/image/validation/kubernetes_test.go
```diff
@@ -15,7 +15,20 @@ import (
 
 var validNetwork = image.Network{
 	APIHost: "host.com",
-	APIVIP:  "127.0.0.1",
+	APIVIP:  "192.168.1.1",
+}
+
+var multiNode = []image.Node{
```
atanasdinov, line 21: This shouldn't be global variable.

pkg/image/validation/kubernetes.go
```diff
@@ -117,6 +113,48 @@ func validateNodes(k8s *image.Kubernetes) []FailedValidation {
...
+		parsedIP, err := netip.ParseAddr(ip)
+		if err != nil {
+			failures = append(failures, FailedValidation{
+				UserMessage: fmt.Sprintf("Invalid APIVIP address %q for field 'apiVIP'.", ip),
+				Error:       err,
+			})
+		}
+
+		if !parsedIP.Is4() && !parsedIP.Is6() {
+			failures = append(failures, FailedValidation{
```
atanasdinov, line 142: Do we return after this? There's no point to confuse users with unicast errors if the IP isn't valid.

pkg/image/validation/kubernetes.go
```diff
@@ -117,6 +113,48 @@ func validateNodes(k8s *image.Kubernetes) []FailedValidation {
...
+	if ip == "" {
+		failures = append(failures, FailedValidation{
+			UserMessage: "The 'apiVIP' field is required in the 'network' section when defining entries under 'nodes'.",
+		})
+	}
+
+	if ip != "" {
+		parsedIP, err := netip.ParseAddr(ip)
+		if err != nil {
+			failures = append(failures, FailedValidation{
```
atanasdinov, line 135: We must return after this.

pkg/image/validation/kubernetes.go
```diff
@@ -117,6 +113,48 @@ func validateNodes(k8s *image.Kubernetes) []FailedValidation {
...
+	ip := k8s.Network.APIVIP
+
+	numNodes := len(k8s.Nodes)
+	if numNodes <= 1 {
+		// Single node cluster, node configurations are not required
+		return failures
+	}
+
+	if ip == "" {
+		failures = append(failures, FailedValidation{
```
atanasdinov, line 127: Returning after this would obsolete the check below and reduce code indentation.

pkg/combustion/templates/rke2-multi-node-installer.sh.tpl
```diff
@@ -91,10 +91,9 @@ systemctl enable kubernetes-resources-install.service
 {{- end }}
 fi
 
-{{- if .apiHost }}
+{{- if and .apiVIP .apiHost }}
 echo "{{ .apiVIP }} {{ .apiHost }}" >> /etc/hosts
 {{- end }}
-
```
atanasdinov, line 97: Why remove this line? 👀

pkg/combustion/templates/rke2-multi-node-installer.sh.tpl
```diff
@@ -91,10 +91,9 @@ systemctl enable kubernetes-resources-install.service
 {{- end }}
 fi
 
-{{- if .apiHost }}
+{{- if and .apiVIP .apiHost }}
```
atanasdinov, line 94: I don't necessarily see an issue with this but if we do it, let's do it in all other kubernetes templates too.

atanasdinov (changes requested):

pkg/combustion/kubernetes_test.go
```diff
@@ -821,3 +821,47 @@ func TestConfigureKubernetes_SuccessfulRKE2ServerWithManifests(t *testing.T) {
...
+	k8s := &image.Kubernetes{
+		Version: "v1.30.3+k3s1",
+		Network: image.Network{
+			APIVIP: "fd12:3456:789a::21",
+		},
+	}
+
+	manifest, err := kubernetesVIPManifest(k8s)
+	require.NoError(t, err)
+	fmt.Println(manifest)
```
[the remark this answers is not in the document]
dbw7: Interesting how the linter didn't pick this up

pkg/image/validation/kubernetes.go
dbw7: A lot of these instances make much more sense in the context of dual-stack handling and I just kept the format while making it pick one or the other but I'll clean everything up.

pkg/combustion/templates/rke2-multi-node-installer.sh.tpl
dbw7: It's already present in all of the other templates, it was just missing from this one for some reason

hardys:
Thanks for working on this @dbw7 - I recall @mchiappero was working on some similar changes in https://github.com/suse-edge/edge-image-builder/pull/513 so perhaps Marco can review to ensure those changes are also captured here

As a next step beyond this, I think we'll need to add dual-stack support too (particularly for the management cluster use-case because I think Rancher only supports dual-stack, not yet single-stack ipv6)

I guess that will require a couple of configuration changes:

- Configure the k8s cluster networks for dual stack e.g https://docs.rke2.io/networking/basic_network_options#dual-stack-configuration
- Configure the APIVIP to support two addresses (perhaps comma delimited like in the RKE2 config above), such that the k8s API is accessible via both address families - I think this will require adding another CIDR to the `addresses` in the MetalLB `IPAddressPool`

atanasdinov:
@hardys We'd indeed want to add dual-stack support - in fact, the first iteration of @dbw7 work was on it. I pushed back on it for now since it's safer (especially with changes to definition file) to do this in multiple steps where the first one is to enable the IPv6 support and have it thoroughly tested (in SV too) before proceeding with dual-stack.

pkg/image/validation/kubernetes.go
```diff
@@ -117,6 +113,48 @@ func validateNodes(k8s *image.Kubernetes) []FailedValidation {
...
 
+func validateNetwork(k8s *image.Kubernetes) []FailedValidation {
+	var failures []FailedValidation
+
+	if k8s.Network.APIVIP == "" {
+		if len(k8s.Nodes) >= 1 {
+			failures = append(failures, FailedValidation{
+				UserMessage: "The 'apiVIP' field is required in the 'network' section when defining entries under 'nodes'.",
+			})
+			return failures
```
atanasdinov, line 124: We can drop this.

pkg/kubernetes/cluster.go
```diff
@@ -54,7 +55,12 @@ func NewCluster(kubernetes *image.Kubernetes, configPath string) (*Cluster, erro
 		return &Cluster{ServerConfig: serverConfig}, nil
 	}
 
-	setMultiNodeConfigDefaults(kubernetes, serverConfig)
+	ip, ipErr := netip.ParseAddr(kubernetes.Network.APIVIP)
```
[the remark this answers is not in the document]
dbw7: Leftover from when it was inside of an if so it was a shadow declaration

atanasdinov (dismissed):

RELEASE_NOTES.md
```diff
@@ -4,6 +4,8 @@
 
 ## General
 
+* Added support for IPV6 addresses in the Kubernetes 'apiVIP' field
```
jdob, line 7:
```suggestion
* Added support for IPv6 addresses in the Kubernetes 'apiVIP' field
```

docs/building-images.md
```diff
@@ -261,7 +261,7 @@ kubernetes:
 * `network` - Required for multi-node clusters, optional for single-node clusters; Defines the network configuration 
 for bootstrapping a cluster.
   * `apiVIP` - Required for multi-node clusters, optional for single-node clusters; Specifies the IP address which
-  will serve as the cluster LoadBalancer, backed by MetalLB.
+  will serve as the cluster LoadBalancer, backed by MetalLB. Supports IPV4 and IPV6 addresses.
```
jdob, line 264:
```suggestion
  will serve as the cluster LoadBalancer, backed by MetalLB. Supports IPv4 and IPv6 addresses.
```

pkg/combustion/kubernetes.go
```diff
@@ -316,11 +317,18 @@ func (c *Combustion) downloadRKE2Artefacts(ctx *image.Context, cluster *kubernet
 }
 
 func kubernetesVIPManifest(k *image.Kubernetes) (string, error) {
+	ip, err := netip.ParseAddr(k.Network.APIVIP)
+	if err != nil {
+		return "", fmt.Errorf("parsing kubernetes APIVIP address: %w", err)
```
jdob, line 322: I feel like this should read as `apiVIP` instead of all caps, since that's what the field is named in the definition. But I don't feel particularly strongly if you decide to keep it like this.

pkg/kubernetes/cluster.go
```diff
@@ -54,7 +55,12 @@ func NewCluster(kubernetes *image.Kubernetes, configPath string) (*Cluster, erro
 		return &Cluster{ServerConfig: serverConfig}, nil
 	}
 
-	setMultiNodeConfigDefaults(kubernetes, serverConfig)
+	ip, err := netip.ParseAddr(kubernetes.Network.APIVIP)
+	if err != nil {
+		return nil, fmt.Errorf("parsing kubernetes APIVIP address: %w", err)
```
jdob, line 60: Same as above. It's an error message and the user-friendly output shows `apiVIP`, so whatever you decide just be consistent between this and the previous message I commented on.

jdob (dismissed):

jdob (approved):

pkg/image/validation/kubernetes.go
```diff
@@ -115,41 +115,38 @@ func validateNodes(k8s *image.Kubernetes) []FailedValidation {
...
 
-	if ip != "" {
-		parsedIP, err := netip.ParseAddr(ip)
-		if err != nil {
-			failures = append(failures, FailedValidation{
-				UserMessage: fmt.Sprintf("Invalid APIVIP address %q for field 'apiVIP'.", ip),
-				Error:       err,
-			})
-		}
+	if !parsedIP.Is4() && !parsedIP.Is6() {
```
mchiappero, line 139: These two checks are redundant, when the address is parsed successfully it's either IPv4 or IPv6. And would also not pass the following following IsGlobaUnicast().

pkg/image/validation/kubernetes_test.go
```diff
@@ -1084,192 +1079,90 @@ func TestValidateAdditionalArtifacts(t *testing.T) {
 	}
 }
 
-func TestValidateNodes(t *testing.T) {
+func TestValidateNetwork(t *testing.T) {
```
mchiappero, line 1069: Hey Danial, glad to see that most of my code has been reused :)
I saw you re-wrote and simplified this function, which on the one hand is good, but has a much smaller coverage. The one I wrote was carefully calibrated around the IsGlobalUnicast() check (and internal sub-checks), so that the different cases of non-global unicast values were sufficiently tested. I think it make sense to include a couple of additional malformed IPs (e.g. 500.168.1.1), but I'd bring back the cases from my initial work as it's just a matter of copying and pasting.

mchiappero (commented):
I'm not sure if I'll have time today to continue my review (by the way, sorry for being late), but in the meantime please have a look at my couple of comments.

Edit: I don't have anything to add. Sorry for not doing a review in a one shot, I wasn't sure I'd have had time but didn't want to hold things until next week, in case. It looks good to me and can't wait to see it included, but I'd address the two main comments right now rather than later.

pkg/combustion/kubernetes_test.go
```diff
@@ -821,3 +821,46 @@ func TestConfigureKubernetes_SuccessfulRKE2ServerWithManifests(t *testing.T) {
 	assert.Contains(t, contents, "name: my-nginx")
 	assert.Contains(t, contents, "image: nginx:1.14.2")
 }
+
+func TestKubernetesVIPManifestValidIPV4(t *testing.T) {
```
mchiappero, line 825: Nit-picking, but I'd align these too names from "IPV4"/"IPV6") to IPv4 as the rest of the code/prints.

Issue #591, opened 2024-10-22
Allow additional module registration during package resolution


atanasdinov:
Currently we only [register](https://github.com/suse-edge/edge-image-builder/blob/4548fa26f56e58b02e54c58133c01a1112485e5c/pkg/rpm/resolver/templates/rpm-resolution.sh.tpl#L15) the default SLE Micro product during package resolution when SCC code is specified. This is not enough to fetch RPMs under additional product modules e.g. `SL-Micro-Extras/6.0/x86_64`.

To achieve this, specifying the following parameter of SUSEConnect should be set:
```
    -p, --product [PRODUCT]  Specify a product for activation/deactivation. Only
                             one product can be processed at a time. Defaults to
                             the base SUSE Linux Enterprise product on this
                             system. Product identifiers can be obtained
                             with `--list-extensions`.
                             Format: <name>/<version>/<architecture>
```

e-minguez:
As a workaround in the meantime:
* Spin up a fresh SUSE Linux Micro with the same version needed
* Register it
* Activate the module/product required, for Extra for example:
```
transactional-update register -p SL-Micro-Extras/6.0/aarch64
```
* Reboot
* Run a temporary and unsupported shell to download the required packages:
```
transactional-update shell -d
...
zypper in --download-only traceroute
exit
```
The packages should be in the `/var/cache/zypp/` folder:
```
find /var/cache/zypp/ -iname "*traceroute*"
/var/cache/zypp/packages/SUSE_Linux_Micro_6.0_aarch64:SL-Micro-6.0-Pool/aarch64/traceroute-2.1.3-1.3.aarch64.rpm
```

jdob:
We discussed this in voice, but to capture the outcome:

The fact that the repo URL contains the version number causes a headache. If it was as simple as saying "SL-Micro-Extras", or even with the arch, we could easily add this into the template. 

The problem comes with the fact that the SL Micro version is in the string. We don't have any way inside of EIB to either detect or have the user specify the version of Micro. 

The initial plan for 1.2 (with a backport to 1.1) was going to be to hardcode the 6.0 string in the RPM resolution and then revisit in the future when we had more time to think of a clean solution. Now it sounds like 6.1 is a requirement for EIB 1.2, which means we need to figure out a more robust solution than we had hoped.

jdob:
[...]

PR #600, opened 2024-10-30
add sl micro 6.0 extras repo for rpm resolution

jdob:
closes: #591 

See #591 for more information. The summary is that EIB 1.2 only needs to support SL Micro 6.0. This patch is a temporary fix that relies on that assumption, hardcoding the version into the repo URL.

Eventually, this will need to be migrated to dynamically introduce the SL Micro version based on the base image. I know how to get that, from `/etc/os-releases`. For the RAW image case, it's somewhat simple: a short guestfish script to extract the file and then some basic parsing for a variable. For ISOs, however, it's more invasive, so we currently unsquash the raw image at rebuild time, meaning it's not available early enough to extract the Micro version. It's fixable, but much more invasive and out of scope for 1.2.

To verify this, try to install the package `cpp`. In SL Micro 6.x, this is no longer in the base repository, now located in the extras repo that is added by this patch.


e-minguez:
I don't think we want to enable this on every build. Honestly I was thinking of enabling a new parameter under the packages section like the repos one but for 'extraProducts'. AFAIK there are at least two, the Extras and the one for SUMA (I don't remember the name exactly).
So users will need to put the list of extraProducts they need to enable and then pass those directly to suseconnect -p

jdob:
There's only one as far as I can see, the extras. That's one reason why I wanted to opt for a simplistic approach for the time being; I'm not sure it's clear how those will shake out, which may lead us to want a more fine grained ability to indicate which to enable. 

Similar to what I said in #591, I'd rather not introduce new definition parameters until we have a better idea of what the end goal is from Micro side. Once an API change like that is out in the wild, we need to continue to support it and deal with a deprecation process, neither of which are particularly fun. IMO, definition additions like this require more information than we have right now to do it correctly.

For example, adding `extraProducts` as a list is excessive if there is only one extra repository, whereas making that a boolean is potentially limiting if they do introduce more than one extra repository. Once their strategy on using these repos becomes more clear, we can make this more configurable.

Let me flip the question around. You said you don't think we want to enable this on every build. Why not? If there's a compelling reason, we may have to revisit this approach, but as of right now I think the tradeoff of enabling it in every build for a less volatile API later is worth it.

e-minguez:
For 5.5 there are at least two, package hub and live patching https://documentation.suse.com/sle-micro/5.5/single-html/SLE-Micro-deployment/#sec-adding-extension so I guess these changes won't work on 5.5
For 6.X I cannot find it right now

dbw7 (dismissed):

atanasdinov (approved):

jdob:
I converted this to draft for now. I'm playing around with adding a definition flag to enable it.

jdob:
Ok, so I tested this and it works. The build will fail if `cpp` is a specified package and the `enableExtras` flag isn't included.

I went with a single flag for the implementation. From the sound of it, I don't anticipate they are going to introduce a ton of these sorts of repos which would necessitate a list. If we do need to deprecate this in the future, it shouldn't be too difficult.

I explicitly did not add any of the validation pieces around the api version in this PR. Once we figure out what's going on with dual stack support, we can add that in a separate PR follow up.

@atanasdinov @dbw7 - What do you guys think?

pkg/rpm/resolver/templates/rpm-resolution.sh.tpl
```diff
@@ -10,9 +10,13 @@ set -euo pipefail
 #  LocalGPGList - list of local GPG keys that will be imported in the resolver image
 #  NoGPGCheck   - when set to true skips the GPG validation for all third-party repositories and local RPMs
 #  Arch         - sets the architecture of the rpm packages to pull
+#  EnableExtras - registers the SL-Micro-Extras repo for use in resolution
 
 {{ if ne .RegCode "" }}
 suseconnect -r {{ .RegCode }}
+{{ if $.EnableExtras -}}
```
atanasdinov, line 17: Nit: The dollar sign shouldn't be necessary.

atanasdinov (approved):
LGTM!

pkg/rpm/resolver/templates/rpm-resolution.sh.tpl
jdob: Admittedly, I didn't even pull up the template syntax docs. I looked at the GPG conditional a few lines below and followed that model. Still, I appreciate the heads up.

atanasdinov (approved):

PR #612, opened 2024-11-14
Dual-stack Handling

dbw7:


pkg/combustion/templates/set-node-ip.sh.tpl
```diff
@@ -0,0 +1,146 @@
...
+        if [ -n "$IPv4_ADDRESS" ]; then
+            echo "Successfully got IPv4 address: $IPv4_ADDRESS"
+            return 0
+        fi
+
+        if [ $attempt -lt $MAX_ATTEMPTS ]; then
+            echo "Waiting $DELAY seconds before next attempt..."
+            sleep $DELAY
+        fi
+        ((attempt++))
```
atanasdinov, line 70: This should come before the attempt count check above.

pkg/combustion/templates/set-node-ip.sh.tpl
```diff
@@ -0,0 +1,146 @@
...
+        if [ -n "$IPv6_ADDRESS" ]; then
+            echo "Successfully got IPv6 address: $IPv6_ADDRESS"
+            return 0
+        fi
+
+        if [ $attempt -lt $MAX_ATTEMPTS ]; then
+            echo "Waiting $DELAY seconds before next attempt..."
+            sleep $DELAY
+        fi
+        ((attempt++))
```
atanasdinov, line 90: Same as above.

pkg/combustion/templates/set-node-ip.sh.tpl
```diff
@@ -0,0 +1,146 @@
...
+}
+
+echo "----------------------------------------"
+echo "Results:"
+if [ "${IPv4}" = "true" ]; then
+    echo "IPv4 Address: ${IPv4_ADDRESS:-Not found}"
+fi
+if [ "${IPv6}" = "true" ]; then
+    echo "IPv6 Address: ${IPv6_ADDRESS:-Not found}"
+fi
```
atanasdinov, line 138: Please see whether we can reuse the printing statements.

pkg/combustion/templates/set-node-ip.sh.tpl
```diff
@@ -0,0 +1,146 @@
...
+
+update_config() {
+    if [ "${IPv4}" = "true" ] && [ "${IPv6}" = "true" ]; then
+        if [ "$prioritizeIPv6" = "false" ]; then
+            echo "node-ip: ${IPv4_ADDRESS},${IPv6_ADDRESS}" >> "$CONFIG_FILE"
+            echo "Added IPv4,IPv6 addresses to config (IPv4 prioritized)"
+        else
+            echo "node-ip: ${IPv6_ADDRESS},${IPv4_ADDRESS}" >> "$CONFIG_FILE"
+            echo "Added IPv6,IPv4 addresses to config (IPv6 prioritized)"
+#            if [ "$IS_SERVER" = "true" ]; then
```
atanasdinov, line 113: If these are no longer necessary, we should remove them.

pkg/combustion/templates/set-node-ip.sh.tpl
```diff
@@ -0,0 +1,146 @@
...
+#              echo "kube-apiserver-arg:" >> "$CONFIG_FILE"
+#              echo "  - \"advertise-address=${IPv6_ADDRESS}\"" >> "$CONFIG_FILE"
+#            fi
+        fi
+    elif [ "${IPv4}" = "true" ]; then
+        echo "node-ip: ${IPv4_ADDRESS}" >> "$CONFIG_FILE"
```
atanasdinov, line 119: Ensure this value is removed even if provided or prioritised.

pkg/combustion/templates/set-node-ip.sh.tpl
```diff
@@ -0,0 +1,146 @@
...
+
+if [ "${IPv6}" = "true" ]; then
+    try_get_ipv6
+fi
+
+update_config() {
+    if [ "${IPv4}" = "true" ] && [ "${IPv6}" = "true" ]; then
+        if [ "$prioritizeIPv6" = "false" ]; then
+            echo "node-ip: ${IPv4_ADDRESS},${IPv6_ADDRESS}" >> "$CONFIG_FILE"
+            echo "Added IPv4,IPv6 addresses to config (IPv4 prioritized)"
```
atanasdinov, line 109:
```suggestion
            echo "Added IPv4 and IPv6 addresses to config (IPv4 prioritized)"
```

RELEASE_NOTES.md
```diff
@@ -4,6 +4,9 @@
 
 ## General
 
+* Added 'apiVIP6' in Kubernetes field for IPv6 addresses
```
atanasdinov, line 7: Figure out the proper places to document this.

pkg/image/validation/kubernetes.go
```diff
@@ -14,10 +16,13 @@ import (
...
-	httpScheme   = "http"
-	httpsScheme  = "https"
-	ociScheme    = "oci"
+	k8sComponent     = "Kubernetes"
+	httpScheme       = "http"
+	httpsScheme      = "https"
+	ociScheme        = "oci"
+	k8sDir           = "kubernetes"
+	k8sConfigDir     = "config"
+	serverConfigFile = "server.yaml"
```
atanasdinov, line 25: Try to reuse the constants defined somewhere else -- we don't want to run into the case of some of these getting modified in one place but not the other.

pkg/combustion/kubernetes.go
```diff
@@ -147,32 +155,34 @@ func (c *Combustion) configureK3S(ctx *image.Context, cluster *kubernetes.Cluste
...
+		"apiVIP6":         ctx.ImageDefinition.Kubernetes.Network.APIVIP6,
 		"apiHost":         ctx.ImageDefinition.Kubernetes.Network.APIHost,
 		"binaryPath":      binaryPath,
 		"imagesPath":      imagesPath,
 		"manifestsPath":   manifestsPath,
-		"configFilePath":  prependArtefactPath(k8sDir),
-		"registryMirrors": prependArtefactPath(filepath.Join(k8sDir, registryMirrorsFileName)),
+		"configFilePath":  prependArtefactPath(K8sDir),
+		"registryMirrors": prependArtefactPath(filepath.Join(K8sDir, registryMirrorsFileName)),
+		"setNodeIPScript": setNodeIPScript,
```
atanasdinov, line 166: Can we get this value from somewhere else? It's better to "tie" it with the rest of the code, where this is a variable that's returned by the function which has created it. Not an arbitrary constant.

pkg/kubernetes/cluster.go
```diff
@@ -288,3 +329,15 @@ func ServersCount(nodes []image.Node) int {
 
 	return servers
 }
+
+func IsIPv6Priority(serverConfig map[string]any) bool {
+	if clusterCIDR, ok := serverConfig["cluster-cidr"].(string); ok {
+		cidrs := strings.Split(clusterCIDR, ",")
+		if len(cidrs) > 0 {
+			fmt.Println(strings.Contains(cidrs[0], ":"))
```
atanasdinov, line 337: Must be removed.

pkg/kubernetes/cluster.go
```diff
@@ -196,13 +230,20 @@ func setClusterCNI(config map[string]any) {
 	config[cniKey] = cniDefaultValue
 }
 
-func setClusterAPIAddress(config map[string]any, apiAddress string, port int) {
-	if apiAddress == "" {
+func setClusterAPIAddress(config map[string]any, ip4 *netip.Addr, ip6 *netip.Addr, port uint16, prioritizeIPv6 bool) {
+	if ip4 == nil && ip6 == nil {
 		zap.S().Warn("Attempted to set an empty cluster API address")
```
atanasdinov, line 235: Let's panic() instead -- we have a graceful handling for it and we shouldn't proceed anyway.

pkg/kubernetes/cluster.go
```diff
@@ -196,13 +230,20 @@ func setClusterCNI(config map[string]any) {
...
 
-	config[serverKey] = fmt.Sprintf("https://%s:%d", apiAddress, port)
+	switch {
+	case ip6 != nil && prioritizeIPv6:
+		config[serverKey] = fmt.Sprintf("https://%s", netip.AddrPortFrom(*ip6, port).String())
+	case ip6 != nil && ip4 == nil:
+		config[serverKey] = fmt.Sprintf("https://%s", netip.AddrPortFrom(*ip6, port).String())
+	default:
+		config[serverKey] = fmt.Sprintf("https://%s", netip.AddrPortFrom(*ip4, port).String())
+	}
```
atanasdinov, line 246: If I had to improve, I'd go with this. But I'm not convinced it should be part of this function. The code is the same for both so the checks could be extracted out of this and only a single address could be passed.

```suggestion
	if ip6 != nil && (prioritizeIPv6 || ip4 == nil) {
		config[serverKey] = fmt.Sprintf("https://%s", netip.AddrPortFrom(*ip6, port).String())
		return
	}

	config[serverKey] = fmt.Sprintf("https://%s", netip.AddrPortFrom(*ip4, port).String())
```

pkg/combustion/templates/set-node-ip.sh.tpl
atanasdinov: This value should:
- be verified upon validation
- cause the build to fail if >1 servers are configured
- not be updated if supplied
- print a warning statement if supplied

pkg/image/validation/kubernetes_test.go
```diff
@@ -116,6 +122,9 @@ func TestValidateKubernetes(t *testing.T) {
 				"Helm chart 'name' field must be defined.",
 				"Helm repository 'name' field for \"apache-repo\" must match the 'repositoryName' field in at least one defined Helm chart.",
 				"Helm chart 'repositoryName' \"another-apache-repo\" for Helm chart \"\" does not match the name of any defined repository.",
+				"Invalid non-unicast cluster API address (127.0.0.1) for field 'apiVIP'.",
+				"Invalid non-unicast cluster API address (ff02::1) for field 'apiVIP6'.",
+				fmt.Sprintf("Kubernetes server config could not be found at '%s,' dualstack configuration requires a defined cluster-cidr and service-cidr.", filepath.Join(configDir, "kubernetes", "config", "server.yaml")),
```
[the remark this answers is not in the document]
dbw7: Can't hardcode it here as the config directory is a temp directory, we could cut the error message short and omit it.

pkg/image/validation/kubernetes.go
```diff
@@ -117,6 +116,326 @@ func validateNodes(k8s *image.Kubernetes) []FailedValidation {
...
+				Error:       err,
+			})
+			return nil, failures
+		}
+		extractedIP = parsedIP
+	}
+
+	if !extractedIP.IsGlobalUnicast() && !hasPrefix {
+		failures = append(failures, FailedValidation{
+			UserMessage: fmt.Sprintf("Kubernetes server config node-ip %s must be a valid unicast IP", ip),
```
atanasdinov, line 426: Shouldn't contain hardcoded values.

pkg/image/validation/kubernetes.go
```diff
@@ -117,6 +116,326 @@ func validateNodes(k8s *image.Kubernetes) []FailedValidation {
...
+			failures = append(failures, FailedValidation{
+				UserMessage: fmt.Sprintf("Kubernetes server config %s value '%s' could not be parsed", configField, ip),
+				Error:       err,
+			})
+			return nil, failures
+		}
+		extractedIP = parsedIP
+	}
+
+	if !extractedIP.IsGlobalUnicast() && !hasPrefix {
```
atanasdinov, line 424: Can be simplified by executed once.

pkg/image/validation/kubernetes.go
```diff
@@ -117,6 +116,326 @@ func validateNodes(k8s *image.Kubernetes) []FailedValidation {
...
+		}
+	}
+
+	return firstCIDRIP.Is6(), failures
+}
+
+func validateNodeIP(k8s *image.Kubernetes, serverConfig map[string]any) []FailedValidation {
+	var failures []FailedValidation
+
+	if nodeIPs, ok := serverConfig["node-ip"].(string); ok {
```
atanasdinov, line 341: We can invert this statement to have a better layout -- no street fighter code!

pkg/image/validation/kubernetes.go
```diff
@@ -117,6 +116,326 @@ func validateNodes(k8s *image.Kubernetes) []FailedValidation {
...
+		if len(k8s.Nodes) > 1 {
+			var serverCount int
+			for _, nodes := range k8s.Nodes {
+				if nodes.Type == "server" {
+					serverCount++
+				}
+			}
```
atanasdinov, line 348: Maybe could be substituted by `kubernetes.ServersCount`.

pkg/combustion/kubernetes.go
```diff
@@ -147,23 +155,26 @@ func (c *Combustion) configureK3S(ctx *image.Context, cluster *kubernetes.Cluste
...
+		"getNodeIP":       kubernetes.GetNodeIP(cluster.ServerConfig),
```
atanasdinov, line 167: Let's drop this and instead make the change where we create the script.

atanasdinov:
A missing piece here is a bump of the `apiVersion`.

atanasdinov (commented):
Another missing piece here is a bump of the apiVersion.

pkg/kubernetes/cluster.go
atanasdinov: Let's do something about this. The whole `ipv4Holder` / `ipv6Holder` pointer shenanigans also doesn't seem worth it to me.

pkg/image/validation/kubernetes_test.go
```diff
@@ -1079,3 +1070,691 @@ func TestValidateAdditionalArtifacts(t *testing.T) {
...
+				},
+			},
+			ExpectedFailedMessages: []string{
+				"Non-unicast cluster API address (0.0.0.0) for field 'apiVIP' is invalid.",
+			},
+		},
+		`undefined v6 VIP`: {
+			K8s: image.Kubernetes{
+				Network: image.Network{
+					APIHost: "host.com",
```
atanasdinov, line 1216: Any reason to set all these API hosts?

pkg/image/validation/kubernetes_test.go
```diff
@@ -1079,3 +1070,691 @@ func TestValidateAdditionalArtifacts(t *testing.T) {
...
+	}}
+
+	failures := validateNetworkingConfig(&k8s, "fake-path")
+
+	assert.Len(t, failures, 1)
+
+	var foundMessages []string
+	for _, foundValidation := range failures {
+		foundMessages = append(foundMessages, foundValidation.UserMessage)
+	}
```
atanasdinov, line 1324: Do we have to do this given there's a single value in the slice?

pkg/image/validation/kubernetes_test.go
```diff
@@ -1079,3 +1070,691 @@ func TestValidateAdditionalArtifacts(t *testing.T) {
...
+}
+
+func TestValidateConfigValidAPIVIPNotConfigured(t *testing.T) {
+	k8s := image.Kubernetes{}
+
+	failures := validateNetworkingConfig(&k8s, "")
+	assert.Len(t, failures, 0)
+}
+
+func TestValidateConfigValidIPv4Prio(t *testing.T) {
```
atanasdinov, line 1336: I believe this is a leftover. Can we double check?

pkg/image/validation/kubernetes_test.go
```diff
@@ -1079,3 +1070,691 @@ func TestValidateAdditionalArtifacts(t *testing.T) {
...
+
+	configFile := filepath.Join(serverConfigDir, "server.yaml")
+	require.NoError(t, os.WriteFile(configFile, []byte(b), 0o600))
+
+	failures := validateNetworkingConfig(&k8s, configFile)
+
+	assert.Len(t, failures, 0)
+}
+
+func mockValidateConfig(k8s *image.Kubernetes, serverConfig map[string]any) []FailedValidation {
```
atanasdinov, line 1368: Mocking means to provide fake data as a result and we do calls to the actual functions we want to test within this function.

We cannot rely on the current implementation to only include the calls to `validateNodeIP` and `validateCIDRConfig` in the future. Removing either of those or adding anything else will render the  `TestValidateConfig` "useless". Let's drop it since we already have separate test tables covering both of these functions separately.

pkg/image/validation/kubernetes_test.go
```diff
@@ -1079,3 +1070,691 @@ func TestValidateAdditionalArtifacts(t *testing.T) {
...
+			ServerConfig: map[string]any{
+				"cluster-cidr": "10.42.0.0/16,FF01::/48",
+				"service-cidr": "10.43.0.0/16,FF01::/112",
+			},
+			ExpectedFailedMessages: []string{
+				"Kubernetes server config cluster-cidr value 'FF01::/48' must be a valid unicast address",
+				"Kubernetes server config service-cidr value 'FF01::/112' must be a valid unicast address",
+			},
+		},
+		`invalid node ip`: {
```
atanasdinov, line 1579: Duplicate of the previous case.

pkg/image/validation/kubernetes_test.go
```diff
@@ -1079,3 +1070,691 @@ func TestValidateAdditionalArtifacts(t *testing.T) {
...
+			},
+		},
+		`single cidr IPv6`: {
+			K8s: image.Kubernetes{},
+			ServerConfig: map[string]any{
+				"cluster-cidr": "fd12:3456:789b::/48",
+				"service-cidr": "fd12:3456:789c::/112",
+			},
+		},
+		`single cidr IPv4`: {
```
atanasdinov, line 1612: We're missing the valid dual-stack cases.

pkg/image/validation/kubernetes_test.go
```diff
@@ -1079,3 +1070,691 @@ func TestValidateAdditionalArtifacts(t *testing.T) {
...
+
+			for _, expectedMessage := range test.ExpectedFailedMessages {
+				assert.Contains(t, foundMessages, expectedMessage)
+			}
+
+		})
+	}
+}
+
+func TestValidateCIDRConfig(t *testing.T) {
```
atanasdinov, line 1449: We're missing the following cases:
- No CIDRs when API VIPs are unset
- More than two addresses in a given CIDR field

pkg/image/validation/kubernetes_test.go
```diff
@@ -1079,3 +1070,691 @@ func TestValidateAdditionalArtifacts(t *testing.T) {
...
+			K8s: image.Kubernetes{
+				Nodes: []image.Node{
+					{
+						Hostname: "server1",
+						Type:     image.KubernetesNodeTypeServer,
+					},
+					{
+						Hostname:    "agent",
+						Type:        image.KubernetesNodeTypeAgent,
+						Initialiser: true,
```
atanasdinov, line 1673: Initialiser cannot be an agent.

pkg/image/validation/kubernetes_test.go
```diff
@@ -1079,3 +1070,691 @@ func TestValidateAdditionalArtifacts(t *testing.T) {
...
+					},
+					{
+						Hostname:    "agent2",
+						Type:        image.KubernetesNodeTypeAgent,
+						Initialiser: true,
```
atanasdinov, line 1678: Initialiser cannot be an agent.

pkg/image/validation/kubernetes_test.go
```diff
@@ -1079,3 +1070,691 @@ func TestValidateAdditionalArtifacts(t *testing.T) {
...
+				"node-ip": "10.42.0.0",
+			},
+		},
+		`single IPv6`: {
+			K8s: image.Kubernetes{},
+			ServerConfig: map[string]any{
+				"node-ip": "fd12:3456:789a::21",
+			},
+		},
+		`multiple nodes valid`: {
```
atanasdinov, line 1663:
```suggestion
		`multiple nodes, one server`: {
```

pkg/image/validation/kubernetes_test.go
```diff
@@ -1079,3 +1070,691 @@ func TestValidateAdditionalArtifacts(t *testing.T) {
...
+						Type:        image.KubernetesNodeTypeAgent,
+						Initialiser: true,
+					},
+				},
+			},
+			ServerConfig: map[string]any{
+				"node-ip": "10.42.0.0",
+			},
+		},
+		`multiple servers invalid`: {
```
atanasdinov, line 1686:
```suggestion
		`multiple nodes, multiple servers`: {
```

pkg/image/validation/kubernetes_test.go
```diff
@@ -1079,3 +1070,691 @@ func TestValidateAdditionalArtifacts(t *testing.T) {
...
+		`node ip IPv6 invalid`: {
+			K8s: image.Kubernetes{},
+			ServerConfig: map[string]any{
+				"node-ip": "xxxx:3456:789a::21",
+			},
+			ExpectedFailedMessages: []string{
+				"Kubernetes server config node-ip value 'xxxx:3456:789a::21' could not be parsed",
+			},
+		},
+	}
```
atanasdinov, line 1740: These two are confusing. The first one is dual-stack and implies we should also test a non-unicast IPv6. The second one is single-stack and implies we should also test an invalid IPv4. Are we missing those cases on purpose?

pkg/image/validation/kubernetes_test.go
dbw7: I left this and the one below it because they are the only ones that actually test that function, but I don't mind removing.

pkg/combustion/kubernetes_test.go
```diff
@@ -556,9 +1114,11 @@ func TestConfigureKubernetes_SuccessfulMultiNodeRKE2Cluster(t *testing.T) {
 	assert.Contains(t, contents, "cp $CONFIGFILE /etc/rancher/rke2/config.yaml")
 	assert.Contains(t, contents, "if [ \"$HOSTNAME\" = node1.suse.com ]; then")
 	assert.Contains(t, contents, "echo \"192.168.122.100 api.cluster01.hosted.on.edge.suse.com\" >> /etc/hosts")
+	assert.Contains(t, contents, "echo \"fd12:3456:789a::21 api.cluster01.hosted.on.edge.suse.com\" >> /etc/hosts")
 	assert.Contains(t, contents, "export INSTALL_RKE2_ARTIFACT_PATH=$ARTEFACTS_DIR/kubernetes/install")
 	assert.Contains(t, contents, "sh $ARTEFACTS_DIR/kubernetes/install-kubernetes.sh")
 	assert.Contains(t, contents, "systemctl enable rke2-$NODETYPE.service")
+	assert.NotContains(t, contents, "sh set-node-ip.sh")
```
atanasdinov, line 1121: I'd suggest to go through all cases and ensure assertion for `set-node-ip` exists.

pkg/combustion/kubernetes_test.go
```diff
@@ -821,3 +1511,191 @@ func TestConfigureKubernetes_SuccessfulRKE2ServerWithManifests(t *testing.T) {
...
+		},
+	}
+
+	manifest, err := kubernetesVIPManifest(k8s)
+	require.NoError(t, err)
+
+	assert.Contains(t, manifest, "- 192.168.1.1/32")
+	assert.Contains(t, manifest, "- name: rke2-api")
+	assert.NotContains(t, manifest, "ipFamilies:\n      - IPv6")
+	assert.NotContains(t, manifest, "ipFamilyPolicy: SingleStack")
```
atanasdinov, line 1529: We can have assertions for both `SingleStack` and `RequireDualStack` in all cases in order to easily make the difference. All templated values under the different conditions should be tested.

pkg/combustion/kubernetes_test.go
```diff
@@ -821,3 +1511,191 @@ func TestConfigureKubernetes_SuccessfulRKE2ServerWithManifests(t *testing.T) {
...
+	}
+
+	serverConfig := map[string]any{
+		"cluster-cidr": "fd12:3456:789b::/48,10.42.0.0/16",
+		"service-cidr": "fd12:3456:789c::/112,10.43.0.0/16",
+	}
+
+	nodeIPScript, err := createNodeIPScript(ctx, serverConfig)
+	require.NoError(t, err)
+	assert.Contains(t, nodeIPScript, "set-node-ip.sh")
```
atanasdinov, line 1588: Please also adjust all other references. Using `Contains()` will make assertions pass but it's much more prone to missing a symbol or two in strings that we need to ensure are as expected.
```suggestion
	assert.Equal(t, "set-node-ip.sh", nodeIPScript)
```

pkg/combustion/kubernetes_test.go
```diff
@@ -821,3 +1511,191 @@ func TestConfigureKubernetes_SuccessfulRKE2ServerWithManifests(t *testing.T) {
...
+	require.NoError(t, err)
+
+	assert.Contains(t, manifest, "- 192.168.1.1/32")
+	assert.Contains(t, manifest, "- fd12:3456:789a::21/128")
+	assert.Contains(t, manifest, "- name: k8s-api")
+	assert.NotContains(t, manifest, "ipFamilies:\n      - IPv6")
+	assert.NotContains(t, manifest, "ipFamilyPolicy: SingleStack")
+}
+
+func TestCreateNodeIPScriptDualstackIPv6PrioK3s(t *testing.T) {
```
atanasdinov, line 1569: Nit: Such names are very hard to read. We can use underscores as many times as necessary to make it better. E.g.
```suggestion
func TestCreateNodeIPScript_DualStack_IPv6Prio_K3s(t *testing.T) {
```

pkg/combustion/kubernetes_test.go
```diff
@@ -821,3 +1511,191 @@ func TestConfigureKubernetes_SuccessfulRKE2ServerWithManifests(t *testing.T) {
...
+	require.NoError(t, err)
+	assert.Contains(t, nodeIPScript, "set-node-ip.sh")
+
+	nodeIPScriptPath := filepath.Join(ctx.CombustionDir, setNodeIPScript)
+	b, err := os.ReadFile(nodeIPScriptPath)
+	require.NoError(t, err)
+
+	contents := string(b)
+
+	assert.Contains(t, contents, "IPv6=true")
```
atanasdinov, line 1679: This value is always true. In fact it's obsolete to include in the first place and this also applies to the other cases. `IPv4` is what we always want to assert because it's templated.

pkg/combustion/kubernetes_test.go
```diff
@@ -523,8 +1209,11 @@ func TestConfigureKubernetes_SuccessfulMultiNodeRKE2Cluster(t *testing.T) {
 		"token": "123",
 		"cni":   "canal",
 		"tls-san": []string{
-			"192-168-122-100.sslip.io",
+			"k8s-host.com",
 		},
+		"cluster-cidr": "fd12:3456:789b::/48,10.42.0.0/16",
+		"service-cidr": "fd12:3456:789c::/112,10.43.0.0/16",
+		"node-ip":      "192.168.122.101,fd12:3456:789a::21",
```
atanasdinov, line 1216: Should the priority be inverted?

pkg/combustion/kubernetes_test.go
```diff
@@ -821,3 +1513,190 @@ func TestConfigureKubernetes_SuccessfulRKE2ServerWithManifests(t *testing.T) {
...
+
+	manifest, err := kubernetesVIPManifest(k8s)
+	require.NoError(t, err)
+
+	assert.Contains(t, manifest, "- 192.168.1.1/32")
+	assert.Contains(t, manifest, "- fd12:3456:789a::21/128")
+	assert.Contains(t, manifest, "- name: k8s-api")
+	assert.NotContains(t, manifest, "ipFamilies:\n      - IPv6")
+	assert.NotContains(t, manifest, "ipFamilyPolicy: SingleStack")
+	assert.Contains(t, manifest, "ipFamilyPolicy: RequireDualStack")
```
atanasdinov, line 1571: Let's also add:
```
assert.Contains(t, manifest, "ipFamilies:\n - IPv4\n - IPv6")
```

pkg/combustion/kubernetes_test.go
```diff
@@ -821,3 +1513,190 @@ func TestConfigureKubernetes_SuccessfulRKE2ServerWithManifests(t *testing.T) {
...
+}
+
+func TestCreateNodeIPScript_Dualstack_IPv6_PrioK3s(t *testing.T) {
```
atanasdinov, line 1574:
```suggestion
func TestCreateNodeIPScript_DualStack_K3s_PrioIPv6(t *testing.T) {
```

pkg/combustion/kubernetes_test.go
```diff
@@ -821,3 +1513,190 @@ func TestConfigureKubernetes_SuccessfulRKE2ServerWithManifests(t *testing.T) {
...
+	require.NoError(t, err)
+
+	contents := string(b)
+
+	assert.Contains(t, contents, "IPv4=true")
+	assert.Contains(t, contents, "prioritizeIPv6=true")
+	assert.Contains(t, contents, "CONFIG_FILE=\"/etc/rancher/k3s/config.yaml\"")
+}
+
+func TestCreateNodeIPScript_Dualstack_IPv4_PrioRke2(t *testing.T) {
```
atanasdinov, line 1606:
```suggestion
func TestCreateNodeIPScript_DualStack_RKE2_PrioIPv4(t *testing.T) {
```

pkg/combustion/kubernetes_test.go
```diff
@@ -821,3 +1513,190 @@ func TestConfigureKubernetes_SuccessfulRKE2ServerWithManifests(t *testing.T) {
...
+	nodeIPScript, err := createNodeIPScript(ctx, serverConfig)
+	require.NoError(t, err)
+	assert.Equal(t, "set-node-ip.sh", nodeIPScript)
+
+	nodeIPScriptPath := filepath.Join(ctx.CombustionDir, setNodeIPScript)
+	b, err := os.ReadFile(nodeIPScriptPath)
+	require.NoError(t, err)
+
+	contents := string(b)
+	assert.Contains(t, contents, "CONFIG_FILE=\"/etc/rancher/rke2/config.yaml\"")
```
atanasdinov, line 1681: Missing `IPv4` and `prioritizeIPv6` assertions.

atanasdinov (approved):
Great work!

PR #618, opened 2024-11-20
Added raw combustion output option as default

dgiebert:


rdoxenham:
Thanks so much for the comprehensive contribution @dgiebert - can you help me understand this a little bit? Looks like you're using EIB to generate just the combustion directory and generating a tarball? Can you please expand on this and perhaps also add some example documentation to your commit? Thanks again.

dgiebert:
Yes, for our Hackweek project to combine EIB with SUMA/Kiwi, we had the idea to reverse the image creation process.
1. Create the Combustion + Artefact directory 
2. Build the custom image using SUMA

This has the advantage of having all the benefits of Kiwi, only building once, and an easier way to enable frozen patch levels.

atanasdinov (commented):
I know it's been quite a while since this was proposed but it seems we are now finally able to proceed with it.

I really appreciate the effort that went into this and we'd like to integrate it in EIB, but it'd require some additional work around extracting this feature in a separate subcommand. I'll follow with details offline.

PR #634, opened 2025-01-04
Create the required Elemental Agent directory structure during Combustion

### LINKED ISSUE #632: Elemental integration with EIB 1.1 and SL Micro 6.1 requires additional directory to be created
Just did a quick test using EIB 1.1 with SL Micro 6.1 based on https://github.com/Martin-Weiss/kiwi-image-micro-gpu-60 and during bootup elemental failed.

Workaround was 

`mkdir -p /etc/rancher/elemental/agent`

So maybe we need to create this sub-directory as part of the elemental part in EIB?

rdoxenham:
In this PR we create the required /etc/rancher/elemental/agent directory structure during the initial Elemental setup code. A config file is checked within the systemd startup condition, and it seemingly fails if this directory isn't already in-place. This seems to only be an issue in SL Micro 6.1, but it shouldn't harm for previous versions of Micro.

Fixes #632

### DIFF
--- a/pkg/combustion/templates/31-elemental-register.sh.tpl
+++ b/pkg/combustion/templates/31-elemental-register.sh.tpl
@@ -1,6 +1,7 @@
 #!/bin/bash
 set -euo pipefail
 
+mkdir -p /etc/rancher/elemental/agent
 mkdir -p /etc/elemental
 cp ./{{ .ConfigFile }} /etc/elemental/config.yaml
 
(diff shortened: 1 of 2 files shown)


pkg/combustion/templates/31-elemental-register.sh.tpl
```diff
@@ -79,7 +80,7 @@ cat <<- \EOF > /opt/edge/elemental_node_cleanup.sh
 
 # This script attempts to cleanup a node that has been deployed via Edge Image
 # Builder with the integrations for Elemental registration; in other words,
-# vanilla SLE Micro 5.5, *not* SLE Micro for Rancher (also known as Elemental
+# vanilla SLE Micro, *not* SLE Micro for Rancher (also known as Elemental
```
atanasdinov, line 83: Should we change the wording on `SUSE Linux Micro` now?

atanasdinov (commented):
Can we also drop a line in the release notes for this? Other than that, LGTM!

rdoxenham:
Thanks @atanasdinov - done!

atanasdinov (approved):

PR #672, opened 2025-03-14
Fix stale cache

### LINKED ISSUE #625: Rebuilds don't check for image cache staleness of :latest tag
If you build an image that includes OCI artifacts with the `:latest` tag then the container is updated on the registry, the new version is not pulled when EIB runs again.

dbw7:
Closes #625 

For images tagged `latest`, before pulling the image we get the digest for the correct architecture and platform and append that digest to the name of the image to suit our cache implementation. This allows our caching to now properly handle images tagged `latest`. 

As a dependency to reduce the requirement for processing this programmatically, `jq` was added to the EIB container image.

### DIFF
--- a/pkg/combustion/registry.go
+++ b/pkg/combustion/registry.go
@@ -238,7 +238,7 @@ func (c *Combustion) populateRegistry(ctx *image.Context, images []string) error
 		if strings.Contains(img, ":latest") {
 			var digest string
 			digest, err = c.ImageDigester.ImageDigest(img, arch)
-			if err != nil || digest == "" {
+			if err != nil {
 				zap.S().Warnf("Failed getting digest for %s: %s", img, err)
 				cacheImage = false
 			} else {
--- a/pkg/combustion/registry.go
+++ b/pkg/combustion/registry.go
@@ -9,11 +9,13 @@ import (
 	"path/filepath"
 	"slices"
 	"strings"
+	"time"
 
 	"github.com/schollz/progressbar/v3"
 	"github.com/suse-edge/edge-image-builder/pkg/fileio"
 	"github.com/suse-edge/edge-image-builder/pkg/image"
 	"github.com/suse-edge/edge-image-builder/pkg/log"
+	"github.com/suse-edge/edge-image-builder/pkg/podman"
 	"github.com/suse-edge/edge-image-builder/pkg/template"
 	"go.uber.org/zap"
 )
@@ -207,8 +209,13 @@ func registryArtefactsPath(ctx *image.Context) string {
 }
 
 func populateRegistry(ctx *image.Context, images []string) error {
+	p, err := podman.New(ctx.BuildDir)
+	if err != nil {
+		zap.S().Warnf("Setting up Podman instance: %v", err)
+	}
+
 	imageCacheDir := filepath.Join(ctx.CacheDir, "images")
-	if err := os.MkdirAll(imageCacheDir, os.ModePerm); err != nil {
+	if err = os.MkdirAll(imageCacheDir, os.ModePerm); err != nil {
 		return fmt.Errorf("creating container image cache dir: %w", err)
 	}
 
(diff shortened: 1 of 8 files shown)


pkg/combustion/registry.go
```diff
@@ -262,3 +276,53 @@ func populateRegistry(ctx *image.Context, images []string) error {
 
 	return nil
 }
+
+func pullImageDigest(img string, arch string, outputWriter io.Writer) (string, error) {
+	podmanCmd := exec.Command("/usr/bin/podman", "manifest", "inspect", img)
```
atanasdinov, line 281: Did we investigate the approach of using the Hauler library itself?

pkg/combustion/registry.go
```diff
@@ -207,8 +209,13 @@ func registryArtefactsPath(ctx *image.Context) string {
 }
 
 func populateRegistry(ctx *image.Context, images []string) error {
+	p, err := podman.New(ctx.BuildDir)
+	if err != nil {
+		zap.S().Warnf("Setting up Podman instance: %v", err)
```
atanasdinov, line 214: We cannot proceed execution because `p` is nil after this. Shouldn't we just error out?

pkg/combustion/registry.go
```diff
@@ -234,6 +241,18 @@ func populateRegistry(ctx *image.Context, images []string) error {
...
 		convertedImage := strings.ReplaceAll(img, "/", "_")
 		convertedImageName := fmt.Sprintf("%s-%s", convertedImage, registryTarSuffix)
+		if strings.Contains(img, ":latest") {
+			var digest string
+			digest, err = p.Inspect(img, arch)
+			if err != nil {
+				zap.S().Warnf("Failed getting digest for %s: %s", img, err)
+
+				// In the case where we're not able to find a digest, we'll use a timestamp to prevent staleness
+				digest = fmt.Sprintf("%d", time.Now().Unix())
```
atanasdinov, line 251: Are we sure we want to do this? What happens on subsequent runs when such images are available?

pkg/combustion/registry.go
```diff
@@ -234,6 +241,18 @@ func populateRegistry(ctx *image.Context, images []string) error {
 	for _, img := range images {
 		convertedImage := strings.ReplaceAll(img, "/", "_")
 		convertedImageName := fmt.Sprintf("%s-%s", convertedImage, registryTarSuffix)
+		if strings.Contains(img, ":latest") {
+			var digest string
+			digest, err = p.Inspect(img, arch)
```
atanasdinov, line 246:
```suggestion
			digest, err := p.Inspect(img, arch)
```

pkg/podman/podman.go
```diff
@@ -144,6 +146,36 @@ func (p *Podman) Copy(id, src, dest string) error {
...
+			digest := foundDigest
+			if strings.Contains(foundDigest, ":") {
+				parts := strings.Split(foundDigest, ":")
+				if len(parts) == 2 {
+					digest = parts[1]
+				}
+			}
+
+			return digest, nil
+		}
```
atanasdinov, line 173: Simplifying suggestion:
```suggestion
	for _, manifest := range m.Manifests {
		if manifest.Platform.OS == "linux" && manifest.Platform.Architecture == arch {
			digest := manifest.Digest.String()

			// Remove "sha:" from the digest if present
			if parts := strings.Split(digest, "sha:"); len(parts) == 2 {
				digest = parts[1]
			}

			return digest, nil
		}
	}
```

pkg/podman/podman.go
```diff
@@ -144,6 +146,36 @@ func (p *Podman) Copy(id, src, dest string) error {
 	return nil
 }
 
+// Inspect returns the sha256 digest for a given container image on a specific architecture
+func (p *Podman) Inspect(img string, arch string) (string, error) {
```
atanasdinov, line 150: This function should be called `Digest()` and not `Inspect()` to properly reflect what it is doing.

pkg/podman/podman.go
```diff
@@ -144,6 +146,36 @@ func (p *Podman) Copy(id, src, dest string) error {
...
+				if len(parts) == 2 {
+					digest = parts[1]
+				}
+			}
+
+			return digest, nil
+		}
+	}
+
+	return "", nil
```
atanasdinov, line 176: Is this an error case? We never handle empty strings in the caller.

atanasdinov (changes requested):

pkg/podman/podman.go
dbw7: This is done because the linter complains about 
```
pkg/combustion/registry.go:289:2: rangeValCopy: each iteration copies 176 bytes (consider pointers or indexing) (gocritic)
```

pkg/podman/podman.go
dbw7: This would just be if there's no error in getting the manifest, but no manifest is found. In both this case and the error case we'd just use a timestamp so each image is always the newest one.

pkg/combustion/registry.go
dbw7: In that case it will just always grab the newest ":latest" image. The only concern would be the image cache will grow, but that's inconsequential IMO compared to not actually having the latest image.

pkg/combustion/registry.go
dbw7: This is done to avoid a shadow declaration that the linter complains about

atanasdinov (approved):

PR #676, opened 2025-03-14
Custom tempdir for EAR

dbw7:
This was a huge oversight. A solution to the RAM issue was introduced in `v1.0.2` that I completely missed. This change now allows Hauler to create the temporary directory that containers are loaded into in `/opt/hauler` as opposed to `/tmp`.

I have tested both versions of this and found that the final storage usage of `/opt/hauler` remains the same regardless of using `/tmp` or `/opt/hauler` as `--tempdir`.


rdoxenham (approved):
LGTM! 🚀

atanasdinov (approved):

PR #680, opened 2025-03-20
allow duplicate helm chart names

### LINKED ISSUE #606: How to handle duplicate helm chart with different valuesfile
EIB not support duplicate chart name currently, but no document for how to handle follow cases:

1. Different repositories with same chart name.
2. Same chart but want to use it with different valuesfile.

It does not make sense to rebuild a chart which add more external effort to maintain it.

Any suggestions?

dbw7:
Closes #606 

It is a valid configuration to install multiple Helm charts to the same or different namespaces as long as the name of the release is different. In our case this would be the metadata name of the Helm charts.

This PR removes the validation that prevents defining multiple Helm charts with the same name and instead automatically appends a suffix to the metadata names to be unique.

E.g. if ‘cert-manager’ is defined twice, the metadata names in the Helm CRDs would become ‘cert-manager-1’, ‘cert-manager-2’ which deploy with no issue.

```
kube-system    job.batch/helm-install-cert-manager-1                     Complete   1/1           43s        16m
kube-system    job.batch/helm-install-cert-manager-2                     Complete   1/1           43s        16m
```

```
cert-manager                      deployment.apps/cert-manager-1                         1/1     1            1           16m
cert-manager                      deployment.apps/cert-manager-1-cainjector              1/1     1            1           16m
cert-manager                      deployment.apps/cert-manager-1-webhook                 1/1     1            1           16m
cert-manager                      deployment.apps/cert-manager-2                         1/1     1            1           16m
cert-manager                      deployment.apps/cert-manager-2-cainjector              1/1     1            1           16m
cert-manager                      deployment.apps/cert-manager-2-webhook                 1/1     1            1           16m
```

We will also now instead spit out a warning so that users are aware of this in case it is not intentional
```
WARNING: The Helm chart 'rancher' has been defined multiple times, while this is a valid configuration, please ensure this is intentional.
WARNING: The Helm chart 'cert-manager' has been defined multiple times, while this is a valid configuration, please ensure this is intentional.
```

### DIFF
--- a/pkg/registry/registry.go
+++ b/pkg/registry/registry.go
@@ -121,29 +121,29 @@ func storeHelmCharts(ctx *image.Context, helmClient helmClient) ([]*helmChart, e
 	chartRepositories := mapChartsToRepos(helm)
 
 	var charts []*helmChart
-	seenHelmCharts := make(map[string]string)
+	helmChartPaths := make(map[string]string)
 
 	for i := range helm.Charts {
 		chart := helm.Charts[i]
-		chartConfig := fmt.Sprintf("%s-%s-%s", chart.RepositoryName, chart.Name, chart.Version)
+		chartID := fmt.Sprintf("%s-%s-%s", chart.RepositoryName, chart.Name, chart.Version)
 
 		repository, ok := chartRepositories[helm.Charts[i].RepositoryName]
 		if !ok {
 			return nil, fmt.Errorf("repository not found for chart %s", helm.Charts[i].Name)
 		}
 
-		if _, exists := seenHelmCharts[chartConfig]; !exists {
+		if _, exists := helmChartPaths[chartID]; !exists {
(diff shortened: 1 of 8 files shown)


pkg/registry/registry.go
```diff
@@ -121,21 +121,29 @@ func storeHelmCharts(ctx *image.Context, helmClient helmClient) ([]*helmChart, e
...
+				return nil, fmt.Errorf("downloading chart: %w", err)
+			}
+
+			seenHelmCharts[chartConfig] = localPath
 		}
 
 		charts = append(charts, &helmChart{
 			HelmChart:     helm.Charts[i],
-			localPath:     localPath,
+			localPath:     seenHelmCharts[chartConfig],
```
dbw7, line 146: This is so that we don't download the same chart multiple times

atanasdinov (dismissed):

atanasdinov (approved):

Issue #593, opened 2024-10-24
OS files script should mount /var


atanasdinov:
The current state of the script will not mount /var which means that /os-files/var/* files will not be copied over. 

Related: https://github.com/suse-edge/edge-image-builder/issues/526

PR #710, opened 2025-05-06
explicitly mount and unmount /usr/local for os files handling

Cavaler:
Similar to https://github.com/suse-edge/edge-image-builder/issues/593, files in os-files/usr/local are not copied properly


atanasdinov (commented):
Hello! Thanks for this contribution. Could you please add an entry in the release notes?

PR #715, opened 2025-05-15
Clean up version usage

dbw7:
This PR largely cleans up a lot of things and improves the handling for the new `apiVersion`

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -30,6 +30,8 @@
 
 ### Image Definition Changes
 
+* The current version of the image definition has been incremented to `1.2` to include the changes below 
+  * Existing definitions using the `1.0` and `1.1` versions of the schema will continue to work with EIB
 * Added `kubernetes.network.apiVIP6` field to enable cluster LoadBalancer based on IPv6 address
 * Added `operatingSystem.enableExtras` flag to enable the SUSE Linux Extras repository during RPM resolution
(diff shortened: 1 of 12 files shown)


RELEASE_NOTES.md
```diff
@@ -85,6 +87,7 @@
 * The current version of the image definition has been incremented to `1.1` to include the changes below 
   * Existing definitions using the `1.0` version of the schema will continue to work with EIB
 * Introduced a dedicated FIPS mode option (`enableFIPS`) which will enable FIPS mode on the node
+* Adds an optional `apiVersions` field under Helm charts
```
dbw7, line 90: This was in the tagged notes, but missing from the release notes, I just copied it over as is

pkg/image/validation/version.go
```diff
@@ -21,22 +21,61 @@ func validateVersion(ctx *image.Context) []FailedValidation {
 		}
 	}
 
-	if definition.APIVersion == "1.0" && apiVersionsDefined {
-		failures = append(failures, FailedValidation{
-			UserMessage: "Helm chart APIVersions field is not supported in EIB version 1.0, must use EIB version 1.1",
-		})
-	}
+	if definition.APIVersion == "1.0" {
```
atanasdinov, line 24: Can we explore some generic way of doing this? It is clearly not going to scale well with 1.3 and higher.

Perhaps a switch statement for the version and then a map of keys and fields which are only supported from a specific version forward.

atanasdinov: Submitted the changes below.

atanasdinov (dismissed):

PR #750, opened 2025-07-15
Generate combustion drive

dbw7:
Initial setup to see if everything looks solid, will do the finishing touches once we decide that these design decisions are correct.

Usage examples
```
podman run --rm -it -v $(pwd):/eib eib:local generate --definition-file test.yaml --output-type combustion-iso
podman run --rm -it -v $(pwd):/eib eib:local generate --definition-file test.yaml --output-type tar
```


pkg/image/definition.go
```diff
@@ -13,8 +13,10 @@ import (
 )
 
 const (
-	TypeISO = "iso"
-	TypeRAW = "raw"
+	TypeISO           = "iso"
+	TypeRAW           = "raw"
+	TypeTar           = "tar"
+	TypeCombustionIso = "combustion-iso"
```
atanasdinov, line 19: These should be named `CombustionTypeTar` and `CombustionTypeISO` in my opinion. The value for the latter should also just be `iso`.

pkg/cli/build/generate.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	buildDir, err := eib.SetupBuildDirectory(rootBuildDir)
+	if err != nil {
+		log.Audit("The build directory could not be set up.")
+		return err
+	}
+
+	// This needs to occur as early as possible so that the subsequent calls can use the log
+	log.ConfigureGlobalLogger(filepath.Join(buildDir, buildLogFilename))
```
atanasdinov, line 35: Does it make sense to use different log file?

pkg/cli/cmd/flags.go
```diff
@@ -14,4 +14,21 @@ var (
 		Value:       "/eib",
 		Destination: &BuildArgs.ConfigDir,
 	}
+	GenerateDefinitionFileFlag = &cli.StringFlag{
```
atanasdinov, line 17: In fact, why do we have to duplicate all these? Can't we reuse the exact same ones we already have?

pkg/cli/build/generate.go
```diff
@@ -0,0 +1,83 @@
...
+	artifactSources, err := parseArtifactSources()
+	if err != nil {
+		log.Auditf("Loading artifact sources metadata failed. %s", checkBuildLogMessage)
+		zap.S().Fatalf("Parsing artifact sources failed: %v", err)
+	}
+
+	ctx := buildContext(buildDir, combustionDir, artefactsDir, generateArgs.GenerateConfigDir, imageDefinition, artifactSources)
+
+	// Set the image type for combustion - either tar or combustion-iso
+	ctx.ImageDefinition.Image.ImageType = generateArgs.GenerateOutputType
```
atanasdinov, line 63: I believe this is not the way to go.

I know it makes the code easier but the output type can be part of the Generator type.

The validation would obviously have to be adjusted but is the `image` field of the definition something that we'd want to drop altogether in this flow? What we're doing here is not building an image, hence usage of "image type", "image output name", etc. seems confusing to me. I appreciate we shouldn't try and restructure the whole repository in this feature but this change doesn't seem quite right.

pkg/cli/build/generate.go
```diff
@@ -0,0 +1,83 @@
...
+	defer func() {
+		if r := recover(); r != nil {
+			log.Auditf("Build failed unexpectedly. %s", checkBuildLogMessage)
+			zap.S().Fatalf("Unexpected error occurred: %s", r)
+		}
+	}()
+
+	if err = eib.Generate(ctx, rootBuildDir); err != nil {
+		log.Audit(checkBuildLogMessage)
+		zap.S().Fatalf("An error occurred building the image: %s", err)
```
atanasdinov, line 79: The logs can use some work, e.g. `An error occurred while generating the config drive` or something similar.

pkg/eib/eib.go
```diff
@@ -211,6 +224,53 @@ func buildCombustion(ctx *image.Context, rootDir string) (*combustion.Combustion
 	return combustionHandler, nil
 }
 
+func generateCombustion(ctx *image.Context, rootDir string) (*combustion.Combustion, error) {
```
atanasdinov, line 227: This brings duplication without an obvious benefit. Especially as `combustion.SkipRPMComponent` already accounts for the new flow. Let's see how this evolves after we apply the rest of the changes.

atanasdinov:
The validation should also be adapted to be stricter. Adding in RPMs or kernel args for `eib generate` should fail as these are not valid and the user should be fully understanding what they're doing.

In fact, I'm now wondering if introducing a `ConfigDriveDefinition` makes sense where that would satisfy both the validation and API constraints and the `Context` type can work with a `Definition` interface that is backed by different implementations for the different flows. Let's discuss further offline.

atanasdinov (approved):
🚀

PR #793, opened 2025-10-13
Elemental service tweaks

### LINKED ISSUE #784: dbus and elemental-system-agent Race Condition Causes Elemental Downstream Cluster Deployment Failures
**Problem Description**

During Elemental downstream cluster provisioning, a race condition occurs between the `dbus.service` and `elemental-system-agent.service`.

The [elemental-system-agent.service](https://github.com/suse-edge/edge-image-builder/blob/main/pkg/combustion/templates/31-elemental-register.sh.tpl#L29-L49) is defined by EIB and requires `dbus` to be fully operational to perform its tasks, such as starting `rancher-system-agent`. However, it sometimes starts before `dbus.service` is fully up, leading to a "_Message recipient disconnected from message bus without replying_" error. This failure prevents the `rancher-system-agent` from starting, causing the Elemental downstream cluster deployment to fail. Here is a sample log snippet.

`Sep 09 19:38:07 elementalvm elemental-system-agent[3671]: time="2025-09-09T19:38:07Z" level=info msg="[6b20fe64c854da2639804884b34129bb8f718eb59578111da58d9de1509c24db_1:stderr]: Failed to restart rancher-system-agent.service: Message recipient disconnected from message bus without replying"`

**Proposed Workaround**

To temporarily mitigate this issue, a systemd service override can be used. This override ensures `elemental-system-agent.service` waits for `dbus.service` to be active. The override file, `override.conf`, should be placed at `/etc/systemd/system/elemental-system-agent.service.d`. The contents of `override.conf` is below.

```
[Unit]
Wants=dbus.service network-online.target
After=dbus.service network-online.target time-sync.target

[Service]
ExecStartPre=/bin/bash -c 'echo "Waiting for dbus to become active..." | systemd-cat -p info -t elemental-system-agent; sleep 15; timeout 300 bash -c "while ! systemctl is-active --quiet dbus.service; do sleep 15; done"'
```

A custom script is required to copy this override file into place before the EIB's [31-elemental-register.sh](https://github.com/suse-edge/edge-image-builder/blob/main/pkg/combustion/templates/31-elemental-register.sh.tpl) script runs during the combustion phase. (e.g., `30a-copy-elemental-system-agent-override.sh`)

```
#!/bin/bash

/bin/mkdir -p /etc/systemd/system/elemental-system-agent.service.d
/bin/cp -f elemental-system-agent-override.conf /etc/systemd/system/elemental-system-agent.service.d/override.conf
```

While the workaround helps, it's not a complete solution. The core problem lies in the dynamic nature of the `dbus.service`. During the boot process, `dbus.service` starts and then gets restarted when the hostname is set. This restart can occur while `elemental-system-agent` is already running and performing tasks, leading to the same "_disconnected from message bus_" error if the timing is off.

```
Sep 17 19:38:51 localhost systemd[1]: Starting D-Bus System Message Bus...
Sep 17 19:38:51 localhost systemd[1]: Started D-Bus System Message Bus.
Sep 17 19:39:23 elementalvm systemd[1]: Stopping D-Bus System Message Bus...
Sep 17 19:39:23 elementalvm systemd[1]: dbus.service: Deactivated successfully.
Sep 17 19:39:23 elementalvm systemd[1]: Stopped D-Bus System Message Bus.
Sep 17 19:39:27 elementalvm systemd[1]: Starting D-Bus System Message Bus...
Sep 17 19:39:27 elementalvm dbus-daemon[2244]: [system] SELinux support is enabled
Sep 17 19:39:27 elementalvm systemd[1]: Started D-Bus System Message Bus.
```

The current workaround's `ExecStartPre` command includes a `sleep 15` and then checks for `dbus` to be `active`. If the initial `dbus.service` takes longer than 15 seconds to come up, or if it restarts while the agent is active, the race condition reappears. A more robust solution is needed to handle these dynamic `dbus` restarts effectively, ensuring a persistent and reliable connection for the `elemental-system-agent`. The current workaround addresses the initial race but not the subsequent, hostname-related restart of `dbus.service`.

dbw7:
Closes #784

The Elemental system agent service can be started before the dbus service is started. This can result in dbus errors when clusters are deployed on the Elemental instance. This change to the elemental-system-agent service Binds it to the dbus service so that it restarts when the dbus service restarts, always keeping it in the latest state and preventing the elemental cluster deployment error.

A regression is possible in untested scenarios where the dbus service can behave unexpectedly due to system timing or user configuration.


pkg/combustion/templates/31-elemental-register.sh.tpl
```diff
@@ -34,6 +34,9 @@ Documentation=https://github.com/rancher/system-agent
 Wants=network-online.target
 After=network-online.target
 After=time-sync.target
+Wants=dbus.service
+After=dbus.service
+BindsTo=dbus.service
```
atanasdinov: I'm really conscious of establishing such a strong dependency on dbus. Can we get away with `Requires=` instead and / or depending on the dbus.socket rather than the service?

PR #794, opened 2025-10-13
Kubernetes resource install service changes

### LINKED ISSUE #761: Kubernetes Resource install one-shot-service does not restart in case of failures
# Observation

In case a system is installed with an EIB built ISO with pre-configured helm charts, it fails to install these charts in case network is not connected during initial boot. This also occurs in case e.g. a DHCP server is slow (> +-20s) to hand out an IP address.

Once network connection is restored, K3s will start. But only a set of 5 core pods will be running.

Only after a restart where a network is available sufficiently fast, the Kubernetes resources will be installed.

Impact of this, is that installing a system in unstable network conditions (or even our enterprise network) results in unpredictable installs and effort spent on faultfinding.

# Configuration used

* Edge Image Builder 1.2.0
* SLE Micro 6.1

Definition file:
```YAML
apiVersion: 1.1
image:
  imageType: iso
  arch: x86_64
  baseImage: SL-Micro.x86_64-6.1-Base-SelfInstall-GM.install.iso
  outputImageName: my-image.iso
[...]  
kubernetes:
  version: v1.33.1+k3s1
  helm:
    repositories:
      - name: jetstack
        url: https://charts.jetstack.io/
    charts:
      - name: cert-manager
        repositoryName: jetstack
        version: v1.17.1
        targetNamespace: cert-manager
        installationNamespace: kube-system
        valuesFile: cert-manager.yaml
```

# Reproduce with bare-metal or VM

* Build an image with at minimal 1 Kubernetes resource configured
* Install the image on the target
* Ensure that during first boot, no network with gateway is connected (K3s will fail to start)
* wait 1 minute
* Observe that the one-shot-service "Kubernetes Resource install" is dead

* Fix network connection (the ip address is shown on the console)
* Wait until K3s is started (check: http://127.0.0.1:6443/readyz?verbose returns `readyz check passed` )
* Observe that only core K3s pods are running, and chart specified in the EIB definition file is not installed
* Observe that the one-shot-service "Kubernetes Resource install" remains dead

* Reboot
* Observe that after a while the "Kubernetes Resource install" is finished, and the specified chart is installed

# Analysis

Unit file

```
[Unit]
Description=Kubernetes Resources Install
Requires=k3s.service
After=k3s.service
ConditionPathExists=/opt/bin/kubectl
ConditionPathExists=/etc/rancher/k3s/k3s.yaml

[Install]
WantedBy=multi-user.target

[Service]
Type=oneshot
Restart=on-failure
RestartSec=60
ExecStartPre=/bin/sh -c 'until [ "\$(systemctl show -p SubState --value k3s.service)" = "running" ]; do sleep 10; done'
ExecStart=/opt/eib-k8s/create_manifests.sh
# Disable the service and clean up
ExecStartPost=/bin/sh -c "systemctl disable kubernetes-resources-install.service"
ExecStartPost=rm -f /etc/systemd/system/kubernetes-resources-install.service
ExecStartPost=rm -rf /opt/eib-k8s
```

A `oneshot` service is defined, which requires `k3s.service` and has a condition on path-exist of `k3s.yaml`. 
Both conditions will not be met if K3s crashes due to a missing network connection.

Due to this resulting in a failing precondition on start, there will be no retries.

# Workaround applied

In our own build, we now apply a workaround to edit the above unit file and remove lines `Requires=k3s.service`, `ConditionPathExists=/etc/rancher/k3s/k3s.yaml`. With those changes, the `ExecStartPre` will wait until K3s is ready, and then the Kubernetes resources are installed once network connection is restored.

dbw7:
Closes #761

The kubernetes resources install service requires the kubernetes service to be running, as well as certain paths to exist. When the kubernetes service is not running or these paths do not exist, the kubernetes-resources-install service will just die and not restart as it fails to meet its necessary conditions. This change removes the need for those conditions, and instead has it rely on `ExecStartPre` where it will wait continuously for the kubernetes service to be started before it starts.

No obvious possible regressions come to mind. (The `ExecStartPre` issue of not running long processes doesn't seem to be an issue here because it's ok for the watcher process it performs to be killed before started the actual process as that is the intention of the service).

### DIFF
--- a/pkg/combustion/templates/k3s-multi-node-installer.sh.tpl
+++ b/pkg/combustion/templates/k3s-multi-node-installer.sh.tpl
@@ -74,7 +74,7 @@ WantedBy=multi-user.target
 Type=oneshot
 Restart=on-failure
 RestartSec=60
-ExecStartPre=/bin/sh -c 'until [ "\$(systemctl show -p SubState --value k3s.service)" = "running" ]; do sleep 10; done'
+ExecStartPre=/bin/sh -c 'until /opt/bin/kubectl get nodes >/dev/null 2>&1; do sleep 10; done'
 ExecStart=/opt/eib-k8s/create_manifests.sh
 # Disable the service and clean up
 ExecStartPost=/bin/sh -c "systemctl disable kubernetes-resources-install.service"
--- a/pkg/combustion/templates/k3s-single-node-installer.sh.tpl
+++ b/pkg/combustion/templates/k3s-single-node-installer.sh.tpl
@@ -48,7 +48,7 @@ WantedBy=multi-user.target
 Type=oneshot
 Restart=on-failure
 RestartSec=60
-ExecStartPre=/bin/sh -c 'until [ "\$(systemctl show -p SubState --value k3s.service)" = "running" ]; do sleep 10; done'
+ExecStartPre=/bin/sh -c 'until /opt/bin/kubectl get nodes >/dev/null 2>&1; do sleep 10; done'
 ExecStart=/opt/eib-k8s/create_manifests.sh
 # Disable the service and clean up
 ExecStartPost=/bin/sh -c "systemctl disable kubernetes-resources-install.service"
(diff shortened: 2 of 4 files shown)


rdoxenham:
> @dbw7 I advise to check for errors in journalctl after this change. My patched version of the systemd unit file works, but also leads to a failure logged every 70s:

Hey @pbronneberg -- thanks for your inputs here.

Do you see this even after the service has completed?

pbronneberg:
I see this constantly in the logs. Even for a system that had its first install 2 weeks ago

PR #911, opened 2026-09-10
Set a default libguestfs backend

### LINKED ISSUE #910: 4K sector fallback logic drops direct libguestfs backend, causing libvirt socket crash
**Symptom:**

Building with base images that use 4K sectors fails with the following error, whereas 512-byte sector images succeed:

`libvirt: XML-RPC error : Failed to connect socket to '/var/run/libvirt/virtqemud-sock': No such file or directory
`
**Root Cause:**
When the initial 512-byte probe fails, the fallback logic (from PR #447) creates a new guestfs handle to retry with 4096 bytes. However, this new handle fails to inherit or re-apply the direct backend configuration. It defaults to the OS libvirt backend and crashes because the socket is not mounted inside the container.

**Workaround:**
Passing -e LIBGUESTFS_BACKEND=direct to the podman run command globally forces the retry loop to stay in direct mode.

This issue might be related to #447.

atanasdinov:
libguestfs upstream defaults to the libvirt backend, which needs virtqemud/libvirtd listening on /var/run/libvirt/virtqemud-sock. The EIB container installs libguestfs/guestfs-tools but no libvirt daemon, so the default only ever worked because the packaging previously shipped direct as the default.

A package update has likely flipped that back to the upstream default. This change ensures that the direct backend is used, unless modified by the user.

Closes #910


dbw7:
Doing it like this is good, a possible alternative is also adding:
```
ENV LIBGUESTFS_BACKEND=direct
```

To the main `Dockerfile` under `FROM opensuse/leap:15.6` after the zypper install step

atanasdinov:
Sure, but the code change ensures that the logic is always going to work regardless of the packaging process and mechanism.

fdegir (approved):
LGTM

dbw7 (approved):
LGTM

